Issue #8, opened 2016-06-14
Manual installation download links to owncloud


Whoops:
Found on: https://docs.nextcloud.org/server/9/admin_manual/installation/source_installation.html

The link that says "Nextcloud Download Page" goes to https://owncloud.org/install

mar1u5:
Fixed with #9 :wink: Thank you for reporting :muscle:

PR #9, opened 2016-06-14
update download link in upgrades and correct title underline length after rename to nextcloud

adsworth:


adsworth:
this also fixes #8

LukasReschke:
Awesome! Thanks a lot 🚀

PR #10, opened 2016-06-14
Update Oracle DB from Requirements page

mar1u5:
cc @LukasReschke

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -26,7 +26,7 @@ Supported Platforms
   Red Hat Enterprise Linux/Centos 6.5 and 7 (7 is 64-bit only), Ubuntu 12.04 
   LTS, 14.04 LTS, 14.10)
 * Web server: Apache 2 with mod_php
-* Databases: MySQL/MariaDB 5.5+; Oracle 11g (Nextcloud Enterprise edition only); PostgreSQL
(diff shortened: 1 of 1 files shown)


MorrisJobke:
I guess we need to figure out if this is okay for customers first :see_no_evil: 

cc @karlitschek @jospoortvliet

mar1u5:
Oracle is not compatible with Nextcloud at this time, because it was realized with an enterprise app.

LukasReschke:
> Oracle is not compatible with Nextcloud at this time, because it was realized with an enterprise app.

It depends. Actually the whole Oracle logic lives in the server and the Enterprise app just set a config flag that enabled it.

That said, Oracle support is inherently broken. Many things do not work in the server properly, apps like calendar and contacts break as well, and PHP 7 has a broken Oracle adapter.

I really would advise anybody to stay away from using Nextcloud with Oracle. You enter a world of mainly untested pain ;-)

karlitschek:
I don't think we should drop support for this at the moment without more information. Let's discuss again for the next major release.

MorrisJobke:
@mar1u5 Could you adjust the text to state that the support needs to be contacted in advance to setup a Oracle based instance (*hint* Subscription is then needed *hint*)

mar1u5:
Done 😉

MorrisJobke:
Awesome :) Thanks :+1:

MorrisJobke:
master f962503

PR #18, opened 2016-06-16
revise updater.server.url to reflect nextcloud.org URL

hailthemelody:
Please see https://github.com/nextcloud/server/issues/102 for further information.

(meta: including from conversation in above url @LukasReschke )

PR #20, opened 2016-06-16
[stable9] revise updater.server.url to reflect nextcloud.org URL

hailthemelody:
Please see https://github.com/nextcloud/documentation/pull/18 for more info

With thanks to @MorrisJobke and @mar1u5 for their help!

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -601,7 +601,7 @@ available.
 
 ::
 
-	'updater.server.url' => 'https://updates.owncloud.com/server/',
+	'updater.server.url' => 'https://updates.nextcloud.org/server/',
 
 URL that Nextcloud should use to look for updates
 


mar1u5:
👍

hailthemelody:
Noob aside question - I forgot to add my user.name and user.email on the original commits (docs and server). Is there an easy-ish way to add these? I've tried a few things from online documentation but it gets complicated quick. I can let it go, the project it the priority.

mar1u5:
See here: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

I'm sorry but I think it is not possible to add your email afterwards 😁 ...Maybe one of the **Git pros** here (yeah ... I mean you @MorrisJobke 😜) does know a way 😉

MorrisJobke:
> Maybe one of the Git pros here (yeah ... I mean you @MorrisJobke 😜) does know a way 😉

I know a way, but this means rewriting git history. I would not do it and leave it as it is. Is this okay for you? Otherwise we need to do some git magic.

hailthemelody:
Thank you, though it's not worth the amount of time it would take for the group. I'll let it go.

PR #25, opened 2016-06-20
Updated GPG key and {own,next}cloud references

ilpianista:
The GPG path was still pointing to owncloud server.

Also ownCloud was still mentioned.

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -2,16 +2,16 @@
 Manual Installation on Linux
 ============================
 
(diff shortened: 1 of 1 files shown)


schiessle:
:+1:

Btw, we should probably remove the information about the packages at the build server completely. For now we don't provide packages

schiessle:
Merged to solve the naming issues. But would be great to have a follow up pr for the issue above.

Do we need a backport? Can you check it? Thanks!

ilpianista:
Thanks for merging.

Yes we need a backport, I'm going to make it.

Ok, I'll drop the part about build servers with another PR.

PR #27, opened 2016-06-20
Drop lines about OBS packages

ilpianista:


schiessle:
:+1: Thanks! Seems like this one doesn't need a backport. The stable9 documentation doesn't mention the packages.

PR #32, opened 2016-06-21
Remove trailing / to reduce the risk of running find on /

twouters:
Removed trailing slash, if someone runs these commands without setting `$ncpath` this doesn't break the whole system.


MorrisJobke:
Makes sense 👍 Thanks

MorrisJobke:
master 09c2ee4

PR #48, opened 2016-07-08
Fix NGINX config for acme challenge

MorrisJobke:
fixes #46

PR #49, opened 2016-07-08
Fix NGINX config for acme challenge

MorrisJobke:
* same as #48 but for master

cc @sameliotfisher @LukasReschke @rullzer @josh4trunks

We should not document stuff that has nothing to do with Nextcloud ;)


josh4trunks:
looks good to me

PR #52, opened 2016-07-10
hide anchors in headers

MorrisJobke:
Looks like this:
bildschirmfoto 2016-07-10 um 10 50 48

Instead of this:
bildschirmfoto 2016-07-10 um 10 50 51

@jancborchardt Or did you add this by intention? I don't know if this is needed that much on mobile. So I dropped it there (because hover doesn't exist there)

### DIFF
--- a/_shared_assets/themes/nextcloud_com/static/styles.css
+++ b/_shared_assets/themes/nextcloud_com/static/styles.css
@@ -1076,3 +1076,13 @@ img.desaturate {
 .youtube i:hover {
     color: #CC181E;
 }
+
+h1:hover a.headerlink,
+h2:hover a.headerlink,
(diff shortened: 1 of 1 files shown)


mar1u5:
Very Nice :+1:

nickvergessen:
Can the link still be copied somehow?

jancborchardt:
I didn’t mess around with this, no. :) Showing only on hover is fine, as is also done for Github.

MorrisJobke:
> Can the link still be copied somehow?

@nickvergessen Sure: Click on it and copy the address bar or simply right click the character.


I will merge it :)

MorrisJobke:
master 28d3664]

PR #85, opened 2016-07-27
update supported apps

MorrisJobke:
cc @karlitschek @jospoortvliet @nickvergessen @oparoz

### DIFF
--- a/admin_manual/installation/apps_supported.rst
+++ b/admin_manual/installation/apps_supported.rst
@@ -6,28 +6,37 @@ AGPL Apps
 ---------
 
 * Activity
+* Admin Audit Log
 * AntiVirus
+* Calendar
 * Collaborative Tags
-* Comments 
+* Comments
+* Contacts
 * Encryption
 * External Sites
 * External Storage
-* Nextcloud WebDAV Endpoint (handles old and new webdav endpoints)
 * Federated File Sharing (allows file sharing across Nextcloud instances)
 * Federation (allows usernname auto-complete across Nextcloud instances)
 * Files (cannot be disabled)
+* Files Access Control
+* Files Automated Tagging
 * Files PDF Viewer
 * Files Sharing
-* Files TextEditor
+* Files Text Editor
 * Files Trashbin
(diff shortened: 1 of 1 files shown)


MorrisJobke:
cc @mar1u5

admin_manual/installation/apps_supported.rst
```diff
@@ -6,28 +6,37 @@ AGPL Apps
...
+* Calendar
 * Collaborative Tags
-* Comments 
+* Comments
+* Contacts
 * Encryption
 * External Sites
 * External Storage
-* Nextcloud WebDAV Endpoint (handles old and new webdav endpoints)
+* WebDAV Endpoint (handles old and new webdav endpoints)
```
nickvergessen: alphabet is a mess

MorrisJobke: damn -.-

karlitschek:
👍

admin_manual/installation/apps_supported.rst
```diff
@@ -6,28 +6,37 @@ AGPL Apps
...
 * Notifications
 * Object Storage (Swift)
+* Password Policy
 * Provisioning API
 * Template Editor (for notification emails)
+* Theming
 * Update Notifications
 * User External
 * User LDAP
+* User SAML
```
oparoz: I would use something like Shibboleth/SAML authentication which may be more meaningful to people looking at this list.

oparoz:
What about Windows Network Drive storage and Sharepoint?

MorrisJobke:
> What about Windows Network Drive storage and Sharepoint?

I haven't seen a repo here. So I would go for the "put it in the list once we have something" 😉

MorrisJobke:
> Should a list of supported backends be added to "External Storage"?

This is documented in the files_external setup. I don't think we want this, because then we need to write this for all the different apps: Like "Files Access Control supports to deny a request based on URL, IP, Browser, ..." which is kind of overkill for this rough overview list.

PR #111, opened 2016-08-02
give more info about logo size

joergmschulz:
should add info about graphic format for logo and for background of login page.


mar1u5:
Could you confirm this @jancborchardt :wink:

PR #113, opened 2016-08-04
reorder pages 

### LINKED ISSUE #112: Order of pages wrong?
Hi

A user has a problem with the documentation (https://help.nextcloud.com/t/installation-wizard-but-how/2059) and I confirm it's a bit weird. 

So we have this page which explains the installation wizard https://docs.nextcloud.com/server/9/admin_manual/installation/installation_wizard.html , but like the user says, there is nothing about the installation of Nextlcoud before this page. (Like downloading the tarball, extracting etc). I think this is caused because the `Preferred Linux Installation Method` page was removed. (Since we don't provide packages). 
Thus  the prefered way of installing Nextcloud is actually (for now) the manual installation https://docs.nextcloud.com/server/9/admin_manual/installation/source_installation.html .

Shouldn't this page (about the manual installation) be moved before the installation wizard, and before the pages about `installation apps`? Maybe this page should be renamed too?

To be clear with the order of the pages I mean the order they are presented on this page: https://docs.nextcloud.com/server/9/admin_manual/installation/index.html

hagengraf:
fixes #112 
@mar1u5 @LEDfan

### DIFF
--- a/admin_manual/installation/index.rst
+++ b/admin_manual/installation/index.rst
@@ -7,11 +7,11 @@ Installation
 
    system_requirements
    deployment_recommendations
+   source_installation
    installation_wizard
    command_line_installation
    apps_management_installation
    apps_supported
-   source_installation
    appliance_installation
    php_54_installation
    php_55_installation
--- a/admin_manual/installation/index.rst
+++ b/admin_manual/installation/index.rst
@@ -12,7 +12,6 @@ Installation
    command_line_installation
    apps_management_installation
    apps_supported
-   appliance_installation
    php_54_installation
    php_55_installation
    selinux_configuration


mar1u5:
I think that fixes the problem :sweat_smile: ... :+1:

admin_manual/installation/index.rst
```diff
@@ -7,11 +7,11 @@ Installation
 
    system_requirements
    deployment_recommendations
+   source_installation
    installation_wizard
    command_line_installation
    apps_management_installation
    apps_supported
-   source_installation
    appliance_installation
```
nickvergessen: since we don't have appliance atm comment this out?

LEDfan:
Great thanks!

PR #126, opened 2016-08-13
Contacts App Overview

hagengraf:


mar1u5:
Great Initiative 🎆 However... We should overthink that 😉 I think it does not make sense to create an overview without any more instructions and tutorials (Create a Contact, Find CardDAV-Links, Create Adressbook...) ... could you please add more and deeper going instructions in the coming days? ;)
 
Small Design thing: Screenshots should have larger resolutions 😁

mar1u5:
I have created a ToDo-List 😉 #127

nickvergessen:
I think finding the carddav link and hooking it up to the users phone or something like that is what most people will do, so I agree that showing the carddav link is the most important thing.

PR #131, opened 2016-08-19
Update google.rst

tacruc:
The URL changed in NC 10


admin_manual/configuration_files/external_storage/google.rst
```diff
@@ -71,6 +71,7 @@ for example ``https://example.com``, without a trailing slash. You need two
 
   https://example.com/nextcloud/index.php/settings/personal
   https://example.com/nextcloud/index.php/settings/admin
+  https://example.com/nextcloud/settings/admin/externalstorages
```
MorrisJobke: Could you add `/index.php` before `/settings`. Then the URL also works on instances that doesn't have shortened URLs 😉 

Thanks

tacruc: I added it, but I think both versions are needed.

MorrisJobke:
Thanks :)

PR #133, opened 2016-08-22
Admin manuel -> nginx config: Fixed fastcgi_split_path_info regex

zyberspace:
While installing nextcloud on my server with nginx i ran into the issue of not being able to open the installation wizard because the regex for the fastcgi_split_path_info didn't catch the installation url (`/index.php/`). The previous regex required something behind the `/`, which doesn't work for the installation url.

If you access `/index.php/` on an already running nextcloud version it just redirects you to `/apps/files/` so that's probably the reason it never was a problem.

### DIFF
--- a/admin_manual/installation/nginx_examples.rst
+++ b/admin_manual/installation/nginx_examples.rst
@@ -277,7 +277,7 @@ Add *inside* the ``server{}`` block, as an example of a configuration::
    fastcgi_ignore_headers Cache-Control Expires Set-Cookie;
    
    location ~ \.php(?:$/) {
-         fastcgi_split_path_info ^(.+\.php)(/.+)$;
+         fastcgi_split_path_info ^(.+\.php)(/.*)$;
        
          include fastcgi_params;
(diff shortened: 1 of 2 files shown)


MorrisJobke:
cc @josh4trunks

josh4trunks:
Agreed, we already implemented this fix in ownCloud.
Please also change it for the subdir example and I'll approve.

zyberspace:
Nice, will add the commit tomorrow then.

nickvergessen:
@zyberspace want to do this this weekend? Otherwise we can also take care of that.

zyberspace:
Hi, and sry, yesterday was a busy day and i simply forgot about it. I now added the change to the subdir example and the 'nginx_examples.rst` file. It's one commit to keep the history clean.

nickvergessen:
:+1: thanks and no problem :wink:

nickvergessen:
Should backport to 10 and 9?

josh4trunks:
yeah, this should go in anything version >=9

MorrisJobke:
Backported :)

PR #161, opened 2016-09-05
Add some docs for developers wanting to the OCS and the AppFramework

rullzer:
CC: @MorrisJobke @nickvergessen @LukasReschke 

I doubt it is used much. But better to update the docs a bit.


nickvergessen:
:+1:

nickvergessen:
Is this in 10 as well?

nickvergessen:
Also please post in https://help.nextcloud.com/c/dev/app-dev

rullzer:
so a part of it is in 10. But some stuff is missing that I don't want to backport. So I'd say leave it only in 11 and don't hit at 10 devs that they have to change already.

PR #171, opened 2016-09-10
Correcting webdav path for NC10

SlavikCA:


MorrisJobke:
The `/webdav` one is the old path which acts like a link to this new URL. I'm still not sure if we should go for the easier link or if this is even needed at all.

cc @icewind1991 @LukasReschke

SlavikCA:
I see.
Would be nice to see some consistency.

Currently, the `/webdav` is used in the screenshots in the docs.

Also, `/webdav` is given to me, when I go to Settings -> WebDAV in my NextCloud system

MorrisJobke:
> Currently, the /webdav is used in the screenshots in the docs.
> 
> Also, /webdav is given to me, when I go to Settings -> WebDAV in my NextCloud system

Yes. They will work both. So I don't see currently why we should change them. Sorry that this was pending for that long. Thanks for your help anyways

PR #174, opened 2016-09-13
fix rewrite and broken error pages

RubenKelevra:


MorrisJobke:
cc @josh4trunks

josh4trunks:
@RubenKelevra can you explain why you are making a change to the rewrite?

I would expect if you have `rewrite ^ /index.php;` as you suggest pretty URLs would not work. For example if I browse to `/login`,`/index.php` execute without `/login` in the `PATH_INFO`?

Also if you make any suggestions please always implement it in both 'Nextcloud in the webroot of nginx' and 'Nextcloud in a subdir of nginx'

RubenKelevra:
Yep: using this in my setup leads to 503 errors in Chrome under Linux and Safari under iOS. While Firefox is fine. In any cases this happens while just start typing the url and hit enter to let autocomplete make the rest.

I've debugged them and found a call of the 403 error page did not work, this 403 seems to be caused by the changed rewrite-rule.

Feel free to suggest other changes, I don't found any.

Currently I'm just using NextCloud in a webroot and don't plan to use it in a subdir I cannot test anything with webdirs. So feel free to adopt my changes to webroots as well after testing.

josh4trunks:
I don't understand exactly what you are saying is broken? 403 errors aren't properly being given from Nextcloud or NGINX?
Also, earlier I was asking you about the `/index.php` rewrite and I don't believe you explained that to me.

I just tested browsing to a forbidden URL /.htaccess on my production system and am properly served the 403.php page.
image
404's actually would rarely be served because in most cases of a file not existing /index.php is executed, but if someone wants to test this I think they could try the URL `/updater/missing.file`.

----

I have an feeling any issues you are hitting are because wrong configuration. If you post your nginx configuration I can take if I spot anything.

RubenKelevra:
The index-rewrites seems to give a 403 back sometimes, and the 403-error config which I commented out are not processable by nginx. So nginx just give a 503 back to the client without any error-page.

josh4trunks:
@RubenKelevra I cannot reproduce the issue you are experiencing and therefor do not agree with the 'fixes' that you are recommending.

MorrisJobke:
> @RubenKelevra I cannot reproduce the issue you are experiencing and therefor do not agree with the 'fixes' that you are recommending.

Here too. The documented config works fine here without the issues you are describing, @RubenKelevra

RubenKelevra:
The documented config does not work out of the box, else I would not recommend this changes, don't I?

Which version of nginx and php are you using?

PR #207, opened 2016-10-14
strict perms bash script update & related instruction

affinityv:
- automates selection of standard htuser:htgroup for various distros
- automates selection of Nextcloud installation & data paths
- upgrade mode added (for less strict perms)

### DIFF
--- a/admin_manual/installation/installation_wizard.rst
+++ b/admin_manual/installation/installation_wizard.rst
@@ -137,8 +137,7 @@ use :ref:`label-phpinfo` (Look for the **User/Group** line).
 .. note:: When using an NFS mount for the data directory, do not change its 
    ownership from the default. The simple act of mounting the drive will set 
    proper permissions for Nextcloud to write to the directory. Changing 
-   ownership as above could result in some issues if the NFS mount is 
-   lost.
+   ownership as above could result in some issues if the NFS mount is lost.
 
 The easy way to set the correct permissions is to copy and save this script to
 a suitable path location, perhaps ``/usr/local/bin/`` - then run the script
@@ -146,6 +145,7 @@ from within your Nextcloud installation directory.  As per the comments in
 the script, you may setup additional environment variables on the command line
 if required.  The script can also be used to prepare for less strict
 permissions as required to do an upgrade::
+
 
  #!/bin/bash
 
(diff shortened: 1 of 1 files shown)


admin_manual/installation/installation_wizard.rst
```diff
@@ -140,50 +140,210 @@ use :ref:`label-phpinfo` (Look for the **User/Group** line).
...
+
+ echo "Creating possible missing Directories"
+ do_cmd mkdir -p "${nc_path}"/{assets,updater} "${nc_data}"
+
+ echo "chmod Files and Directories"
+ do_cmd find "${nc_path}"/ "${nc_data}"/ -type f -exec chmod 0640 {} \;
+ do_cmd find "${nc_path}"/ "${nc_data}"/ -type d -exec chmod 0750 {} \;
+
+ echo "chown Directories"
+ do_cmd chown -R 0:${config[htgroup]} "${nc_path}"/
```
affinityv: Using `0` as `root` user, this will always work and it will **always** be root anyway.

admin_manual/installation/installation_wizard.rst
```diff
@@ -140,50 +140,210 @@ use :ref:`label-phpinfo` (Look for the **User/Group** line).
...
+     exit ${error_exit}
+ }
+
+
+ #
+ # Main
+ #
+
+ # Default DocumentRoot
+ DefaultDocumentRoot=/var/www/owncloud
```
schnello: Hello.

Should be:
DefaultDocumentRoot=/var/www/nextcloud

affinityv: Okay, adjusted script.  Now doesn't use `DefaultDocumentRoot` variable, but it does check for occ in both `/var/www/nextcloud` then `/var/www/owncloud` locations to try and find `occ` file.

affinityv:
If it is preferred, I'll merge the two commits into one.

**Note:** also that `/var/www` is also distro specific... which can also be fixed if necessary.

jospoortvliet:
@LukasReschke @MorrisJobke you two know about this stuff ;-)

MorrisJobke:
We changed this whole section completely - see #431

The strict permissions script is not available anymore.

PR #220, opened 2016-10-27
Make sure mbstring, xml and zip are installed

nickvergessen:
@MorrisJobke @MariusBluem 

Maybe for Nextcloud 11 we should update the manual to Ubuntu 16.04 LTS + php 7.0


MorrisJobke:
> Maybe for Nextcloud 11 we should update the manual to Ubuntu 16.04 LTS + php 7.0

:+1:

PR #225, opened 2016-11-09
Improve reache-ablity of settings chapter

LiamHD:
The settings chapter was only reachable via the forward / back navigation at the top. It was not linked from inside the index page.


MorrisJobke:
👍

PR #226, opened 2016-11-10
Add documentation for enabling and disabling users via the Nextcloud provisioning API

benediktg:
Since Nextcloud 10 it is possible to en- and disable users with `occ` as well as the Nextcloud provisioning API but the latter information was missing in the docs.


MorrisJobke:
cc @rullzer

MorrisJobke:
👍

MorrisJobke:
backport to stable10: #228 

Thanks @benediktg ;)

PR #230, opened 2016-11-15
Remove owncloud reference from Backport docs

eppfel:
removed Thomas, updated example branches and removed mailing list info.

I don't know, if this reflects the current process, but at least I removed the parts, which are false.

@MorrisJobke @jospoortvliet


MorrisJobke:
:+1: Makes sense

PR #234, opened 2016-11-18
Add a warning for a correct restore

nickvergessen:
@MorrisJobke

### DIFF
--- a/admin_manual/maintenance/restore.rst
+++ b/admin_manual/maintenance/restore.rst
@@ -30,6 +30,38 @@ install and data folder) to your Nextcloud environment. You could use this comma
 Restore Database
 ----------------
 
+Clean Database Before Restoring
+===============================
+
+.. warning:: Before restoring a backup you need to make sure to delete all existing database tables.
+
+The easiest way to do this is to drop and recreate the database.
+SQLite does this automatically.
+
+MySQL
+^^^^^
+
+MySQL is the recommended database engine. To restore MySQL::
+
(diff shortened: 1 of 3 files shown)


admin_manual/maintenance/restore.rst
```diff
@@ -30,6 +30,38 @@ install and data folder) to your Nextcloud environment. You could use this comma
 Restore Database
 ----------------
 
+Clean Database before restoring
```
ChristophWurst: From looking at the other headings I assume we should use caps for all words: "Clean Database Before Restoring"

admin_manual/maintenance/restore.rst
```diff
@@ -30,6 +30,38 @@ install and data folder) to your Nextcloud environment. You could use this comma
...
+
+MySQL
+^^^^^
+
+MySQL is the recommended database engine. To restore MySQL::
+
+   mysql -h [server] -u [username] -p[password] -e "DROP DATABASE nextcloud"
+   mysql -h [server] -u [username] -p[password] -e "CREATE DATABASE nextcloud"
+
+If you use UTF8 with multibyte support (e.g. for emoijs in filenames), use::
```
ChristophWurst: is it possible to restore a non-multibyte backup into a multibyte database?

nickvergessen: Jup

admin_manual/maintenance/restore.rst
nickvergessen: Fixed

PR #237, opened 2016-11-22
Add docs about MySQL 4-byte support

nickvergessen:
@MorrisJobke

### DIFF
--- a/admin_manual/maintenance/mysql_4byte_support.rst
+++ b/admin_manual/maintenance/mysql_4byte_support.rst
@@ -1,0 +1,31 @@
+=============================
+Enabling MySQL 4-byte support
+=============================
+
+.. note:: This feature is currently **experimental**.
+
+In order to use Emojis on your Nextcloud server with a MySQL database, the
+installation needs to be tweaked a bit.
+
+1. Update your Nextcloud server to Nextcloud 11 or later.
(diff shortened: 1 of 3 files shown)


admin_manual/maintenance/mysql_4byte_support.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    [mysqld]
+    innodb_large_prefix=true
+    innodb_file_format=barracuda
+    innodb_file_per_table=true
+
+3. Restart the MySQL server in case you changed the configuration in step 2.
+4. Change your databases character set and collation::
+
+    ALTER DATABASE nextcloud CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
```
LukasReschke: This won't do. Also as per http://dba.stackexchange.com/questions/8239/how-to-easily-convert-utf8-tables-to-utf8mb4-in-mysql-5-5

nickvergessen: This only converts the database, tables are converted with the repairstep in Step 6.

I don't know where I got `CONVERT TO` from, will fix.

LukasReschke:
Tested. Seems to work. 👍

PR #263, opened 2016-12-09
Remove note about  mysterious 8.2.1

blizzz:
Might makes sense in Nc 9 perhaps, but can also be removed in the 10 docs I think.


MorrisJobke:
:+1:

PR #272, opened 2016-12-14
Update user_provisioning_api.rst

schiessle:
Mention that the content type "application/x-www-form-urlencoded" is needed for POST request, see https://secure.php.net/manual/en/reserved.variables.post.php

cc @nickvergessen


nickvergessen:
As discussed 👍

PR #276, opened 2016-12-14
Update devenv.rst

jospoortvliet:
Just went through it and this was one step with a mistake...

### DIFF
--- a/developer_manual/general/devenv.rst
+++ b/developer_manual/general/devenv.rst
@@ -54,7 +54,7 @@ After the development tool installation make the directory writable::
 
 Then install Nextcloud from git::
 
-  git clone git@github.com:nextcloud/server.git --branch $BRANCH /var/www/<folder>
+  git clone git@github.com:nextcloud/server.git /var/www/<folder>
   cd /var/www/<folder> 
   git submodule update --init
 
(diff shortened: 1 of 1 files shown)


developer_manual/general/devenv.rst
```diff
@@ -55,9 +55,10 @@ After the development tool installation make the directory writable::
 Then install Nextcloud from git::
 
   git clone git@github.com:nextcloud/server.git --branch $BRANCH /var/www/<folder>
+  cd /var/www/<folder>/server  
   git submodule update --init
 
-where <folder> is the folder where you want to install Nextcloud.
+where <folder> is the folder where you want to install Nextcloud. Occasionally, the git submodule command gives errors, it might work to just try again. Magic!
```
MorrisJobke: Ummm ... no? There should be no error. What was the exact error?

developer_manual/general/devenv.rst
```diff
@@ -55,9 +55,10 @@ After the development tool installation make the directory writable::
 Then install Nextcloud from git::
 
   git clone git@github.com:nextcloud/server.git --branch $BRANCH /var/www/<folder>
+  cd /var/www/<folder>/server
```
MorrisJobke: `/server` shouldn't be needed here

developer_manual/general/devenv.rst
```diff
@@ -55,9 +55,10 @@ After the development tool installation make the directory writable::
 Then install Nextcloud from git::
 
   git clone git@github.com:nextcloud/server.git --branch $BRANCH /var/www/<folder>
```
MorrisJobke: `-branch $BRANCH` is not needed. Master is used by default and then git can do the rest.

MorrisJobke:
stable11: #281

PR #284, opened 2016-12-16
Remove reference to not existing file

MorrisJobke:
Ref https://github.com/nextcloud/server/issues/2686#issuecomment-267335756


MorrisJobke:
cc @nickvergessen

nickvergessen:
Should we add something like check core/l10n/ instead? Otherwise someone might add "Deutsch"?

MorrisJobke:
I would say that we should list the underlying ISO standard and a link to the Wikipedia page of it.

nickvergessen:
That's awesome

nickvergessen:
Backport to stable11

MorrisJobke:
> Backport to stable11

#302

PR #300, opened 2017-01-04
Remove APC in caching settings since php >= 5.6

tflidd:
With NC 11, we are only supporting PHP 5.6+. So all old references on the caching page referring to APC and other settings for PHP <5.6 can be removed.


nickvergessen:
Thanks, had the same thing in mind but didn't see your PR before starting it.
So I took over remaining changes in #301

PR #313, opened 2017-01-16
Fix style

senden9:
Add missing grave accent to fix the highlighting.


MorrisJobke:
Hi @senden9 - sorry to say that, but this is generated from the config.php in the server repo. Could I ask you to do the change there? 

Thanks a lot for spotting this :)

https://github.com/nextcloud/server/blob/master/config/config.sample.php#L541

senden9:
Thanks for linking to the right file. I reported it at nextcloud/server#3142 & nextcloud/server#3143.

PR #314, opened 2017-01-16
Fix style

senden9:
Add missing grave accent to fix the highlighting. Same as PR #313 but for the stable11 branch.


MorrisJobke:
See https://github.com/nextcloud/documentation/pull/313#issuecomment-273004762

PR #323, opened 2017-01-20
Explicitly type $userManager

alantygel:
Hi, I'm getting started on NC apps. On my app, the dependency injection only worked after declaring the type of UserManager explicitly.

I'm not 100% sure, but it worked here.

best,
alan

### DIFF
--- a/developer_manual/app/hooks.rst
+++ b/developer_manual/app/hooks.rst
@@ -48,12 +48,13 @@ The hook logic should be in a separate class that is being registered in the :do
 
     <?php
     namespace OCA\MyApp\Hooks;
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> I'm not 100% sure, but it worked here.

Yes - this is correct. Otherwise the variable needs to be named `$iUserManager` to properly detect the type. But you should always type hint the variable type. This makes your code more save and it's faster from a dependency injection point of view.

Perfect 👍 

If you need help don#t hesitate to ask in our IRC channel #nextcloud-dev on freenode ;)

Cheers

MorrisJobke:
I also ported it to stable11

PR #328, opened 2017-01-24
Remove documentation about multiple smtp hosts

MorrisJobke:
* it doesn't work for a long time

see https://github.com/nextcloud/server/issues/2271

cc @lukasreschke @nickvergessen


nickvergessen:
:+1: :see_no_evil:

Issue #54, opened 2016-07-10
Proper documentation needed for pretty URLs


oparoz:
## Missing doc

This is the only paragraph found in [the documentation](https://docs.nextcloud.org/server/9/admin_manual/installation/source_installation.html#pretty-urls):

> Pretty URLs are created automatically when .htaccess is writable by the HTTP user, mod_env and mod_rewrite are installed, and 'overwrite.cli.url' in your config.php is set to any non-null value.

This means that if your setup wasn't a 100% as expected by Nextcloud before creating the admin account, you will have missed your chance to setup pretty URLs, because there isn't any documentation available on how to set it up manually.
## Config sample is unclear

From the [config sample](https://github.com/nextcloud/server/blob/master/config/config.sample.php#L376-L399)

> - Note that above rule is not valid in every case, there are some rare setup
>   - cases where this may not apply. However, to avoid any update problems this
>   - configuration value is explicitly opt-in.

This means that unlike all other parameters, this needs to be added to config.php. 
I think this should be the first line in that large comment and the setting should be commented out since it's not working when found in the sample.

---

@MorrisJobke

MorrisJobke:
cc @LukasReschke

MorrisJobke:
> The most obvious one being that the theme you've created with the built-in editor will be wiped.

Really? Why?

MorrisJobke:
> There is no mention anywhere about the side effects of switching.

There are none AFAIK. Only when switching back some URLs won't work.

oparoz:
>>There is no mention anywhere about the side effects of switching.

>There are none AFAIK. Only when switching back some URLs won't work.

I've removed that section. It's not a side-effect, just that theming is currently not a 100% compatible with pretty URLs

MorrisJobke:
> I've removed that section. It's not a side-effect, just that theming is currently not a 100% compatible with pretty URLs

Could you open an issue for this in the server repo? This is definitely a bug and should be addressed instead of documented ;)

oparoz:
It's [there](https://github.com/nextcloud/server/issues/315) already and I just have to create the PR to fix the problem.

MorrisJobke:
> It's there already and I just have to create the PR to fix the problem.

Awesome. I also just found it :) 👍

tflidd:
A user just found out how to implement this on nginx:
https://help.nextcloud.com/t/pretty-urls-for-nextcloud-11-with-nginx-and-php5/11921

[...]

PR #338, opened 2017-01-27
admin_manual/source_installation: Fixes in PHP prerequisites and Apac…

pmattern:
(Please see commit message.)

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -56,9 +56,9 @@ Required:
 
 Database connectors (pick the one for your database:)
 
-* PHP module sqlite (>= 3, usually not recommended for performance reasons)
+* PHP module pdo_sqlite (>= 3, usually not recommended for performance reasons)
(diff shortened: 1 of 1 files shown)


admin_manual/installation/source_installation.rst
```diff
@@ -270,7 +270,7 @@ Additional Apache Configurations
 Pretty URLs
 -----------
 
-Pretty URLs are created automatically when ``.htaccess`` is writable by the 
+Pretty URLs are created automatically when ``.htaccess`` is readable by the
```
MorrisJobke: That is incorrect. The HTTP user needs to write the htaccess because we will add some more details there.

pmattern:
@MorrisJobke 
Fixed. I wasn't fully aware of the way Pretty URLs are dealt with, which has quite a bit to do with #54.
Also, at first glance it seems a bit odd that you seem to intend to have `./.htaccess` writable by the web server user all the time. Different story, of course.

PR #343, opened 2017-01-30
Small typo

mcm007:
Small typo in "Example Installation on Ubuntu 16.04 LTS Server" section.
php-imagick > php7.0-imagick


mcm007:
Indeed, it may be the dotdeb repository that i used on Debian 8.7.

tflidd:
Let's keep the documentation with Ubuntu/Debian packages names. If you use other repos, you will have to search in some cases like this.

PR #349, opened 2017-02-03
Add custom data path to strong permissions script in installation wizard

### LINKED ISSUE #348: Set strong permissions script assumes data path
On the installation wizard page in the script that sets strong directory permissions, the data path is hard coded although this is something that can be configured differently. I personally use a different data path because of the way I structured my backup scheme, and so I needed to modify this script to account for that. Basically I just added another variable for the data path, and updated all hard coded paths to data to use the variable instead. I'll send a pull request soon.

linucksrox:
Resolves #348


MorrisJobke:
Thanks for your contribution 👍

Issue #334, opened 2017-01-26
Favorite API


MorrisJobke:
We need to document the favorite API to use it in the mobile apps (cc @tobiasKaminsky @AndyScherzinger @mario @marinofaggiana )

Something like that seems to be the API for now:

They can be exposed from within WebDAV. During the PROPFIND for a folder just add `<oc:favorite />` to the request body:

```xml
<?xml version="1.0"?>
<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns">
 <d:prop>
   <d:getlastmodified />
   <d:getetag />
   <d:getcontenttype />
   <d:resourcetype />
   <oc:fileid />
   <oc:permissions />
   <oc:size />
   <d:getcontentlength />
   <nc:has-preview />
   <oc:tags />
   <oc:favorite />
   <oc:comments-unread />
   <oc:owner-display-name />
   <oc:share-types />
 </d:prop>
</d:propfind>
```

Then the properties of that file contain: `<oc:favorite>1</oc:favorite>` if the file is a favourite or nothing it not.

Beside that there is a general API to get all files: `https://example.org/apps/files/api/v1/tags/_%24!%3CFavorite%3E!%24_/files`

This then returns a JSON object like here: 

```json
{"files":[{"id":23,"parentId":-1,"mtime":1484799137000,"name":"asd","permissions":9,"mimetype":"httpd\/unix-directory","size":163,"type":"dir","etag":"58803ca182cac","shareOwner":"abc@localhost\/server\/","isShareMountPoint":true,"mountType":"shared-root","path":"\/","tags":["_$!\u003CFavorite\u003E!$_"]}]}
```

@nickvergessen @schiessle @rullzer Can somebody of you please properly document this?

cc @karlitschek

MorrisJobke:
> @nickvergessen @schiessle @rullzer Can somebody of you please properly document this?
> 
> cc @karlitschek

Ping. @marinofaggiana really needs this.

marinofaggiana:
Yes, urgent please, GET & POST

nickvergessen:
I have no idea, I just know that it's somewhere in dav. Maybe you can have a look at the requests the browser sends when you favor a file in the web UI

mario:
You get favorites just like @MorrisJobke mentioned, and here's how to set/unset them:


Set favorite 

POST /apps/files/api/v1/files/path/to/file 

[...]

PR #351, opened 2017-02-09
document client api's

icewind1991:
We should have some proper documentation for the various client api's we have (ocs and the dav's).

To start out here is an overview of some of the things that a client can do over WebDAV.

The intend of this doc is not to provide a full documentation of all the webdav stuff (that's what the rfc's are for) but to give an aspiring developer an overview of the capabilities our webdav implementation has an some pointers to more info/how to get started.

This is especially useful for things like tags, favorites, search, etc. Where while we do nicely follow the dav rfc's, the actual logic is not determined by any rfc but custom to our dav implementation.

cc @mario @MorrisJobke @nickvergessen

Fixes https://github.com/nextcloud/documentation/issues/334

### DIFF
--- a/developer_manual/client_apis/WebDAV/index.rst
+++ b/developer_manual/client_apis/WebDAV/index.rst
@@ -1,0 +1,242 @@
+.. _webdavindex:
+
+===============================
+WebDAV Client API's
+===============================
+
+This document provides a quick overview of the WebDAV operations supported in Nextcloud, to keep things readable it won't go into many details
+for each operation, further information for each operation can be found in the corresponding rfc where applicable
+
+----
+WebDAV Basics
+----
+
+The base url for all WebDAV operations for a Nextcloud instance is :code:`/remote.php/dav`.
+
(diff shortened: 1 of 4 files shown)


developer_manual/client_apis/WebDAV/index.rst
```diff
@@ -0,0 +1,242 @@
...
+			 <oc:favorite>1</oc:favorite>
+		 </oc:filter-rules>
+	 </oc:filter-files>
+
+File properties can be requested by adding a :code:`<d:prop/>` element to the request listing the requested properties in the same way as it would be done for a :code:`PROPFIND` request.
+
+When listing favorites, the request will find all favorites in the folder recursively, all favorites for a user can be found by sending the request to :code:`remote.php/dav/files/user`
+
+----
+Settings tags
```
nickvergessen: Since these are not the "visible collaborative" tags, but user depending tags which have no UI anywhere, I wouldn't document them.

icewind1991: Yeah, figured that one out while trying to figure out how to search by tag

PR #368, opened 2017-02-27
Remove invalid Shibboleth documentation

LukasReschke:
This is for the old Shibboleth app and not the new fancy one we have.


MariusBluem:
Please backport this 👊

PR #40, opened 2016-07-01
Fix order of Nginx config options

MorrisJobke:
See https://github.com/nextcloud/server/issues/237#issuecomment-229923262

PR #34, opened 2016-06-22
Changes for Nextcloud

JSoko:
Changes for Nextcloud

PR #371, opened 2017-03-01
Fix Nginx config

MorrisJobke:
* see https://github.com/nextcloud/server/issues/237#issuecomment-229923262
* redo of #40 because it was reverted in #34 

cc @josh4trunks @tx7


MorrisJobke:
stable11 53c115b
stable10 59290ad

PR #383, opened 2017-03-11
Update manual_upgrade.rst

Nono-m0le:
as per the archive type modification (it's now a .zip):

tar xjf *.bz2 => unzip *.zip


MariusBluem:
There is both currently I think 😶 @MorrisJobke

MorrisJobke:
Yep there are both, but in the description we say: "download from nextcloud.com/install/" which only links to the zip files. So this is correct 👍 

Thanks

MorrisJobke:
stable11 87f695e

PR #385, opened 2017-03-13
Add note about security scanner to documentation

jospoortvliet:
We could tell people we have the scanner so they can use it.

No idea if we ever again would scan systems out there but we should tell people it is possible. Of course, various (both good and bad intentioned) parties can and do scan the web all the time, so this isn't exactly news to any user who knows how the web works :(


MorrisJobke:
blubb

PR #387, opened 2017-03-14
Fix pdf viewer cannot load l10n files bug

mayswind:
PDF.js use .properties file as l10n file format, but this configuration is missing redirecting this type to correct path.


tflidd:
There are some conflicts, I suppose this file does not exist any more in the current repo. @mayswind can you make the changes to the current nginx-config-page (admin_manual/installation/nginx.html) and update this pull request?

Issue #291, opened 2016-12-29
Outdated install instructions for CentOS regarding php 5.6


tflidd:
lamixer reported this on the forum (https://help.nextcloud.com/t/update-docs-re-php-and-centos/6656):

At this page https://docs.nextcloud.com/server/11/admin_manual/installation/php_55_installation.html you kindly explain how to get PHP5.5 on Centos 7. I'm bringing up a virgin installation of Nextcloud 11 (the current stable release) on Centos 7 1611 just installed.

It is a bit frustrating to go through your installation docs, get PHP5.5.x working at some difficulty (5.4.x being the 'normal' one) only to find upon trying to configure Nextcloud that PHP 5.6 is required!

Please update your docs for Nextcloud 11 to give help/instructions to install a version of PHP on Centos 7 which WILL ACTUALLY WORK! Or, if that is too difficult to advise in short order, then at least update that page of the docs to say you must not install the OS default PHP 5.4 but you must seek out a solution for yourself that yeilds PHP 5.6 +.

Thanks. I wasn't expecting this frustration so early in the install process. First time with Nextcloud, as a former Owncloud user.

tflidd:
Duplicate of https://github.com/nextcloud/documentation/issues/140

MorrisJobke:
Comment by @oparoz:

> Adapt this page to PHP 5.6
https://docs.nextcloud.com/server/10/admin_manual/installation/php_55_installation.html

LukasReschke:
Or if one goes crazy: PHP 7.0 :)

https://access.redhat.com/errata/RHEA-2016:2730

oparoz:
Finally... 
We can recommend Red Hat again :)

xMartin:
While you're at it: https://docs.nextcloud.com/server/11/admin_manual/installation/php_54_installation.html

lamixer:
It's a bit sad no action has been taken on this. Imagine you are an admin of limited experience and you are following the Nextcloud installation manual here: https://docs.nextcloud.com/server/11/admin_manual/contents.html and you get to the section on PHP, follow the instructions, and it doesn't work. 

This is not an 'enhancement' to the docs but a correction, as one assumes following the instructions will yield a working result. Can it at least get assigned to someone? 

[...]

PR #388, opened 2017-03-14
PHP upgrade for 5.6 and 7.0 on RHEL/CentOS

michag86:
Fix fox #291

I have tested PHP 7 on CentOS 7. 
The path informations for PHP 5.6 are copied from the rpms from 
http://mirror.centos.org/centos/6/sclo/x86_64/rh/rh-php56/
http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-php56/

Can somebody check this?
@tflidd @lamixer


MorrisJobke:
@oparoz Weren't you interested in this lately?

oparoz:
Thanks for the heads up @MorrisJobke :)

oparoz:
I've tested the instructions and the biggest problem is that it's aimed at people already running PHP and Apache.
Those are upgrade instructions and not installation instructions.

michag86:
@oparoz What do you think about changing "Disable loading the old PHP 5.3 Apache module" to "Disable loading the old PHP 5.3 Apache module (only required if php is already installed)"?

MorrisJobke:
@oparoz @karlitschek Would be nice to get some testing on this. At least we say to support this and should be happy about this nice writeup 😉

tflidd:
This topic is stuck, would be a pity to lose this contribution. I asked on the forum for CentOS users to help us reviewing this issue:
https://help.nextcloud.com/t/php-7-on-centos-install-instructions-are-updated-review-needed/27732

michag86:
I've tested and used 5.6.
I can also add the instructions for PHP 7.1.

KB7777:
Wouldn't be better to use remi repo for PHP7.x for CentOS7?
Installation is very easy -- https://help.nextcloud.com/t/what-is-the-supported-php7-installation-method-for-centos-7/9832/25?u=therion7777

Works fine at my NC instance from NC11-NC13 versions for few k users.

oparoz:
I agree that using Remi's repo would be the more universal solution.

michag86:
https://wiki.centos.org/de/AdditionalResources/Repositories/
Software Collections (SCL) is listes as Additional CentOS Provided Repositories and Remi is listed as Community Approved Repositories, so I decided to use SCL.

The manual for updating to newer php version for centos/rhel was also referencing to scl.
I think SCL is the "supported" source for newer php versions on rhel/centos.

tflidd:
It's now over a year. It's really missing in the documentation and there were no principal objections. And if it is really a matter of personal preference, we could perhaps link to a place with some comparison of the different ways to handle newer php versions e.g. https://www.hogarthuk.com/?q=node/15 (found on the page @michag86 linked)

PR #392, opened 2017-03-20
No more 403 and 404 redirecting

nickvergessen:
Seems to be causing many issues lately:
Fix https://github.com/nextcloud/server/issues/3847

There is no way we can notify people about such changes, right?

### DIFF
--- a/admin_manual/installation/nginx_nextcloud_9x.rst
+++ b/admin_manual/installation/nginx_nextcloud_9x.rst
@@ -84,9 +84,6 @@ your nginx installation.
       # This module is currently not supported.
       #pagespeed off;
   
-      error_page 403 /core/templates/403.php;
-      error_page 404 /core/templates/404.php;
-  
       location / {
           rewrite ^ /index.php$uri;
       }
@@ -221,10 +218,7 @@ your nginx installation.
           # Uncomment if your server is build with the ngx_pagespeed module
           # This module is currently not supported.
           #pagespeed off;
-  
-          error_page 403 /nextcloud/core/templates/403.php;
-          error_page 404 /nextcloud/core/templates/404.php;
-  
+
           location /nextcloud {
               rewrite ^ /nextcloud/index.php$uri;
           }


MorrisJobke:
cc @josh4trunks but for me this looks okay, because we render a proper HTML page in such cases anyway.

josh4trunks:
This should also be removed in the "Nextcloud in a subdir of nginx" example.

nickvergessen:
I didn't notice it's twice in the docs. fixed now

MorrisJobke:
stable11 6533a9c

josh4trunks:
I had a thought. For the `deny all;` location, those will still be handled by Nginx and never actually be passed to the now internal error page right?

Should we instead forward these all to index.php?

MorrisJobke:
> Should we instead forward these all to index.php?

I would say so.

josh4trunks:
I believe the **.htaccess** does not currently rewrite these, and a 404 would be served by nginx.

Also, I'm wondering why I originally included the following files to be executed directly. I know there was a reason for it, but I don't see reference to them in **.htaccess** anymore.
```
cron.php
core/ajax/update.php
ocs/v1.php
ocs/v2.php
updater/*.php
ocs-provider/*.php
```

MorrisJobke:
> Also, I'm wondering why I originally included the following files to be executed directly. I know there was a reason for it, but I don't see reference to them in .htaccess anymore.

They should be executed directly. The rewrite rules are extended in https://github.com/nextcloud/server/blob/ccd0ded812c2039c14293b78310f144f1b5d7538/lib/private/Setup.php#L459-L485

josh4trunks:
ahh great, I'll review that a bit

josh4trunks:
That is where the static files were referenced from!
`(css|js|svg|gif|png|html|ttf|woff|ico|jpg|jpeg)`

But... I do not think Nextcloud serves anything but **svg**, **woff**, and occasionally a **png**. I did see a **ttf** file in Nextcloud's source but I'm not sure if it's ever served.

MorrisJobke:
> That is where the static files were referenced from!
> (css|js|svg|gif|png|html|ttf|woff|ico|jpg|jpeg)

A right ... they were still in the cache. Let me revert the PR.

MorrisJobke:
> A right ... they were still in the cache. Let me revert the PR.

And then you can create a new one with all the stuff (including the TTL for the cache)

josh4trunks:
I don't fully understand. Is Nextcloud setting a TTL / **max-age** for these somewhere?

MorrisJobke:
> I don't fully understand. Is Nextcloud setting a TTL / max-age for these somewhere?

I meant: https://github.com/nextcloud/server/blob/bff6c8aafc49a3260294c3244571e3a31fd09cca/.htaccess#L24-L24

MorrisJobke:
because I reverted the other PR and there was also the update for the max-age included.

josh4trunks:
Ok, makes sense.
I will do that and also make those regexs not case insensitive. No need to check for uppercase if apache is not.

josh4trunks:
See PR here https://github.com/nextcloud/documentation/pull/468

PR #428, opened 2017-04-25
Documentation for "Handle password expiry in user_ldap"

GitHubUser4234:
Documentation for "Handle password expiry in user_ldap" [#1023](https://github.com/nextcloud/server/pull/1023)


GitHubUser4234:
> The feature is specifically crafted for ppolicy, correct? If so, perhaps it might make sense to state this clearly. Good otherwise.

Thanks for the review! I have revised it a bit as suggested~

PR #430, opened 2017-04-25
Add chmod to update permissions script, resolves #429

### LINKED ISSUE #429: Update Permissions script assumes some permissions
https://docs.nextcloud.com/server/11/admin_manual/maintenance/update.html#setting-permissions-for-updating
In the script provided, it's assumed that all permissions are already set correctly, but only that the owner needs to be changed to the web user to allow the updater to overwrite files. However, in some cases this is not true.

A case in point would be when Nextcloud is installed and configured, and you've already run the strong permissions script. Then later on, the admin decides to install some new app, running as root (oops), so now the owner of these files is root, and some permissions are 444 (not even write permissions for the owner 😢 ). Everything seems to work fine at this point. But now the admin wants to upgrade to the latest version, so he/she runs through the updater permission script (no problem), setting the owner recursively, and then they run into permission problems.

I suggest we simply add the same chmod commands that are in the set strong permissions script to this updater script, so that people can avoid this pain.

linucksrox:
I think this would simplify upgrades for users, it's really more of a double check and doesn't hurt anything.

### DIFF
--- a/admin_manual/maintenance/update.rst
+++ b/admin_manual/maintenance/update.rst
@@ -113,6 +113,9 @@ with the path to your Nextcloud directory, and replace the ``htuser`` and
     htgroup='www-data'
     
     chown -R ${htuser}:${htgroup} ${ocpath}
+	
+	find ${ocpath}/ -type f -print0 | xargs -0 chmod 0640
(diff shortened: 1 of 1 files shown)


MorrisJobke:
We removed that script - for details see #431

Thanks for your help anyways :)

PR #431, opened 2017-04-26
Remove recommendation to change permissions

LukasReschke:
1. The wording here is really off. We don't really recommend this highly.
2. This is even counterproductive as it disallows updating via the app store. Also makes the updater unusable.
3. Once we have automatic updates this will effectively decrease security from my PoV
4. The script that is there had even some bugs and once destroyed the complete root file system…

Also, an attacker can execute code also using other ways once they have write access to Nextcloud or the data folder.

As we get a ton of bug reports about issues with that, let's kill it. I never liked it anyways…

### DIFF
--- a/admin_manual/configuration_files/big_file_upload_configuration.rst
+++ b/admin_manual/configuration_files/big_file_upload_configuration.rst
@@ -162,7 +162,4 @@ To be able to use this input box you need to make sure that:
 * your Web server is able to use the ``.htaccess`` file shipped by Nextcloud (Apache only)
 * the user your Web server is running as has write permissions to the files ``.htaccess`` and ``.user.ini``
 
-:ref:`strong_perms_label` might prevent write access to these files. As an admin you need
-to decide between the ability to use the input box and a more secure Nextcloud installation
-where you need to manually modify the upload limits in the ``.htaccess`` and ``.user.ini``
-files described above.
+
--- a/admin_manual/configuration_files/external_storage/local.rst
+++ b/admin_manual/configuration_files/external_storage/local.rst
@@ -15,9 +15,6 @@ are on Ubuntu Linux::
  sudo -u www-data chown -R www-data:www-data /localdir
  sudo -u www-data chmod -R 0750 /localdir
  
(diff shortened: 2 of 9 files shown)


MorrisJobke:
Backport?

MorrisJobke:
stable11 8edb15a6

MorrisJobke:
stable10 c65b89c6

nachoparker:
yeah, that is actually a good move

nachoparker:
btw, do you have any conversations already about automatic updates? any link?

MorrisJobke:
> btw, do you have any conversations already about automatic updates? any link?

It's still only in our head ... but we want to use our current updater and enhance it a bit.

nachoparker:
are these discussions open? IRC? can I help?

MorrisJobke:
For now they were only temporary in IRCsome time back

MorrisJobke:
Maybe you open a ticket in der updater woo and write down your thoughts and how you think it should work

homelessgamer:
you know that apache users will talk shit about nextcloud's security, cause they dont like to see www-data owned files in their /var/www ?!

linucksrox:
Maybe this is the wrong place to ask, but can you please provide more information or even any external resources about this topic? My understanding was that it's trivial to exploit a website when its application files are writable by the web user, but after a little research I haven't been able to find out how. So does it still require other vulnerabilities for hackers to modify web files, even though they are writable by the web user?

enoch85:
Ehm, it's better to use it than to not use it right? 

Feels like most bug reports comes from that users can't use the updater, and that that is the biggest reason to remove it. Correct me if I'm wrong. The VM uses it, but we also have out own updater written in bash, never had any issues or reports about it.

74cmonty:
Hello!
Does this mean you will not provide any recommendation regarding permissions anymore?

If I don't install NC from source but a package (available for Debian, Alpine Linux, etc.) I depend on the correct permission settings defined by the package provider.

I would appreciate some (official) recommendations from NC to verify at least if the installation from a package provider is correct.

PR #433, opened 2017-04-28
Replace ownCloud Apps Store occurrences

ilpianista:


admin_manual/issues/code_signing.rst
```diff
@@ -47,7 +47,7 @@ Is Code Signing Mandatory For Apps?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Code signing is optional for all third-party applications. Applications
-with a tag of "Official" on apps.owncloud.com require code signing.
+with a tag of "Official" on apps.nextcloud.com require code signing.
```
MariusBluem: With Nextcloud App Store - I think **all** apps (not only official ones) require code-signing. @BernhardPosselt

admin_manual/installation/apps_management_installation.rst
```diff
@@ -42,8 +42,8 @@ page by default.
 
 Click the app name to view a description of the app and any of the app settings in the Application View field.  Clicking the **Enable** button will enable the app.  If the app is not part of the Nextcloud installation, it will be downloaded from the app store, installed and enabled.
 
-Click the gear icon on the lower left to browse experimental apps in the `ownCloud Apps
-Store <https://apps.owncloud.com/>`_. Install experimental apps at your own risk.
+Click the gear icon on the lower left to browse experimental apps in the `Nextcloud App
+Store <https://apps.nextcloud.com/>`_. Install experimental apps at your own risk.
```
MariusBluem: Needs to be rewritten content-wise. `experimental apps` are currently not shipped over Nextcloud App Store I think. @BernhardPosselt

developer_manual/app/code_signing.rst
```diff
@@ -46,7 +46,7 @@ Is Code Signing Mandatory For Apps?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Code signing is optional for all third-party applications. Applications
-with a tag of "Official" on apps.owncloud.com require code signing.
+with a tag of "Official" on apps.nextcloud.com require code signing.
```
MariusBluem: Same here 😅

admin_manual/installation/apps_management_installation.rst
MorrisJobke: There are no experimental apps anymore. We removed that flag. They are just normal apps and if there is something problematic we write the reason to the entry in the apps management (like "This app is not compatible with Nextcloud 12")

admin_manual/issues/code_signing.rst
MorrisJobke: Yes. Only local apps don't need it AFAIK. cc @LukasReschke

PR #438, opened 2017-05-03
remove HAProxy as mysql load balancer

MorrisJobke:
HAProxy is not an option to have in front of the MySQL cluster. It can split READs and WRITEs, but only if your application connects for READs on one port and for WRITEs on a different one. Nextcloud doesn't support this yet.


MorrisJobke:
stable11 fff890c6

MorrisJobke:
stable10 73a6bbb2

PR #445, opened 2017-05-08
Cleanup nginx config for old 403/404 pages

MorrisJobke:
Will not be used directly anymore and always be processed within the index.php. Right @nickvergessen ?

Followup to #392

cc @josh4trunks


josh4trunks:
Awesome, I look forward to testing this with the update.

nickvergessen:
You can also make the change right now. The update will not change the nginx config

PR #463, opened 2017-05-15
Update apps_supported.rst

jospoortvliet:


MariusBluem:
What about the `socialsharing`-Plugins (Facebook, Twitter, Google+ and Diaspora), since it they are inside of an app bundle 🤔 @jospoortvliet

MorrisJobke:
> What about the socialsharing-Plugins (Facebook, Twitter, Google+ and Diaspora), since it they are inside of an app bundle 🤔 @jospoortvliet

Yes - they should be added.

jospoortvliet:
on it.

PR #464, opened 2017-05-15
Add socialsharing_* plugins to apps_supported

MariusBluem:

### DIFF
--- a/admin_manual/installation/apps_supported.rst
+++ b/admin_manual/installation/apps_supported.rst
@@ -38,7 +38,7 @@ Below is the list of apps supported for Nextcloud 12. Supported here means that 
 * Provisioning API
 * Server info (monitoring app)
 * Sharebymail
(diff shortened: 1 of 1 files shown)


admin_manual/installation/apps_supported.rst
```diff
@@ -38,6 +38,7 @@ Below is the list of apps supported for Nextcloud 12. Supported here means that
 * Provisioning API
 * Server info (monitoring app)
 * Sharebymail
+* Socialsharing-plugins (Diaspora, email, Facebook, Google+ and Twitter)
```
MorrisJobke: we should not name them plugins - just social sharing (because the title is "Supported **apps**") 😉

MariusBluem: I understand ... but since this is an open system, there might be other social plugins in the future not supported by GmbH ... but I think this is OK for now 😁

PR #465, opened 2017-05-15
Update apps_supported.rst

jospoortvliet:
improve text above


MorrisJobke:
Could you please remove the version number or replace it with `| version |` otherwise we need to update this and it's likely that we will forget about it.

MorrisJobke:
Added and looks good 👍

PR #491, opened 2017-06-03
Remove hint about maintenance mode

### LINKED ISSUE #490: Upgrade Manually outdated
_Referring to the documentation for [Upgrade Manually](https://docs.nextcloud.com/server/12/admin_manual/maintenance/manual_upgrade.html)._

> Put your server in maintenance mode. This prevents new logins, locks the sessions of logged-in users, and displays a status screen so users know what is happening. There are two ways to do this, and the preferred method is to use the occ command, which you must run as your HTTP user. This example is for Ubuntu Linux:
> 
> `sudo -u www-data php occ maintenance:mode --on`

Currently the upgrade is aborted with the following message if the maintenance mode is already enabled when the upgrade is started.
```
<warning>Nextcloud is in maintenance mode</warning>
Maybe an upgrade is already in process. Please check the logfile (data/nextcloud.log). If you want to re-run the upgrade procedure, remove the "maintenance mode" from config.php and call this script again.
```
While it is not a big issue to disable maintenance mode and restart the upgrade I'd suggest to clarify the documentation that manually enabling maintenance mode is no longer necessary.

Currently I can't make sure but if I recall correctly the maintenance mode is automatically disabled when the upgrade script finishes. Hence the following paragraph shouldn't be needed anymore.
> Assuming your upgrade succeeded, disable the maintenance mode:
> `sudo -u www-data php occ maintenance:mode --off`

MorrisJobke:
* fixes #490

cc @MariusBluem @dimejo

### DIFF
--- a/admin_manual/maintenance/manual_upgrade.rst
+++ b/admin_manual/maintenance/manual_upgrade.rst
@@ -4,17 +4,6 @@ Upgrade Manually
 
 Always start by making a fresh backup and disabling all 3rd party apps.
 
-Put your server in maintenance mode. This prevents new logins, locks the 
-sessions of logged-in users, and displays a status screen so users know what is 
-happening. There are two ways to do this, and the preferred method is to use the 
-:doc:`occ command <../configuration_server/occ_command>`, which you must run as 
-your HTTP user. This example is for Ubuntu Linux::
-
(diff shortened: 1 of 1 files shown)


MorrisJobke:
stable12 5ec35795
stable11 9f97d1fa
stable10 df3306a3

PR #500, opened 2017-06-09
fix saml/login location for Kerberos SSO

jh23453:
According to https://github.com/nextcloud/user_saml/issues/118 and my own experiments the protected path needs to change.
I'm unsure when this changed, but with the change Kerberos SSO works as expected.

### DIFF
--- a/admin_manual/configuration_server/sso_configuration.rst
+++ b/admin_manual/configuration_server/sso_configuration.rst
@@ -13,7 +13,7 @@ The following providers are supported and tested at the moment:
     - Shibboleth
     - Active Directory Federation Services (ADFS)
 - Authentication via Environment Variable
-    - Kerberos (mod_auth_kerb)
+    - Kerberos (mod_auth_gssapi)
     - Any other provider that authenticates using the environment variable
 
 While theoretically any other authentication provider implementing either one of those standards is compatible, we like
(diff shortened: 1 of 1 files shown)


LEDfan:
I can confirm this behavior and the change works for me.

wiswedel:
@schiessle

schiessle:
I don't have a setup to test this currently but it sounds reasonable and it was tested by two independent parties... So let's go for it! :+1:

schiessle:
@kleinem86 Can you name some benefits of mod_auth_gssapi and why we should suggest it instead of mod_auth_kerb. Are they both official Apache modules and available on all distributions?

skjnldsv:
@kleinem86 ? @jh23453  ?
Bump :)

jh23453:
I've moved to SAML in the meantime, so no further test from me right now.
The modules gssapi and krb are available from Debian and CentOS (7), so going with gssapi
seems reasonable.

kleinem86:
Sorry for the late reply, i'm not that active on Github ;)
Cite from mod_gssapi's Intro "This module has been built as a replacement for the aging mod_auth_kerb"

Basically GSSAPI provides an API to provide mutiple different authentication mechanisms.
While mod_auth_krb can only do Kerberos, GSSAPI can also do NTLM or wrap Basic auth if desired - and get a KTGT for those supplied credentials, which can be used to mount a users SMB share with that kerberos ticket.
A feature introduced with Nextcloud 14 for example.

So GSSAPI is basically a wrapper API between any application (not necessarily a browser, but can also be LDAP or SMB) and authentication methods. So the applications don't have to implement new or other authentication methods themselves but rely on said API...

regards,
kleinem

Stanislav1975 (commented):
[](url)

susnux:
This documentation does not exist anymore, closing as stale

PR #503, opened 2017-06-15
Header link tweak

jasonbayton:
I feel like the header link is too close to the heading which doesn't look particularly pleasing on the eye. So this is a very small tweak to add a bit of space and remove the underline (again because it doesn't look pretty).


MorrisJobke:
stable12 90d4e3e2

MorrisJobke:
stable11 7f28b6b9

PR #507, opened 2017-06-18
Add ownCloud migration paths to the docs

nickvergessen:
Since the question arises quite often, lets add it to the docs.

@MorrisJobke @jospoortvliet 

> migrating_from_owncloud_ _nextcloud_12_server_administration_manual_12_documentation_-_2017-06-18_16 25 28

### DIFF
--- a/admin_manual/maintenance/index.rst
+++ b/admin_manual/maintenance/index.rst
@@ -12,3 +12,4 @@ Maintenance
    manual_upgrade
    package_upgrade
    migrating
+   migrating_owncloud
--- a/admin_manual/maintenance/migrating_owncloud.rst
+++ b/admin_manual/maintenance/migrating_owncloud.rst
@@ -1,0 +1,33 @@
+=======================
+Migrating from ownCloud
+=======================
+
+
+.. note:: Especially when migrating from ownCloud to Nextcloud you should
+          create a backup of the config, database and the data directory,
+          in case something goes wrong.
+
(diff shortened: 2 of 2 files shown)


MorrisJobke:
What about the migration script?

nickvergessen:
Couldn't find anything useful with google where I could link to...

MariusBluem:
What is with migrating from ownCloud 8.2 and older?
However: Why do we not mirror the guide from here: https://nextcloud.com/migration?

nickvergessen:
Ah I forgot about that page

Not a fan of mirroring stuff... Maybe link from docs to there or the other way

MariusBluem:
I‘d go with moving it to the documentation, and removing it from the homepage. I will do the same for CODE, since I don’t think it is a good idea to have such manuals on pages where non-technical enterprise users look at ;)

jospoortvliet:
@MariusBluem we have to have some info on the migration on our website - as it is an important subject. We need to have some information on how, too - as in, a step by step like we have there now,because we have to show how easy it is. But we could simplify it and link to our documentation, yes.

jospoortvliet:
Besides all that, this looks good :1st_place_medal: 

I would point out that using our updater is also possible, though. I promote that as the best way of doing this from our website and blog, now, I think our documentation should probably say the same.

MorrisJobke:
stable12 3cc84478

PR #515, opened 2017-06-22
change value of innodb_large_prefix to 1

Alixerid:
Older versions of mysql may not accept `on` as a value and configuration may fail. `1` is sure to work as boolean value. 
See issue [#4846](https://github.com/nextcloud/server/issues/4846) of nextcloud server.


MorrisJobke:
stable12 f961c933
stable11 80082366

PR #520, opened 2017-07-04
Switch to OCSv2

schiessle:
All Nextcloud versions support OCSv2 so our documentation should recommend to use this API which result in a more restful behaviour.

cc @rullzer


MorrisJobke:
stable12 8dc968c3

Issue #532, opened 2017-07-23
Dead link to "client" repo in developer manual 


Vetm:
There is a dead link to the client repo in developer_manual/bugtracker/index.rst
I'd like to do a PR for this.

Vetm:
#535

PR #535, opened 2017-07-23
Fix dead link in bugtracker

Vetm:
Fixes: #532 

Fix dead link to client repo in developer_manual/bugtracker/index.rst


MorrisJobke:
ported to stable12, 11 and 10

MariusBluem:
Thanks for heads up and your contribution 🙈 I am not really happy with this, since this will imply that client issues should be reported in the theming repo - but unfortunately, I have no better solution ATM.

PR #560, opened 2017-09-05
Add error pages (404, 403) to nginx configuration

Kissaki:


Kissaki:
Even though random non-existant URLs (`/wat`) seem to redirect to `/apps/files`, the configuration returns 404 for other paths (for example `updater/wat/`).

The 403 and 404 templates still exist and code with comment `//also provide standalone error page`.

The `.htaccess` does not seem to so restrictive/redirect as often. 404 may be more common there. It defines these template file paths for 403 and 404 as error documents (like this change introduces for nginx).

MorrisJobke:
We want to get rid of this, because the standalone stuff broke too often. That's the reason we removed it from .htaccess and als from the Nginx documentation. Thanks for the hint anyway.

PR #562, opened 2017-09-17
Use IPv6 too in the Nginx config

C0rn3j:
Added IPv6 listens. Without them Nextcloud won't work over IPv6.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -41,6 +41,7 @@ webroot of your nginx installation. In this example it is
 
   server {
       listen 80;
+      listen [::]:80;
       server_name cloud.example.com;
       # enforce https
       return 301 https://$server_name$request_uri;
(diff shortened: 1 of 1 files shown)


MorrisJobke:
cc @josh4trunks

josh4trunks:
I do not personally use IPv6. Do you really need both `listen 80;` and `listen [::]:80;`?
Can you not just put `listen [::]:80;` if you want to listen on both IPv4 and IPv6?

C0rn3j:
There's a **Linux only**(might have changed since I last checked) flag you can set which would result having in having a single IPv6 line. 

I think the way I initially suggested is more clear and cross platform.

listen [::]:443 ssl http2 ipv6only=off;

josh4trunks:
Ok, I agree the way you suggest is best.
👍

MorrisJobke:
stable12 f62ee839

PR #566, opened 2017-10-02
Fix link to theming doc

pierreozoux:


pierreozoux:
I copied from there: https://github.com/nextcloud/documentation/blame/master/admin_manual/operations/scaling_multiple_machines.rst#L38

I never worked with rst, I thought this would help..

juliushaertl:
@pierreozoux It is a bit tricky, since Nextcloud is using three different sphinx projects that are not aware of each other. I've added a commit to include intersphinx which will allow us to reference parts from any documentations. It should work fine now.

MorrisJobke:
Ported to stable12:

d14f8fff
e40b144a

PR #576, opened 2017-10-06
Add version switcher, global configuration and fix some styling issues

juliushaertl:
- Show full title in previous/next links
- Add version switcher to switch between documentation versions
- Add global configuration so we can have values across the different docs

Before
bildschirmfoto vom 2017-10-06 13-05-44

After
bildschirmfoto vom 2017-10-06 13-54-31

cc @nextcloud/documentation


MariusBluem:
Can we just call this „versions“ instead of „Nextcloud versions“

juliushaertl:
@MariusBluem Sure, fixed it. I've also made some adjustments to the nextcloud_release theme: 

bildschirmfoto vom 2017-10-06 18-41-00

_shared_assets/themes/nextcloud_com/static/styles.css
```diff
@@ -1108,7 +1108,6 @@ a.headerlink {
 }
 
 #menu-support > ul > .current {
-    border: 1px solid #999 !important;
```
MorrisJobke: Removing this will remove the left border in the menu for the active entry:

Without:

bildschirmfoto 2017-10-23 um 22 46 57

With:

bildschirmfoto 2017-10-23 um 22 46 53

juliushaertl:
The left border was actually removed by intention, since for me it doesn't add much visual benefit and seems to be somehow inconsistent. But I'm fine to keep it for now. Maybe we can find a better solution for indicating the current section later.

MorrisJobke:
> The left border was actually removed by intention, since for me it doesn't add much visual benefit and seems to be somehow inconsistent. But I'm fine to keep it for now. Maybe we can find a better solution for indicating the current section later.

I'm all in, but for now this was the best that I was able to come up in a quick time back then. Maybe we should use the same design as in Nextcloud itself.

skjnldsv:
I rebased, Let's see if drone is ok.

Issue #509, opened 2017-06-21
Add PHPdoc class documentation to docs.nextcloud.com


juliushaertl:
It is still linking to the other project, i guess we should have an up to date version of the phpdoc class documentation for Nextcloud.

See: Nextcloud App API
https://docs.nextcloud.com/server/12/developer_manual/app/index.html#phpdoc-class-documentation

juliushaertl:
maybe @MorrisJobke ?

MorrisJobke:
Mmmmh ... any good PHPDoc documentation generator? Because I really don't like the default one, because it's not very easy to read :/

But then I'm open do add a cron job that updates this.

juliushaertl:
https://github.com/ApiGen/ApiGen looks promising, but I was not running out of the box with the Nextcloud code base when I gave it a short try.

I'll check if i find anything else.

juliushaertl:
@BernhardPosselt mentioned https://github.com/varspool/sphpdox which would allow us to include to API documentation to the sphinx built developer documentation. Sadly it is not maintained, as it seems in the repo. Documentation builds fine except for one error with the following line: https://github.com/nextcloud/server/blob/5f25dd7095fadccb7458dbe09c048fbf6645552d/lib/public/IPreview.php#L49

Other options i still need to check out:
- http://phpdox.de/
- https://github.com/FriendsOfPHP/Sami

Maybe we can also use PHPdoc (https://www.phpdoc.org) as before and look for a different template that makes it a bit nicer to read.

juliushaertl:
Implemented with #578

PR #578, opened 2017-10-18
Add PHP API documentation to developer manual

juliushaertl:
This is the first step for #509 to add a API reference to the developer manual.

For the purpose i've written a small tool called [phpdoc-to-rst ](https://github.com/juliushaertl/phpdoc-to-rst)that will generate sphinxcontrib-phpdomain based rst files out phpDoc annotated source code.

Here is a preview of the current state:
https://bitgrid.net/~jus/nextcloud-api/com/api.html

New features:
- References to classes/methods/... are possible from within our developer manual
- Any element has a link to the source code on github.
- Each interface also shows a list of classes that implement it. This might be quite useful when you want to check how something is actually implemented.

The current state is that both lib/public/ and lib/private/ are scanned, so we can have nice linking to classes that implement interfaces from the \OCP namespace.

I welcome any feedback, especially in terms of readability.

Todo:
- [x] Drone test for building (requires https://github.com/nextcloud/docker-ci/pull/58)
- [x] Fix some issues in the servers phpDoc comments https://github.com/nextcloud/server/pull/6864


nickvergessen:
Looks promising 😎 

* Deprecation should be more prominent (is there a way to correctly link to another method? example https://bitgrid.net/~jus/nextcloud-api/com/api/OCP/Activity/IManager.html#OCP\Activity\IManager::publishActivity )
* Multi type definitions are not linked: https://bitgrid.net/~jus/nextcloud-api/com/api/OCP/Activity/IEventMerger.html#OCP\Activity\IEventMerger::mergeEvents
* Does source always link to master? Or does it link to the respective stable* branch?

juliushaertl:
@nickvergessen Good points. The linking issue might need some fixes in sphinxcontrib-phpdomain. I'll have a look.

> Does source always link to master? Or does it link to the respective stable* branch?
Right now, only to master, but of course it should link to the according branch.

skjnldsv:
I like it a lot!!!

juliushaertl:
> Deprecation should be more prominent (is there a way to correctly link to another method? example https://bitgrid.net/~jus/nextcloud-api/com/api/OCP/Activity/IManager.html#OCP\Activity\IManager::publishActivity )

I've added styles so that deprecated methods are crossed out. The correct way according to the phpDoc reference is to add a @see tag. That one will be linked automatically. http://docs.phpdoc.org/references/phpdoc/tags/deprecated.html

> Multi type definitions are not linked: https://bitgrid.net/~jus/nextcloud-api/com/api/OCP/Activity/IEventMerger.html#OCP\Activity\IEventMerger::mergeEvents

Fixed as well.

> Does source always link to master? Or does it link to the respective stable* branch?

The branch can now be specified in build/generateApiDoc.php

The preview URL has been updated as well.

skjnldsv:
So, where we at here? :)

tflidd:
@nickvergessen can you have a look again?

.drone.yml
```diff
@@ -2,4 +2,5 @@ pipeline:
     documentation:
         image: nextcloudci/documentation:documentation-4
         commands:
+            - make api-docs
```
MorrisJobke: Best would be to include "api-docs" in "all", because we also use that on our docs system deployment. :wink: Let me have a look into this.

MorrisJobke: Or even better: Include it in "make dev-html" ;) Then it's just at the right spot.

MorrisJobke:
@juliushaertl Any idea regarding this:

```
PHP Notice:  Undefined index: opts in /drone/src/github.com/nextcloud/documentation/build/vendor/juliushaertl/phpdoc-to-rst/src/Builder/NamespaceIndexBuilder.php on line 171
```

MorrisJobke:
Should we keep that task as an on-demand step or better check the computed stuff in? (like we do for the config sample as well - see https://github.com/nextcloud/documentation/commits/master/admin_manual/configuration_server/config_sample_php_parameters.rst)

juliushaertl:
@MorrisJobke The commit looks good.

> Should we keep that task as an on-demand step or better check the computed stuff in? 

I see no real benefit in committing the computed files. It may lead to some confusion when people start to fix something right in the rst files, which get's overwritten later by the automated documentation generation.

PR #581, opened 2017-10-24
Add basic redis configuration example

C0rn3j:
This is enough to get Redis up and running.

Currently there is no information about actually setting up redis in a way that works on my setup, this fixes it.

Tested on Ubuntu 16.04 LTS with ondrej/php and chris-lea/redis-server PPAs.

### DIFF
--- a/admin_manual/configuration_server/caching_configuration.rst
+++ b/admin_manual/configuration_server/caching_configuration.rst
@@ -155,6 +155,17 @@ recommended if Redis is running on the same system as Nextcloud) use this exampl
 
 Only "host" and "port" variables are required, the other ones are optional.
 
+For the Redis side, you may need to add your web server user to the redis group::
+
+  usermod -a -G redis www-data
+
+If you were to use the Unix socket example right above, you'd edit and set the following lines in ``/etc/redis/redis.conf``::
+
+  port 0
+  unixsocket /var/run/redis/redis.sock
+  unixsocketperm 770
+  requirepass secret
+
 Redis is very configurable; consult `the Redis documentation 
 <http://redis.io/documentation>`_ to learn more.
 


MorrisJobke:
I'm not sure if we should include this. usually `apt install redis-server` is fully enough on Ubuntu.

C0rn3j:
Wasn't for me on 16.04 LTS server, it wouldn't work neither through ports or the unix sockets untill I added the www-data user to the redis group.

I used nginx for the webserver.

I tried with the stable Ubuntu 3.x redis release and the PPA 4.x redis.

tflidd:
You have to do these changes manually, see e.g. a popular howto:
https://bayton.org/docs/nextcloud/installing-nextcloud-on-ubuntu-16-04-lts-with-redis-apcu-ssl-apache/#6-2-enable-caching

admin_manual/configuration_server/caching_configuration.rst
```diff
@@ -155,6 +155,17 @@ recommended if Redis is running on the same system as Nextcloud) use this exampl
...
+For the Redis side, you may need to add your web server user to the redis group::
+
+  usermod -a -G redis www-data
+
+If you were to use the Unix socket example right above, you'd edit and set the following lines in ``/etc/redis/redis.conf``::
+
+  port 0
+  unixsocket /var/run/redis/redis.sock
+  unixsocketperm 770
+  requirepass secret
```
tflidd: Do we want to set up a password here? Especially like this, people will just copy & paste and use the password `secret`. And don't you have to put this password in the config.php?

If you only run Nextcloud and don't use redis for anything else (as socket, you obviously can't use it via network).

tflidd: I would just go with a simple setup without password and via socket like in (from @jasonbayton):
https://bayton.org/docs/nextcloud/installing-nextcloud-on-ubuntu-16-04-lts-with-redis-apcu-ssl-apache/#6-2-enable-caching

skjnldsv:

PR #605, opened 2017-11-14
add command to get user DN

enoch85:
I always forget, so better add it to the docs :)

cc @blizzz

### DIFF
--- a/admin_manual/configuration_user/user_auth_ldap.rst
+++ b/admin_manual/configuration_user/user_auth_ldap.rst
@@ -89,6 +89,8 @@ User DN:
   Example:
 
   * *uid=nextcloudsystemuser,cn=sysusers,dc=my-company,dc=com*
+  
+  You can get the user DN for a specific user with this Powershell command: ```dsquery user -name [user]```
 
 Password:
   The password for the user given above. Empty for anonymous access.


MorrisJobke:
cc @blizzz

blizzz:
It's an AD specific client tool. So far we don't have any hint to use any client tool to figure out this or that in this documentation page. If we start with it, we need to add examples for ldap-utils (ldapsearch command etc.), too (which are at least server agnostic), and would extend this page further. 

I would prefer if we either have a link to one or two cheat sheets that handle AD tools and ldap-utils or a seperate page were we would have some tips on these tools.

skjnldsv:

enoch85:
I would prefer to leave this open, if not @blizzz thinks it will never be implemented? With the reasoning above maybe this could be the first command, to add more later?

skjnldsv:
I'll let you two decide on what do to with that :)

blizzz:
No, i don't want to fill it up for commands of every possible client tool out there :) I am good with one of those options (in that order):

* link to cheat sheets instead
* only examples for a server agnostic tool
* have a different page for 3rd party clients.  
 
But that's really not something covering Nextcloud but the flavour of LDAP being used.

juliushaertl:
@enoch85 Mind to adjust this to one of the proposed ways?

enoch85:
I'm OK with closing this, I will find it here anyway. :)

PR #607, opened 2017-11-17
Fix malformed URL

peterdamian:
Change ".comg" to ".com".


MorrisJobke:
stable12 905e06aa
stable11 3925c920

Issue #602, opened 2017-11-11
Inconsistent header styles


peterdamian:
Your style guide documents three headers:

- the equals sign for h1 headers (with both overlining and underlining)
- the minus sign for h2 headers
- the caret for h3 headers

Within the documentation there are many exceptions to this convention. A number of h1 headers are missing the overlining, there are h1 headers used in place of h2 headers, and sometimes an h4 header (~~~)  is used.

Proper use of headers makes it easier to read the raw text and easier to manipulate the text programmatically.

See also #600.

MorrisJobke:
@peterdamian Could you open a PR to fix this? I guess this is a nice improvement 👍

PR #608, opened 2017-11-17
Fix headers in the admin manual only

peterdamian:
These updates affect the headers in the admin manual only. Since issue #602 involves a considerable number of changes, I've created this PR as a checkpoint. The majority of changes are straightforward but I did split one file to create 3 new files. I also removed a couple of headers that seemed redundant and made one or two minor changes to the text. To validate the changes, I generated the PDF of this manual locally and didn't see any problems with the output. (Is there a better way to validate reStructuredText?)

If these changes look fine, I'll submit a PR for changes to the remainder of the documentation.


MorrisJobke:
Looks good 👍

Issue #577, opened 2017-10-12
Backing up Nextcloud with mysqldump while running?


Yvan-Masson:
Hi,

[The documentation](https://docs.nextcloud.com/server/12/admin_manual/maintenance/backup.html) currently does not state if we can back up Nextcloud DB while it is running, while ensuring database consistency and ability to restore.

1. I suppose that Nextcloud should be stopped, at least to ensure that DB and data are in a consistent state regarding each other.

2. For `sqlite3` command, I could not find anything useful in manpage.

3. For MySQL/MariaDB, I read about `--single-transaction` option of `mysqldump` manpage and this makes me think that Nextcloud should be stopped:

`While a --single-transaction dump is in process, to ensure a valid dump file (correct table contents and binary log coordinates), no other connection should use the following statements: ALTER TABLE, CREATE TABLE, DROP TABLE, RENAME TABLE, TRUNCATE TABLE. A consistent read is not isolated from those statements, so use of them on a table to be dumped can cause the SELECT that is performed by mysqldump to retrieve the table contents to obtain incorrect contents or fail.`

4. I did not look PostgreSQL.

Could you write down if Nextcloud should be stopped or in maintenance mode during backup, or not?

Best regards,
Yvan

bakkegaard:
It seems to be a good idea to put your Nextcloud instance in maintenance mode 
https://help.nextcloud.com/t/how-to-backup-nextcloud/21623

Yvan-Masson:
Thanks for the answer. I am cautious so this is what I do, but I think this should be written in the documentation.

Yvan-Masson:
Thanks!
@MorrisJobke @tflidd

PR #623, opened 2017-12-02
Put server in maintenance mode before running backup

tflidd:
Fixes https://github.com/nextcloud/documentation/issues/577
@Yvan-Masson


tflidd:
- changed the order of things to backup to the order it is mentioned in the guide below (first all folders, then the database)
- add a modified version of the "maintenance mode"-section from the upgrade page.

MorrisJobke:
stable12 13615774

PR #625, opened 2017-12-04
clarify permissions needed for ldap password change

### LINKED ISSUE #624: LDAP-Password-Change documentation clarification
In  "[documentation/admin_manual/configuration_user/user_auth_ldap.rst](https://github.com/nextcloud/documentation/blob/stable12/admin_manual/configuration_user/user_auth_ldap.rst)" it is stated:
`Access control policies must be configured on the LDAP server to grant permissions for password changes.`

However this gives a false impression.
Default ldap accessrights allow (besides an admin) only the user himself to change his password, so this sencence can be interpreted in the wrong way.
Since a user has the accessright to change his own password in ldap, I assumed, that this is enough to fulfill the requirement. Also others made this assumption (See nextcloud/server#3200).

Please update the doumentation to reflect the fact that as [explained in this comment](https://github.com/nextcloud/server/issues/3200#issuecomment-280565627) the entity used for connecting to the ldap server needs the accessrights to change user passwords. 
As an alternative, implement a separate ldap connection with the user himself for the passwordchange, which would be securitywise much more sensible.

blizzz:
fixes #624

### DIFF
--- a/admin_manual/configuration_user/user_auth_ldap.rst
+++ b/admin_manual/configuration_user/user_auth_ldap.rst
@@ -375,18 +375,18 @@ Group Member association:
   have a very valid reason and know what you are doing.
 
   * Example: *uniquemember*
-  
+
 Enable LDAP password changes per user:
-  Allow LDAP users to change their password and allow Super Administrators and Group Administrators to change the password of their LDAP users. 
-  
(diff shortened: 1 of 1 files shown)


MorrisJobke:
stable12 3ca878d4

PR #626, opened 2017-12-05
Update system requirements - PHP 7.2 does not work

saper:
As per 12.0.1 changelog and https://github.com/nextcloud/server/pull/5793


saper:
Shall I also push it to stable10 or others?

MorrisJobke:
> Shall I also push it to stable10 or others?

No need - I will take care of the backports.

PR #638, opened 2017-12-21
Content list documentation

skjnldsv:
After https://github.com/nextcloud/server/pull/7195

Please review!
Is it clear enough? :)


jancborchardt:
Nice! :) Should we specifically say something about that we recommend using the `app-content-list-item-icon` for profile photos, or favicons, etc?

cc @nextcloud/designers

skjnldsv:
@jancborchardt yes, it's a requirement :)

juliushaertl:
Looks good to me. Just the screenshot besides the first code snippet should be moved above, because it currently shrinks the code block so that it is quite hard to read.

skjnldsv:
@juliushaertl I'll remove the inline image code block

jancborchardt:
Aaawesome! :) Great job with the docs as always @skjnldsv

skjnldsv:
Thank Jan! 🤗

PR #644, opened 2018-01-07
Change deprecated/removed database query

raimund-schluessler:
The call `getDb` was deprecated ages ago has been removed for NC12 already. I changed it to the correct call.

Also see https://github.com/nextcloud/server/pull/2665

Needs backport to stable12.


MorrisJobke:
stable12 fcb901a6

PR #645, opened 2018-01-08
Fix typo to match image

poldim:
Replace "/var/oc_data" with "/opt/nextcloud/"


juliushaertl:
@poldim Thanks for your pull request. Makes sense to have those be equal.

Issue #643, opened 2018-01-07
Document db:add-missing-indices


rullzer:
We should have info about this in the docs as well

PR #647, opened 2018-01-16
add documentation for the new occ command to add new database indices

schiessle:
fix https://github.com/nextcloud/documentation/issues/643

cc @rullzer @jospoortvliet

### DIFF
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -21,6 +21,7 @@ occ command Directory
 * :ref:`config_commands_label`
 * :ref:`dav_label`
 * :ref:`database_conversion_label`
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/occ_command.rst
```diff
@@ -487,6 +488,17 @@ This is example converts SQLite to MySQL/MariaDB::
...
+
+Add missing indices
+-------------------
+
+It might happen that we add from time to time new indices to already existing database tables,
+for example to improve performance. In order to check your database for missing indices run
+following command:
+
+ sudo -u www-data php occ db:add-missing-indices
+
```
rullzer: Should we add a note here that this can take a significant amount of time?
And that it is recommended to properly put the system in maintenance mode etc etc.?

PR #649, opened 2018-01-22
fix Memcached configuration

benIT:
There is a mistake in the Memcached setup instructions. 
This PR fixes the value to set up for Memcached in `memcache.local` .


tflidd:
Just above the modified section it is written that APCu is kept as local cache:

> This example uses APCu for the local cache, Memcached as the distributed memcache, and lists all the servers in the shared cache pool with their port numbers:

benIT:
sorry, I read too fast. PR closed.
greetings,
ben

PR #650, opened 2018-01-22
Add example for inserting csrf-token inside a form

te-online:
While reading up on CSRF and implementing it into my nextcloud app, I couldn't find how to add a CSRF-token to a form. So, I added this example. Not sure, if this is the best place, but I would look for it there.

### DIFF
--- a/developer_manual/general/security.rst
+++ b/developer_manual/general/security.rst
@@ -229,6 +229,15 @@ To prevent CSRF in an app, be sure to call the following method at the top of al
 
 If you are using the App Framework, every controller method is automatically checked for CSRF unless you explicitly exclude it by setting the @NoCSRFRequired annotation before the controller method, see :doc:`../app/controllers`
 
+When using forms in your app or theme, you need to supply a csrf-token to be checked when submitting the form. Insert a token like this:
+
+.. code-block:: php
+  
+  <form action="/index.php/apps/myapp" method="post">
+    <input type="hidden" name="requesttoken" value="<?php echo $_['requesttoken']; ?>" />
+  </form>
+  
+
 Unvalidated redirects
 ---------------------
 
--- a/developer_manual/general/security.rst
+++ b/developer_manual/general/security.rst
@@ -233,8 +233,13 @@ When using forms in your app or theme, you need to supply a csrf-token to be che
 
 .. code-block:: php
   
-  <form action="/index.php/apps/myapp" method="post">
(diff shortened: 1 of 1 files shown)


juliushaertl:
As @BernhardPosselt has stated on IRC $_['requesttoken'] will expire after some time. But since the oc_requesttoken global javascript value is not updated anyway at the moment I'd say we can add this to the docs for now, as it will help a lot for people who get started with app development. But maybe we should at least state, that the token might expire.

See https://github.com/nextcloud/server/issues/7155 for the progress of updating an expired token.

Any other thoughts @MorrisJobke @ChristophWurst

ChristophWurst:
> $_['requesttoken']

If we add info about that field it kind of sounds like this is a public and thus stable API. However, as far as I can tell, this field is just set because we use it in some internal templates. It could be removed at any point. I'd rather have a OCP interface (there might exist one already) to fetch a token and include that into your templates manually.

te-online:
Thanks for the heads-up. I added usage of the official interface (at least what I think is the official version...)
Could you have another look?

developer_manual/general/security.rst
```diff
@@ -229,6 +229,34 @@ To prevent CSRF in an app, be sure to call the following method at the top of al
...
+
+.. code-block:: php
+  
+  // in itemController.php
+  ...
+  function renderItemTemplate() {
+    ...
+    return new TemplateResponse('appname', 'itemTemplate', array(
+      'items' => ...,
+      'requesttoken' => (\OC::$server->getSession()) ? \OCP\Util::callRegister() : ''
```
ChristophWurst: What's the `->getSession()` check used for? Wouldn't it be easier and sufficient to just use a local variable in the template?

Something like
```php
<?php $CSRFToken = \OCP\Util::callRegister(); ?>
… value="<?php p($CSRFToken); ?>" …
```

te-online: Yeah, maybe you're right. I mean, it would be best practice to get it from the controller, right? But for a short documentation example your snippet is probably sufficient.
As for the `\OC::$server->getSession()`, I have no idea what it does, because I can't find it in the documentation 😆 I've incorporated that because I've seen apps do it like this. Maybe it's wrong? Can you say something about it?

skjnldsv:

ChristophWurst:
> since the oc_requesttoken global javascript value is not updated anyway at the moment

Not true anymore. We try to update it now.

So :+1: on closing this ticket. Apps should add the token via JS to ensure it's up to date

PR #651, opened 2018-01-23
Change starter to good first issue

MorrisJobke:
Ref https://github.com/nextcloud/nextcloud.com/pull/760 & https://github.com/nextcloud/nextcloud.com/pull/761

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,5 +1,5 @@
 pipeline:
     documentation:
-        image: nextcloudci/documentation:documentation-4
+        image: nextcloudci/documentation:documentation-5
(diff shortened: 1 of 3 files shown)


MorrisJobke:
I will just fix the problem with the CI - some dependencies have changed.

MorrisJobke:
@juliushaertl Please have a look at the failures to check if the dependencies needs to be adjusted. Also PHP7.1 instead of PHP 7.0 is required, which could be because master was used in one of the dependencies.

MorrisJobke:
stable12 8eb6b0d1

PR #659, opened 2018-01-29
OCS API also takes the Accept header into account

rullzer:


MorrisJobke:
stable12 35b1abda
stable11 4e3ecdd7

Dagefoerde:
I'd like to propose adding that `?format=` has priority over the `Accept` header (and that's  a really good thing if you can't manipulate the headers :) ).

PR #660, opened 2018-01-30
Master of docs is stable13 manual

MorrisJobke:
cc @juliushaertl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ developer-manual-pdf:
 	@echo "Developer manual build finished; PDF is updated"
 
 api-docs: clean-api-docs
-	cd build && sh get-server-sources.sh master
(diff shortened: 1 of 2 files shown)


juliushaertl:
Is there no stable13 branch for documentation? Because I'd say we should keep this on master for the master branch and just change for the individual release branches.

MorrisJobke:
> Is there no stable13 branch for documentation? Because I'd say we should keep this on master for the master branch and just change for the individual release branches.

No, because usually there happens a lot of documentation after the release and it thus avoids the mass backport of all the tiny things.

MorrisJobke:
I'm fine with reverting this once we split out stable13.

juliushaertl:
Ok, fine then :+1:

PR #671, opened 2018-02-07
add occ commands to upgrade documentation

schiessle:
We should document somehow that people should maually update the fileache table and the share index after upgrading to Nextcloud 13. Therefore I added this section to the upgrade documentation.

@MorrisJobke @rullzer @jospoortvliet please review

### DIFF
--- a/admin_manual/maintenance/upgrade.rst
+++ b/admin_manual/maintenance/upgrade.rst
@@ -91,3 +91,22 @@ Change ``"maintenance" => false`` to ``"maintenance" => true``:
     "maintenance" => true,
 
 Then change it back to ``false`` when you are finished.
+
+Manual steps during upgrade
+---------------------------
+
+Some opertation can be quite time consuming. Therefore we decided not to add them
+the the normal upgrade process. We recommend to run them manually after the upgrade
+was completed. Below you find a list of this commands.
+
+Upgrading to Nextcloud 13
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Nextcloud 13 we added a new index to the share table which should result in
+significant performance improvements:
+
+ $ sudo -u www-data php occ db:add-missing-indice
+
+With Nextcloud 13 we switched to bigint for the file ID's in the file cache table
+
+ $ sudo -u www-data php occ db:convert-filecache-bigint 
--- a/admin_manual/maintenance/upgrade.rst
+++ b/admin_manual/maintenance/upgrade.rst
@@ -39,7 +39,7 @@ Update notifications
 --------------------
 
 Nextcloud has an update notification app, that informs the administrator about
-the availablilty of an update. Then you decide which update method to use.
+the availability of an update. Then you decide which update method to use.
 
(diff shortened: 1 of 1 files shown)


MorrisJobke:
How could we communicate this to the users? Because nobody will look into the documentation after the update. Maybe we add a little notification in nextcloud that shows it when it is needed?

MorrisJobke:
cc @nickvergessen @rullzer

rullzer:
Can we inject a notification? Like. For better performance take a look at:

schiessle:
Yes, injecting a notification would be nice, I think we discussed this already, @rullzer. As a short solution the injected notification could be just "look at the documentation <link>". In the long run for Nextcloud 14 maybe we can even inject the concrete information like a notification for each occ command which should be executed.

But still, I think it is good to have this place in the documentation to collect all the occ commands for the different updates to look them up if needed.

admin_manual/maintenance/upgrade.rst
```diff
@@ -91,3 +91,22 @@ Change ``"maintenance" => false`` to ``"maintenance" => true``:
...
+
+Some operation can be quite time consuming. Therefore we decided not to add them
+the the normal upgrade process. We recommend to run them manually after the upgrade
+was completed. Below you find a list of this commands.
+
+Upgrading to Nextcloud 13
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Nextcloud 13 we added a new index to the share table which should result in
+significant performance improvements:
```
nickvergessen: double colon `::` please so that the command call is a code block

admin_manual/maintenance/upgrade.rst
```diff
@@ -91,3 +91,22 @@ Change ``"maintenance" => false`` to ``"maintenance" => true``:
...
+
+ $ sudo -u www-data php occ db:add-missing-indice
+
+With Nextcloud 13 we switched to bigint for the file ID's in the file cache table
```
nickvergessen: same here

admin_manual/maintenance/upgrade.rst
schiessle: Here we also don't use `::`` https://github.com/nextcloud/documentation/pull/671/files#diff-368ad62123233efc28400a735b869cf0L82
indention should be enough, preview looks fine too.

admin_manual/maintenance/upgrade.rst
schiessle: Here we also don't use `::` https://github.com/nextcloud/documentation/pull/671/files#diff-368ad62123233efc28400a735b869cf0L82
indention should be enough, preview looks fine too.

admin_manual/maintenance/upgrade.rst
nickvergessen: The colon is in line 80 (last none code line).
Fix is in https://github.com/nextcloud/documentation/pull/676

Not sure which preview you looked at, but see the screenshots in my PR

PR #675, opened 2018-02-09
Mention new maintenance:theme:update

ViktorF:
In Nextcloud 13 maintenance:theme:update got added. See nextcloud/server#5441

This mentions the new command.


MorrisJobke:
cc @nextcloud/theming

ViktorF:
Sure! I will add a note there too.

Where would be the best place for that? I'd say under "Inserting your new logo" but I'm not sure.

juliushaertl:
> Where would be the best place for that? I'd say under "Inserting your new logo" but I'm not sure.

Yes, I'd say that is the best place, so that people are aware that this might be needed to update cached image URLs.

PR #680, opened 2018-02-13
Remove Nextant & Add Talk

MariusBluem:
@jospoortvliet


MariusBluem:
Fulltextsearch is in BETA stadium and should not be listed as supported.

PR #682, opened 2018-02-14
Update install instructions for PHP7 on Ubuntu.

maxolasersquad:
Some instructions are PHP5 specific, while someone doing a new install is probably using PHP7. I updated some sections to reflect PHP7.

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -423,14 +423,15 @@ variables in the appropropriate ``php-fpm`` ini/config file.
 
 Here are some example root paths for these ini/config files:
 
-+--------------------+-----------------------+
-| Ubuntu/Mint        | CentOS/Red Hat/Fedora |
-+--------------------+-----------------------+
-| ``/etc/php5/fpm/`` | ``/etc/php-fpm.d/``   |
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> And I would use the reference to php7 first (for Nextcloud 14, we can remove the reference to php5 since it won't be supported any more)

The docs master is still 13 - so this is fine.

c0fe:
Is this still an open thing? I have installed PHP 7.0 and PHP 7.2 on my ends. I can add the instructions if this is still allowed.

MorrisJobke:
> Is this still an open thing? I have installed PHP 7.0 and PHP 7.2 on my ends. I can add the instructions if this is still allowed.

There shouldn't be anything else needed. We have the details for Ubuntu 16.04 already in the docs, which installs PHP 7.0. We should not document how to install third party repos, because they aren't supported by the distros anyways.

maxolasersquad:
I'm not clear on what is blocking this change from being accepted. Nextcloud 13 supports PHP 5 & 7. The "php-fpm configuration notes" only have instructions for configuring PHP5. This MR adds details for PHP 7.

I can add instructions for Debian if someone can provide those additional instructions. I don't have any machines running Debian.

PR #684, opened 2018-02-16
Documentation bump

skjnldsv:
WIP
- [x] New button
- [x] Inputs/form in popover
- [ ] ~~Top right header~~


developer_manual/design/index.rst
```diff
@@ -5,6 +5,7 @@ Design guidelines
 .. toctree::
    :maxdepth: 2
 
+   newbutton
```
juliushaertl: I'd say this should be part of the navigation section. :wink:

skjnldsv: I asked myself the same question! I think you're right :)

juliushaertl: Why is it placed outside? Then it wouldn't be in the sidebar anymore.

skjnldsv: Hu! I'm an idiot, I wanted to say outside the ul navigation, but yes, this is inside the app-navigation. ^^'

skjnldsv: Nevertheless, we also have a specific settings page which is also in the app-navigation div, so which do you prefer? Inside the navigation page or like this ?

juliushaertl: Ah ok :wink: Makes sense to have them grouped by the "app-navigation". Maybe we should also move the settings stuff to there, since that is in the app-navigation as well.

skjnldsv: I agree! :)

skjnldsv:
@juliushaertl Here we go! :)
Can you help and review https://github.com/nextcloud/server/pull/8373 pls?
So I can document this and add it here?

developer_manual/design/navigation.rst
```diff
@@ -2,10 +2,61 @@
...
 ===============
-App navigation
+New button
 ===============
 
+Introduction
+-------------
+
+A 'new' button is just a stylised button located above the navigation part of your app.
+It can be tied to whatever you want as long as it follow those simple rules.
```
skjnldsv: to whatever action?

developer_manual/design/navigation.rst
```diff
@@ -2,10 +2,61 @@
...
+===============
+
+The navigation section of any nextcloud app is the left sidebar.
+It is basically composed of
+
+* a 'new' button
+* a menu
+* a settings area
+
+The 'new' button and the settings area are optional.
```
skjnldsv: Should we add a global code block to show a full example?

skjnldsv:
@juliushaertl Here we go! :)
Can you help and review https://github.com/nextcloud/server/pull/8373 pls?
So I can document this and add it here? 

Also, see comments I made on this pr :)

cc @jancborchardt

developer_manual/design/navigation.rst
juliushaertl: Maybe just call it primary action button, since it is then not limited to a create action. And we can  get rid of the `It can be tied to whatever you want` sentence.

developer_manual/design/navigation.rst
```diff
@@ -2,10 +2,61 @@
...
+        <div class="app-navigation-new">
+            <button type="button" class="icon-add">Add user</button>
+        </div>
+    </div>
+
+Rules
+------
+
+* Stay simple, don't use overcomplicated text in this button.
+* Avoid using sentences longer than one line.
```
juliushaertl: Maybe also:
- Only use a single button

developer_manual/design/navigation.rst
```diff
@@ -387,3 +438,43 @@ Various information
...
+=========
+Settings
+=========
+
+Introduction
+-------------
+
+To create a settings area create a div with the id ``app-settings`` inside the ``app-navigation`` div.
+
+* The data attribute ``data-apps-slide-toggle`` slides up a target area using a jQuery selector and hides the area if the user clicks outside of it.
```
juliushaertl: Is the toggling done by server code or do the app developer need to implement it themselves?

developer_manual/design/navigation.rst
juliushaertl: Should be clear from the other examples imo.

developer_manual/design/navigation.rst
skjnldsv: server :)

skjnldsv:
Let's get this in! :)
@nextcloud/documentation

MorrisJobke:
> Let's get this in! :)

Let's wait until 13 is branched off.

skjnldsv:
@MorrisJobke clear? :)

MorrisJobke:
> @MorrisJobke clear? :)

Pushed `stable13` -  ready to go 👍

PR #705, opened 2018-03-05
New layout and design

### LINKED ISSUE #244: Adjust Header
cc @Espina2 @jospoortvliet

skjnldsv:
Fix #244 
Shameless copy/paste and adapted the correct classes to have a static website.
Of course, no fancy menu nor animations, this is the documentations, there's no need to over complicate things that much :)

- Css are copy/paste from the sass generated one from nextcloud_com git
- Cleaned most of our html in header and footer since there's a lot of functionalities we don't want
- Switched to automatic copyright generation to match our site


_shared_assets/themes/nextcloud_com/static/styles.css
```diff
@@ -7,33 +7,7 @@
...
-    src: local('Open Sans Italic'), local('OpenSans-Italic'), url(fonts/OpenSans-Italic.woff) format('woff'), url(fonts/OpenSans-Bold.ttf) format('ttf');
-}
-
-@font-face {
-    font-family: 'Open Sans';
-    font-style: normal;
-    font-weight: light;
-    src: local('Open Sans Light'), local('OpenSans-Light'), url(fonts/OpenSans-Light.woff) format('woff'), url(fonts/OpenSans-Light.ttf) format('ttf');
-}
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,600,600i,700,700i,800,800i');
```
juliushaertl: Let's just keep the local fonts please :wink:

skjnldsv: @juliushaertl I'm using the same data used by our website. Otherwise we have a font issue as the SemiBold is missing from the documentation ressources :)

juliushaertl: @skjnldsv I've pushed two commits to add the newest OpenSans fonts and revert the font face rules, so fonts will be fetched from the same host.

MorrisJobke:
Deployed already - we maybe should port it later to stable12 and stable11 once we have more feedback.

PR #711, opened 2018-03-07
updated reverse_proxy_configuration.rst

c0fe:
Replaced double quotes with single quotes. Double quotes instead of single quotes creates issues.

### DIFF
--- a/admin_manual/configuration_server/reverse_proxy_configuration.rst
+++ b/admin_manual/configuration_server/reverse_proxy_configuration.rst
@@ -65,11 +65,11 @@ you can set the following parameters inside the :file:`config/config.php`.
 
   <?php
   $CONFIG = array (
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> Replaced double quotes with single quotes. Double quotes instead of single quotes creates issues.

Not really. Those have the same meaning. Except that double quotes allow variable replacements. Something like `"$user"` would put the content of the variable with that name there while `'$user'` does only put that string there. But this should not cause issues. Maybe you used a different kind of double quotes and not the ascii version.

MorrisJobke:
But it should be fine to put single quotes there. There it's less likely that something is messed up.

c0fe:
@MorrisJobke i figure single quotes would be better as it sticks with the format used in nextcloud v13. i am planning on expanding this documentation further.

PR #713, opened 2018-03-09
Update min OSX requirement

rullzer:


MorrisJobke:
stable12 78e6207f 718fbc17

PR #719, opened 2018-03-15
Update index

jospoortvliet:
* updated metadata
* update piwik & mautic code
* update footer

Sorry, I should've noted I was still working on these...

PR #720, opened 2018-03-15
Unified indentdation + some other small things

splitt3r:
Changes requested in #719 unified indetation as per https://docs.nextcloud.com/server/13/developer_manual/general/codingguidelines.html. Also removed a wrong `<font>` tag and some spaces.


jospoortvliet:
I now feel like I need somebody cleaning up after me when I make changes... :see_no_evil:

PR #722, opened 2018-03-19
Fix latexpdf generation for developer_manual (also fixes automatic builds)

julijane:
PDF creation of developer_manual via 'latexpdf' target fails because 
'client_apis/WebDAV/search.rst' uses a unicode character 
(U+274C CROSS MARK) which pdflatex does not recognize. 

Added character definition to preamble in conf.py makes pdflatex replace
it with a plain "X" which is a good enough replacement.

### DIFF
--- a/developer_manual/conf.py
+++ b/developer_manual/conf.py
@@ -183,7 +183,7 @@ latex_elements = {
 #'pointsize': '10pt',
 
 # Additional stuff for the LaTeX preamble.
-'preamble': '\extrafloats{100}\maxdeadcycles=500',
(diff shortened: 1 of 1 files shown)


julijane:
By the way, this is also the reason why the continous-integration builds all failed in the other pull requests. You can see that here the build was successful.

MorrisJobke:
> By the way, this is also the reason why the continous-integration builds all failed in the other pull requests. You can see that here the build was successful.

You are awesome 🚀 Thanks a lot. Was not paying attention to it sadly ...

MorrisJobke:
Feel free to also join the developer channel on IRC on freenode: #nextcloud-dev

MorrisJobke:
stable11 0d1fef57
stable12 e12cd18e

PR #725, opened 2018-03-23
Fixed header on small width

### LINKED ISSUE #716: On small screens the new header is collapsed too far
bildschirmfoto 2018-03-15 um 07 50 07

ref #705 - cc @skjnldsv

skjnldsv:
Fix #716

Honestly, this is quite a mess, but I don't really want to spend too much time on this.
I shouldn't have just copied the css from the server, but it was much quicker to. Especially considering that the css on the main website is very shady (divs that have a 0 height/width with absolute positioning when their children have big contents, etc etc. In the end, it's just the documentation, it looks good and we should focus on the content :) )

### DIFF
--- a/_shared_assets/themes/nextcloud_com/static/styles.css
+++ b/_shared_assets/themes/nextcloud_com/static/styles.css
@@ -391,152 +391,13 @@ nav a {
 @media (max-width: 992px) {
   .nav .container .brand {
     padding-left: 15px; }
-    .nav .container .brand .logo {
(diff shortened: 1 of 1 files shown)


jospoortvliet:
whoah, this was copied from the website CSS which was generated from fancy and complicated SCSS written by Paulo - if this solves it, then great -I'm happy to keep this stuff simple and by golly, you did simplify :dagger: 

I tested it, works for me™ so you get my :+1:

PR #729, opened 2018-03-25
change link to nc 13 version

rm-rf-home:


MorrisJobke:
> @MorrisJobke why do the tests fail for such a minor change?

Because there was an issue on the CI system one week ago that was fixed already -> fine to go in

Issue #726, opened 2018-03-23
Design


splitt3r:
Sorry i messed things up with changing the asset urls in https://github.com/nextcloud/documentation/commit/c59fc4bcacb0a552f332881a2d233e7d1ce91612 seem to be the oC ones. @MorrisJobke could you replace the styles in the 13 folder with the ones from 9?

And the Mautic js code seems to be blocked by cors rule:
`Quellübergreifende (Cross-Origin) Anfrage blockiert: Die Gleiche-Quelle-Regel verbietet das Lesen der externen Ressource auf https://mautic.nextcloud.com/mtc/event. (Grund: CORS-Kopfzeile 'Access-Control-Allow-Origin' fehlt).`

PR #731, opened 2018-03-26
this should (...) fix the header and footer.

jospoortvliet:
It is a seriously horrible hack but I've included the css, js and html of our front page so we get the same fancy header. Of course, if we move a single asset there (which happens a LOT) we break our docu so this should be modified to be stand-alone, which I'll do in another PR. I first want to fix the ownCloud look our docu page... That is terrible.

I also cleaned up a lot of stuff in the styles.css.

If this gets deployed as the Nc 13 documentation, built for user etc etc, it should work. Though the three themes (...) are confusing, I modified the com one, as I think that that is the one we use.

It works locally perfectly but of course - only $DEITY knows what will happen upon deployment, this index.html is black magic :see_no_evil: 
Signed-off-by: Jos Poortvliet <jospoortvliet@gmail.com>

Fixes https://github.com/nextcloud/documentation/issues/726


MorrisJobke:
cc @skjnldsv

MorrisJobke:
Merge conflicts

jospoortvliet:
Maybe @skjnldsv can test it locally - and we should deploy carefully (not just before going to bed or something or risk :dagger: )

jospoortvliet:
Once this is in we can merge https://github.com/nextcloud/documentation/pull/732 too - which makes all assets static so we don't have to depend on our website (which is soon going to be a new site)

jospoortvliet:
@skjnldsv I know it isn't pretty but the current look sucks and if this works for you too, let's deploy it and clean it up later please? Our Documentation page looks horrible.

MorrisJobke:
backported to stable13: 59851ab8

PR #732, opened 2018-03-26
Make static

jospoortvliet:
When #731 is in, this can be merged to make the assets static. Tested locally - as far as is possible of course.


jospoortvliet:
Once this is in we can merge https://github.com/nextcloud/documentation/pull/732 too - which makes all assets static so we don't have to depend on our website (which is soon going to be a new site)

MorrisJobke:
Rebased to only contain the real changes.

MorrisJobke:
Ported to stable13

PR #737, opened 2018-03-30
CalDAV sync with macOS

MariusBluem:
The old docs were not longer working at all and this. The workaround for Snow Leopard and older was removed because it is probably not longer working at all and Snow Leopard and older are not longer supported ... This should work for the most cases :)

@MorrisJobke @georgehrke @raghunayyar

### DIFF
--- a/user_manual/pim/sync_osx.rst
+++ b/user_manual/pim/sync_osx.rst
@@ -1,51 +1,17 @@
-=======================
-Synchronizing with OS X
-=======================
+========================
+Synchronizing with macOS
+========================
 
-To use Nextcloud with iCal you will need to use the following URL::
+With Nextcloud 13 and its CalDAV-backend, syncing your Calendars with your macOS-device is easy.
 
-    https://example.com/remote.php/dav/principals/users/USERNAME/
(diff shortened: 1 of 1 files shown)


georgehrke:
Did you manage to successfully set up a calendar like this in High Sierra? I always have to use advanced for that and do it like in https://github.com/nextcloud/server/issues/7519#issuecomment-360936289

ticevi:
bildschirmfoto 2018-04-09 um 17 34 45
I'm on MacOS 10.13.4 and can't get a proper connection for my calendar even when using the provided work-around. <br>
Any ideas?

user_manual/pim/sync_osx.rst
```diff
@@ -1,51 +1,17 @@
-=======================
-Synchronizing with OS X
-=======================
+========================
+Synchronizing with macOS
+========================
 
-To use Nextcloud with iCal you will need to use the following URL::
+With Nextcloud 13 and its CalDAV-backend, syncing your Calendars with your macOS-device is easy.
```
MorrisJobke: d31db904

MorrisJobke:
Backported to stable13

PR #739, opened 2018-04-04
Example for custom apps directories does not work correctly

julijane:
This configuration example does not work because apps shipped with Nextcloud can
not be updated.

There is a current example. Nextcloud 13.0.1 ships with files_pdfviewer 1.2.0, but an update to 1.2.1 is available. However when using a configuration as shown in the example (and as similarly used in nextcloud/docker) updating does not work. The update is downloaded and installed into the second "writable" app directory, however Nextcloud will still load the old version as shipped and continue to complain about an update being available.

Changing the order so that the "writable" folder comes first fixed this. Updates are put there and Nextcloud uses them. This is arguably a bug in the server but I am not sure about that.

See also https://github.com/nextcloud/docker/pull/308

### DIFF
--- a/admin_manual/installation/apps_management_installation.rst
+++ b/admin_manual/installation/apps_management_installation.rst
@@ -71,14 +71,14 @@ in that folder.
 
     "apps_paths" => array (
         0 => array (
-                "path"     => OC::$SERVERROOT."/apps",
-                "url"      => "/apps",
-                "writable" => false,
-        ),
-        1 => array (
                 "path"     => OC::$SERVERROOT."/apps2",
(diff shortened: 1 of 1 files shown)


tflidd:
From the same doc:

> 
To ensure that the default /apps/ folder only contains apps shipped with Nextcloud, follow this example to setup an /apps2/ folder which will be used to store all other apps.

So I'm asking myself if this is not a wanted behavior that you can only update non-core apps. Why do you want a write protection on your apps folder in the first place?

skjnldsv:
@julijane this was an issue in our code, this has been fixed now :)

PR #746, opened 2018-04-17

shawnbarton:
Please rephrase my edit if necessary. 
Added a note that the nextcloud snap package version may differ from the current stable version. (currently it is stable nc12 and this is the nc13 documentation...)
It might also be wise to discuss with the user in advance that snap has certain drawbacks, such as being largely read only.

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -69,6 +69,9 @@ To install the Nextcloud Snap Package, run the following command in a terminal::
    that powers snaps, and it offers a new way to package, distribute, update and
    run OS components and applications on a Linux system. See more about snaps on
    `snapcraft.io <http://snapcraft.io/>`_.
+   
+.. note:: The snap package version of Nextcloud may deviate from the current stable
+   version.  
 
 .. _prerequisites_label:
 


MorrisJobke:
cc @kyrofa as the maintainer of the snap package.

kyrofa:
Technically any packaged version may lag release of upstream, is it worth calling that out specifically here? The snap will be updated as soon as we sort through the v13 issues. Regarding documenting the technical inner workings of the snap, I'm not sure this is the place to do that. For the VM we don't discuss the technical differences between vmware and kvm, for example. If a reader is interested in a specific technology, they can investigate it further (we give a link to docs, there).

shawnbarton:
Hi guys,
To clarify and summarise the points here:

-- The snap installation method is recommended above the manual installation. This makes the document look very clean but it could be interpreted that it is then the "preferred method" over manual installation, especially when one looks so easy compared to the other.

-- snapcraft.io is not up-front about the limitations of (the nextcloud) snap. This could be jarring for new users. It takes a little bit of study to realise that this isn't your average package. Docs -> "What is Snap?" says "It has a read-only file-system and, once installed, a writable area." - a newbie might not be able to interpret that this they can't actually change any files. I suppose the impact is very minimal since the snap can then simply be removed before proceeding with a manual installation. 

-- Running the command "sudo snap install nextcloud" downloads the latest stable snap which at the moment is 12.0.5snap2. The notification including the version number is only displayed after the package is downloaded and could be overlooked. Again, very minimal impact.

The decision is yours, just trying to help! Have a good one. This can be closed if there is nothing to be done.

PR #751, opened 2018-04-29
Use nextcloud.service/timer instead

francoism90:
The `cron` part should not be needed, there is already a timer.

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -90,9 +90,9 @@ systemd
 
 If systemd is installed on the system, a systemd timer could be an alternative to a cronjob.
 
-This approach requires two files: **nextcloudcron.service** and **nextcloudcron.timer**. Create these two files in ``/etc/systemd/system/``.
+This approach requires two files: **nextcloud.service** and **nextcloud.timer**. Create these two files in ``/etc/systemd/system/``.
 
-**nextcloudcron.service** should look like this::
+**nextcloud.service** should look like this::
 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/background_jobs_configuration.rst
```diff
@@ -90,9 +90,9 @@ systemd
 
 If systemd is installed on the system, a systemd timer could be an alternative to a cronjob.
 
-This approach requires two files: **nextcloudcron.service** and **nextcloudcron.timer**. Create these two files in ``/etc/systemd/system/``.
+This approach requires two files: **nextcloud.service** and **nextcloud.timer**. Create these two files in ``/etc/systemd/system/``.
 
-**nextcloudcron.service** should look like this::
+**nextcloud.service** should look like this::
```
MorrisJobke: I still think that the cron in the `.service` makes sense, because nextcloud itself runs in the apache/php-fpm service. Not that people are confused, because they started the `nextcloud.service` but then the instance is not running at all.

MorrisJobke:
cc @ChristophWurst @nickvergessen @rullzer Opinions?

admin_manual/configuration_server/background_jobs_configuration.rst
ChristophWurst: Makes sense, @MorrisJobke. I'm also in favor of keeping the `cron` part.

skjnldsv:
No feedback for a long time. Let's leave it like that :)
Thanks anyway for your pr @francoism90 :bowing_man:

PR #753, opened 2018-04-30
Fix missing anchor element

### LINKED ISSUE #752: Missing anchor icon when hovering a heading
Until recently, it was possible to hover a heading and a little icon appeared which led to an anchor URL like https://docs.nextcloud.com/server/13/developer_manual/general/devenv.html#enabling-debug-mode. Today, there is no hover effect, which makes it harder to quickly locate and copy those URLs to reference specific sections in our documentations.

I hope my report makes sense. I don't know the technical term for this hover effect thingy :wink:

MorrisJobke:
Fixes #752 

needs to be back ported to stable13

### DIFF
--- a/_shared_assets/themes/nextcloud_com/static/styles.css
+++ b/_shared_assets/themes/nextcloud_com/static/styles.css
@@ -892,6 +892,14 @@ div.thumbnail > img {
         height: 100%;
     }
 }
+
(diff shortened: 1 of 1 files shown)


MorrisJobke:
stable13 d1f09ee3

PR #754, opened 2018-04-30
user_auth_ldap: List password change feature

### LINKED ISSUE #734: LDAP page says LDAP is read only, but actually users can change their password
In "User authentication with LDAP" change "Only read access to your LDAP (edit or delete of users on your LDAP is not supported)" to make clear that users can actually change their LDAP password in Nextcloud.

awesome-manuel:
Fixes #734


MorrisJobke:
stable13 63027e6e

Issue #757, opened 2018-05-03
Broken link: https://docs.nextcloud.com/server/13/user_manual/files/desktop_mobile_sync.html


mathiasconradt:
Reported by a user via website form, details below. 

Comments: There is a broken link on this page: 

https://docs.nextcloud.com/server/13/user_manual/files/desktop_mobile_sync.html 

The link with a title of \"Nextcloud Desktop Client Manual\" points to https://docs.nextcloud.com/desktop/, but when I click on it I get \"Access forbidden\".

MorrisJobke:
stable13 ef206f62

PR #758, opened 2018-05-03
Fix link to client docs

juliushaertl:
Fixes https://github.com/nextcloud/documentation/issues/757


nickvergessen:
@MorrisJobke a sym link from https://docs.nextcloud.com/desktop/ to latest or something else then an error page would be nice, right? I think I have access to docs.nc.c so I could do it.

MorrisJobke:
Upps ...fixed it in 666f8007 🙈

MorrisJobke:
> @MorrisJobke a sym link from https://docs.nextcloud.com/desktop/ to latest or something else then an error page would be nice, right? I think I have access to docs.nc.c so I could do it.

Then also do it on `/server/` as well.

PR #761, opened 2018-05-10
Fix WebDAV URL for public shares

dragetd:
As discussed in nextcloud/server#9449


MorrisJobke:
stable12 dc1f2d48
stable13 efae7338

PR #766, opened 2018-05-16
Fix drone

juliushaertl:
This should™ fix the random failures of the api doc build caused by failing to clone the composer dependencies over git instead of fetching them via https.


MorrisJobke:
stable13 f929eac9 f21284a3 44e39937

PR #769, opened 2018-05-21
It would be safer and tider to make use of symlinks here.  (If the pa…

prynhart:
…ckages are updated then they won't make their way to /etc unless they are manually copied.)


MorrisJobke:
stable13 6c12acbe

PR #771, opened 2018-05-28
Fix nginx pattern which split path to script

r4sas:
Fixes error of parsing path of script, if user uses folders with php script-like names.

```
[error] *****#*****: *<pid> upstream sent invalid status "0" while reading response header from upstream, client: IP, server: cloud.example.com, request: "PUT /remote.php/webdav/vendor/test.php/tst.txt HTTP/2.0", upstream: "fastcgi://...", host: "cloud.example.com"
```

How reproduce error:
1. Auth on your nextcloud instance.
2. Create folder with `.php` at end of name.
3. Try to create file inside that folder.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -122,7 +122,7 @@ webroot of your nginx installation. In this example it is
       }
 
       location ~ ^/(?:index|remote|public|cron|core/ajax/update|status|ocs/v[12]|updater/.+|ocs-provider/.+)\.php(?:$|/) {
-          fastcgi_split_path_info ^(.+\.php)(/.*)$;
+          fastcgi_split_path_info ^(.+?\.php)(/.*)$;
           include fastcgi_params;
           fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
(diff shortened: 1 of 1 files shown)


tflidd:
A folder name with a point like an extension, couldn't this be a problem on some operating systems?

r4sas:
@tflidd idk. I have my instance on Debian 9, and my friend on Ubuntu Server 18.04. We both have same issue with manual-based pattern.

MorrisJobke:
cc @josh4trunks

josh4trunks:
Had to read up on the question mark in this case, but now this totally makes sense so me now. Question mark makes the match lazy, instead of greedy, which is what we want here. Tested this on my system and this fixes creating a text file in a folder named folder.php.

👍 Thanks!

josh4trunks:
I found this in my notes. **(?U)** has the same effect by making the entire (.+\.php) match lazy.
``fastcgi_split_path_info ^((?U).+\.php)(/.*)$;``

But I like your method since it acts directly on the match that matters.

MorrisJobke:
stable12 cf482533
stable13 40210cd3

PR #775, opened 2018-06-06
Fix inconsistencies in development environment setup instructions.

jernst:
As I'm attempting to go through the development environment setup instructions, some of it actually does not work as written and is incomplete, too. Here are some fixes.


MorrisJobke:
stable13 a439fb6f

PR #776, opened 2018-06-07
migrations from oc 9.1 can go to 11 directly

blizzz:
In my test I needed a `insert into oc_appconfig values ('core', 'vendor', 'owncloud');' first to get this going however, however this could not be confirmed by others.


MorrisJobke:
> In my test I needed a `insert into oc_appconfig values ('core', 'vendor', 'owncloud');' first to get this going however, however this could not be confirmed by others.

That is the reason - I upgraded 9.1.x to 10.0.x. and there it worked without problems. So it might be better to not need DB adjustments and is better IMO.

blizzz:
That's likely. In 11's version.php the path from 9.1 is whitelisted, therefore I went with that.

jospoortvliet:
This does make more sense, right? Given the relation in version numbers - 9.1 is close to 11?

MorrisJobke:
> This does make more sense, right? Given the relation in version numbers - 9.1 is close to 11?

Nextcloud external vs internal version numbers:

9.0 -> 9.0
10.0 -> 9.1
11.0 -> 11.0 


Because of historical reasons of easier migration. Thus a 9.1 (Nextcloud 10) can be directly migrated to 11 and doesn't need the step with go to 10 first in between. ;)

MorrisJobke:
> In my test I needed a `insert into oc_appconfig values ('core', 'vendor', 'owncloud');' first to get this going however, however this could not be confirmed by others.

But there is still this and it makes it easier to go to 10 first and then to 11. This is only a problem for .0 and .1, but let's make it not too complicated.

jospoortvliet:
So we don't merge this but close. yes?

MorrisJobke:
> So we don't merge this but close. yes?

Correct

PR #786, opened 2018-06-25
Use $request_uri instead of $uri in nginx config

benediktg:
As mentioned in nextcloud/server#5117, cc @josh4trunks 

(I am using it like this for quite a while now – so it basically works)


josh4trunks:
:+1: I am using this as well on my production configuration.

MorrisJobke:
stable13 1bb5ba25
stable12 56e8949b

rugk:
And the `$uri` in line 139 does not matter or what? (`try_files $uri/ =404;`)

PR #808, opened 2018-07-09
UTF8MB4 amendments

riegercloud:
added create statements and config.php value.


MorrisJobke:
stable13 097267c7
stable12 fb351a54

PR #819, opened 2018-07-24
Removed '=' from php_value statements

NlL5:
There were equal signs in the php_value statement in this section of the big_file_upload_configuration page.
According to the php.net manual, php_value statements must be without equal signs. Simply copying the statements provided in the documentation to a .htaccess or to an Apache config file will lead to a server error.


MorrisJobke:
Thanks for porting this to all branches. We would also have done this for you, but thanks for doing it :)

NlL5:
Oh ok. You're welcome ;)

PR #825, opened 2018-07-25
Update index.html - add links to current pdf manuals.

spiraldancing:
I couldn't find these links anywhere on the site, but the files are generated and available (I deduced the links from the patterns in the historical manual links).


spiraldancing:
BTW, I also couldn't find pdf versions of the desktop client, and android client manuals ... if pdf versions of those exist, I'd love to get them, too. Thanks.

MorrisJobke:
They are generated but contain a lot of failures in the build process. Thus we removed them from the docs. Also they were always a kind of second class citizen, because none of the images are built for the PDFs and the styling is pretty bad. I would love to see some work on this first, before making them available.

Also those PDFs are not updated anymore (last built in January), because they fail hard.

So :-1: for now. I will delete the files on the server for now as well.

PR #836, opened 2018-08-13
Remove XCache

danielkesselberg:
https://github.com/nextcloud/server/pull/10659


MorrisJobke:
This is imported from the config.sample.php ... so we should fix it in the server repo.

MorrisJobke:
See https://github.com/MorrisJobke/nextcloud-config-converter

PR #843, opened 2018-08-26
CentOS 7 walkthrough install created

hanserasmus:
Please check the formatting, especially the linking to documents. They do not display correctly in my .rst preview, but are done according to examples in the docs.


michag86:
I don't know if webtatic repo is a good source:
https://wiki.centos.org/de/AdditionalResources/Repositories/
It is listed under **Known Problem Repositories** with this comment _Provides updated packages with later versions of PHP, MySQL, and other packages. IUS or SCL are better alternatives._

I decided to use software collections, which is a centos provided repository.
There are Installation guides for php5.6 and php7.0:
https://github.com/nextcloud/documentation/pull/388

Meanwhile software collections also provide php7.1

I don't know if it would be better to extend the manual for updating to php 7.1 from scl?

hanserasmus:
Noted. But I have to make the comment about the package being in a production system currently for more than a year without causing any problems. It is also listed as the preferred repo on another site in their Install walk-through. So imho the CentOS wiki page might be overly conservative? @MorrisJobke can you please advise? Seeing as this will land up on Nextcloud's page? If needed I will have to rewrite it all using SCL, but then it will be PHP 7.1 and not 7.2?

@michag86  Did you maybe still test this walkthrough for any issues regarding the install, irrespective of the Webtatic repo not being the one of choice?

hanserasmus:
I have run into another issue whilst trying to use SCL, and that is that commands like `php -v` does not persist after reboots. You have to therefore run `scl enable rh-php71 bash` everytime. I know symlinking the apache files etc. should suffice for Nextcloud, but if a user wants to troubleshoot their php modules etc, they would have to enable this every time before doing that? I found evidence of this [here](https://access.redhat.com/solutions/527703) where they state it is a problem and there are workarounds, but thinking from a sysadmin's perspective, this seems cumbersome and unnecessary. Some input here would be great?
Please note this is a discussion, and not a confrontation by any means. I just think the idea behind a walk through is to get a user up and running as quickly as possible with as little issues as possible. To make the product welcoming basically.

michag86:
In our environment we have only installed php from scl. I've just created a symlink /usr/bin/php to the installed binary in the /opt/-path. I don't know if there is a problem with that. 

For the moment I had no time to test. Maybe on the weekend.

jospoortvliet:
@michag86 had any time to test? The longer it takes for changes to get in the more conflicts one gets and the more work it is to get in :cry:

hanserasmus:
@michag86 Thanks for all the input. I have added all you have suggested. Thanks for the input.

hanserasmus:
I'll look into the conflict tonight when I get home. Keep you updated.

hanserasmus:
@jospoortvliet I cannot resolve any conflicts, only people with write access can, like yourself. However, not sure why it is conflicting, because I have not even touched those lines. From a technical POV it seems fine, and can be submitted imho.

jospoortvliet:
This is the conflict - which part do I keep, top or bottom?

```
<<<<<<< master

Then create a symlink to :file:`/etc/apache2/sites-enabled`::

  ln -s /etc/apache2/sites-available/nextcloud.conf /etc/apache2/sites-enabled/nextcloud.conf

=======
 
Then enable the newly created site::

  a2ensite nextcloud.conf
 
>>>>>>> master
```

hanserasmus:
IMHO the bottom one  is the newest method in which apache enables
virtualhost sites.

jospoortvliet:
@MorrisJobke super, thanks for the backports!

@hanserasmus great work dude. @michag86 thanks for reviewing!!!

:tada: :balloon: :cake:

PR #846, opened 2018-08-30
add code signing instructions

hirschrobert:
I had trouble understanding how to sign the code. @MorrisJobke told me on Monday. Maybe helpfull for others, too. Haven't tested it, though.

### DIFF
--- a/admin_manual/issues/code_signing.rst
+++ b/admin_manual/issues/code_signing.rst
@@ -48,6 +48,11 @@ Is code signing mandatory for apps?
 
 Code signing is required for all applications on apps.nextcloud.com.
 
+How do I sign my code?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


MorrisJobke:
Ah ... Okay ... I only read the text and this is correct, but the location is wrong. There is sign-off and signing code. Let me remove it again. What you are looking for is the sign-off documentation and this is only needed in the nextcloud owner applications and is described in https://github.com/nextcloud/server/blob/master/CONTRIBUTING.md#sign-your-work

MorrisJobke:
> I only read the text and this is correct, but the location is wrong. There is sign-off and signing code

* sign-off: this means you add a statement, that you checked the license of the code or you wrote the code yourself and the stated license is what you apply to it
* code signing: cryptographically sign code to proof that the key owner checked that the code is what was there originally and a 3rd party can't manipulate it on the way to the Nextcloud instance (via appstore, GitHub releases, etc) - this is a way to verify during app installation that the code is still the one that should be used
* commit signing: cryptographically sign the git commits to be able to verify that the commit author was actually the person it claimed to be

PR #852, opened 2018-09-09
Set Referrer-Policy header for nginx installation docs

LouisMT:
This PR adds a `Referrer-Policy` header to the example nginx configuration. Because a warning has been added since Nextcloud 14, I think this header should be included in the example config using the most strict value by default. In special use cases the user can adjust this value and use a less strict value if needed.

See nextcloud/docker#451 and nextcloud/server#9122.

PS do I have to create another PR to target the `stable14` branch?


tilosp:
@LouisMT the headers are set at two locations within the config file. You missed the second location.

skjnldsv:
@tilosp is it? I'm only using it at one place on my config :)
Can you open a pr?

tilosp:
See: #855

tilosp:
Backport in #856

PR #853, opened 2018-09-10
Update Webcron url path

thomasylee:
It looks like the Webcron url path in Nextcloud 14 is at **/cron.php** rather than at **/nextcloud/cron.php**. I discovered this after spending a while trying to figure out why EasyCron wasn't invoking cron.php correctly. It turned out that it just needed to hit /cron.php instead.

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -55,7 +55,7 @@ service (for example, easyCron_), you ensure that background jobs are executed
 regularly. To use this type of service with your server, you must be able to
 access your server using the Internet. For example::
 
-  URL to call: http[s]://<domain-of-your-server>/nextcloud/cron.php
+  URL to call: http[s]://<domain-of-your-server>/cron.php
 
 Cron
 ^^^^


skjnldsv:
Hello and thanks for your pull request!

The whole file assume the installation is done on `/var/www/nextcloud/` so it's pretty logical to keep all things on the same page :)
What are we advertising on the other docs pages? install on `/var/www/` or `/var/www/nextcloud/`

thomasylee:
Ah, I see. I was using the [nextcloud:14 Docker image](https://github.com/nextcloud/docker/blob/f6abe4472b50d5bccff4428329fced339835cf69/14.0/apache/Dockerfile#L96), which uses `/var/www/html` instead of `/var/www/nextcloud` by default.

It looks like most of the pages also use `/var/www/nextcloud`, with a few exceptions like [the General Troubleshooting page](https://docs.nextcloud.com/server/13/admin_manual/issues/general_troubleshooting.html), which seems to go back and forth between the two.

I agree that this page is consistent with itself and the majority of the other pages though, so I'll go ahead and close out this issue. Thanks for the constructive pointer! :)

skjnldsv:
Thank you for the help!!

PR #857, opened 2018-09-12
Update source_installation.rst, OpenSSL PHP extension is required at least since 14

signupforacomment:
Upon installing Nextcloud 14 and accessing the login page, the message appears the OpenSSL PHP extension IS required to even login, so it's not supplemental.
Once done, the login page isn't working anyways so that's just a recommendation updating your documentation.


MorrisJobke:
stable14 b65324b9
stable13 efb73a5e
stable12 3dd29f88

PR #868, opened 2018-09-19
noted required setting for accessing public shares over WebDAV

lpirl:
Recently, we found that the setting "Allow users on this server to send shares to other servers" is required in order to access public shares over WebDAV.
If this setting is not enabled, "HTTP Access Denied" is returned (which is correct, but) what makes it difficult to find out how to enable this feature.
Hence, we should note this in the docs.


MorrisJobke:
stable13 9e9e4103
stable14 b980cbe2

lpirl:
Thanks @MorrisJobke

PR #871, opened 2018-09-20
[14] Adds the note property for the update share endpoint 

weeman1337:
Backport of #870

Closes https://github.com/nextcloud/server/issues/11270


weeman1337:
Harrk - I choosed the wrong base... GitHub doesn't update the changes after changing the base :(
I'll make a new one.

PR #873, opened 2018-09-22
Update contacts.rst

GeraldSoellinger:
I think most of the time we want to add a virtual contact file not the gene sequence of our friends:)
https://en.wikipedia.org/wiki/VCard
https://en.wikipedia.org/wiki/Variant_Call_Format


MorrisJobke:
stable13 1b3fa549
stable14 643011dd

PR #878, opened 2018-09-27
NC 14.0.1 requires the terminal / on overwrite.cli.url

MarcelWaldvogel:
for `overwrite.cli.url` when going to the root. Maybe required as well for other paths?

BTW: Could the following error message be made more specific?

```
# sudo -u www-data  php occ maintenance:update:htaccess
Error updating .htaccess file, not enough permissions or "overwrite.cli.url" set to an invalid URL?
```


MorrisJobke:
stable14 0d76a758

Issue #884, opened 2018-09-30
Debian derivatives installation example uses `ln -s` instead of `a2ensite`


uusijani:
Under [Apache Web server configuration](https://github.com/nextcloud/documentation/blob/02b666abc3397fd65c395fbeca800097e92daac4/admin_manual/installation/source_installation.rst#apache-web-server-configuration), there's a prompt to create a symlink to enable the newly created site:

    ln -s /etc/apache2/sites-available/nextcloud.conf /etc/apache2/sites-enabled/nextcloud.conf

As the instructions in this section are written for Debian, Ubuntu, and their derivatives, [`a2ensite`](https://manpages.debian.org/jessie/apache2/a2ensite.8.en.html) should be used instead:

    a2ensite nextcloud.conf

The commands are currently functionally equivalent (AFAIK), but should this change sometime, the latter would be more future proof, besides being much shorter and easier to type.

Additionally, immediately following the `ln -s` there is a [section about Additional Apache configurations](https://github.com/nextcloud/documentation/blob/02b666abc3397fd65c395fbeca800097e92daac4/admin_manual/installation/source_installation.rst#additional-apache-configurations), which correctly instructs the use of [`a2enmod`](https://manpages.debian.org/jessie/apache2/a2enmod.8.en.html) to create the symlinks. Replacing `ln -s` with `a2ensite` would thus harmonize the two sections.

MorrisJobke:
Hey @uusijani could I ask you to send this change in as a pull request? That would be nice 👍

PR #885, opened 2018-10-02
Use a2ensite instead of ln -s to enable site

uusijani:
Replace instruction using `ln -s` to enable Apache site to use `a2ensite` instead in Debian and derivatives. Closes [#884](https://github.com/nextcloud/documentation/issues/884).


MorrisJobke:
stable14 86362e62
stable13 94719b92

PR #889, opened 2018-10-08
further hints to improve system performance

Florz:
As seen in #2837 poor system settings and a lack of RAM can lead to file locking issues and sync conflicts.


Florz:
whoops, that's too much.

Florz:
@MorrisJobke is the request still OK or should I create a new one against stable14 with just my changes?

MorrisJobke:
> @MorrisJobke is the request still OK or should I create a new one against stable14 with just my changes?

I was about to test if this was maybe a mess-up with the branches but it looked the same against master & stable14. So maybe hand in a new PR.

PR #891, opened 2018-10-08
further hints to improve system performance

Florz:
As seen in #2837 poor system settings and a lack of RAM can lead to file locking issues and sync conflicts.

### DIFF
--- a/admin_manual/configuration_server/server_tuning.rst
+++ b/admin_manual/configuration_server/server_tuning.rst
@@ -11,13 +11,14 @@ benefits.
 Reducing system load
 --------------------
 
-High system load will slow down Nextcloud and might also lead to other unwanted side 
-effects. To reduce load you should first identify the source. Tools such as htop, 
-iotop or [glances](https://nicolargo.github.io/glances/) will help to identify the 
-process or the drive that slows down your system. First you should make sure that 
-you installed/assigned enough RAM. Swap usage should be prevented by all means.
-If you run your database inside a VM, you should not store it inside a VM image file.
-Better put it on a dedicated block device to reduce abstraction layers.
+High system load will slow down Nextcloud and might also lead to other unwanted 
+side effects. To reduce load you should first identify the source of the problem. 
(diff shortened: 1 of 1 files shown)


Florz:
@MorrisJobke I think this request looks better

admin_manual/configuration_server/server_tuning.rst
```diff
@@ -73,6 +99,27 @@ Enable HTTP2 for faster loading
...
+MySQL cache following values in your ``www.conf`` file should work:
+
+.. code:: ini
+
+  pm = dynamic
+  pm.max_children = 120
+  pm.start_servers = 12
+  pm.min_spare_servers = 6
+  pm.max_spare_servers = 18
+  pm.process_idle_timeout = 60s
```
danielkesselberg: > Used only when pm is set to ondemand.

(from http://php.net/manual/en/install.fpm.configuration.php)

admin_manual/configuration_server/server_tuning.rst
```diff
@@ -29,6 +41,20 @@ configure Nextcloud for MySQL or MariaDB. If your installation is already runnin
...
+smaller than 1GB. MySQL will consume up to 1GB of RAM for caching. Please make 
+sure that your system has sufficient free RAM after the change, so that it does 
+not start to use your swap partition, when it receives a burst of requests. In 
+the given example your ``/etc/mysql/conf.d/mysql.cnf`` might contain the 
+following lines. (beware that this is the block for mysqld not mysql)
+
+.. code:: ini
+
+  [mysqld]
+  innodb_buffer_pool_size=1073741824
```
danielkesselberg: You could set this to 1G (https://dev.mysql.com/doc/refman/5.7/en/innodb-buffer-pool-resize.html).

admin_manual/configuration_server/server_tuning.rst
Florz: ah, right pm.process_idle_timeout is unnecessary

PR #914, opened 2018-10-27
include VM as an option

enoch85:
cc @jospoortvliet


MorrisJobke:
I guess this is not the right place to link this, because this is about the specific requirements (Hypervisors) of the VM image itself. In general Nextcloud as such is hypervisor agnostic.

MariusBluem:
I agree 👍

MorrisJobke:
Done via #921

PR #921, opened 2018-11-05
We have two VM's, add Univention.

jospoortvliet:
Of course, we should have had both options here from the beginning... :see_no_evil:


MorrisJobke:
stable13 a03437da
stable14 b19712c7

enoch85:
Fine with me, but would be nice if you added SME as on the downloads page. :)

jospoortvliet:
I called it SMB instead of SME, sorry - otherwise, it's there, right?

enoch85:
Yes otherwise it's correct. I will make some more changes and I can change it to SME in the same PR.

jospoortvliet:
ok!

PR #922, opened 2018-11-05
Remove explanation about "sizelimit exceeded"

blizzz:
even stated in the php docs: http://php.net/manual/en/function.ldap-search.php

> Some directory server hosts will be configured to return no more than a preset number of entries. If this occurs, the server will indicate that it has only returned a partial results set. This also occurs if you use this parameter to limit the count of fetched entries.


blizzz:
actually… if we would wrap the error handler in ldap, we could throw this away.

MorrisJobke:
> **EDIT:** [nextcloud/server#12292](https://github.com/nextcloud/server/pull/12292) – if this is merged, we should scrap this section instead :)

Done and I updated this PR to remove the section.

PR #929, opened 2018-11-16
Path changed for KDE-Webdav

xstable:
Path seems to change.

### DIFF
--- a/user_manual/files/access_webdav.rst
+++ b/user_manual/files/access_webdav.rst
@@ -75,7 +75,7 @@ Accessing files with KDE and Dolphin file manager
 To access your Nextcloud files using the Dolphin file manager in KDE, use
 the ``webdav://`` protocol::
 
-    webdav://example.com/nextcloud/remote.php/dav/files/USERNAME/
+    webdav://example.com/remote.php/webdav
 
 .. image:: ../images/webdav_dolphin.png
    :alt: screenshot of configuring Dolphin file manager to use WebDAV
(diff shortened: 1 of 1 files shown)


user_manual/files/access_webdav.rst
```diff
@@ -75,7 +75,7 @@ Accessing files with KDE and Dolphin file manager
 To access your Nextcloud files using the Dolphin file manager in KDE, use
 the ``webdav://`` protocol::
 
-    webdav://example.com/nextcloud/remote.php/dav/files/USERNAME/
+    webdav://example.com/remote.php/webdav
```
MorrisJobke: Both should work. What problems do you noticed?

xstable: Problem is, that the "nextcloud/remote.php/dav/files...." version don't work at all.
You even don't get a Passwordrequest at all.... I think the error was that it can't connect to the server (or something simlar). I've tried a lot, and only as I found the "Use this address for Webdav" in the Web-GUI of nextcloud, it worked If I use this (which I've added above).

Maybe try and see yourself.

MorrisJobke: We use the new address since 2 years in our clients (Android and iOS) and they seem to work very well, because we didn't received negative feedback there as hard as in "it doesn't work". 😞

jospoortvliet: I'm afraid I can confirm, using ```webdav://example.com/nextcloud/remote.php/dav/files/USERNAME/``` is a no-workie:
```
The file or folder webdavs://cloud.nextcloud.com/nextcloud/remote.php/dav/files/(myusername) does not exist.
```

Using webdav://cloud.nextcloud.com/remote.php/webdav simply asks for username and password.

MorrisJobke: > ```
> The file or folder webdavs://cloud.nextcloud.com/nextcloud/remote.php/dav/files/(myusername) does not exist.
> ```

There is still the difference between username and loginname. If this is an LDAP instance for example those two are different. Anyways - @rullzer your opinion here?

jospoortvliet: Turns out - it was the extra 'nextcloud' in the URL, it works in Dolphin if you remove it.

skjnldsv: Works here as well.
@xstable any news on this?

skjnldsv:

PR #932, opened 2018-11-19
Revamp 2FA admin docs

ChristophWurst:
https://github.com/orgs/nextcloud/projects/17

# Admin
- [ ] Admin interface https://github.com/orgs/nextcloud/projects/17#card-13042458
- [ ] occ interface https://github.com/orgs/nextcloud/projects/17#card-13042440
- [ ] Ref/link to admin provider docs https://github.com/orgs/nextcloud/projects/17#card-13042448

# User
- [ ] user_manual/user_2fa.rst

# Developer
- [ ] Update developer_manual/app/two-factor-provider.rst

PR #936, opened 2018-11-22
Revert "Revamp 2FA admin docs"

MorrisJobke:
Reverts nextcloud/documentation#932

### DIFF
--- a/admin_manual/configuration_user/two_factor-auth.rst
+++ b/admin_manual/configuration_user/two_factor-auth.rst
@@ -1,23 +1,17 @@
 =========================
-Two-factor authentication
+Two factor authentication
 =========================
 
-Two-factor authentication adds an additional layer of security to user accounts. In order to log
-in on an account with two-factor authentication (2FA) enabled, it is necessary to provide both the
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> Hi Christoph, I wanted to know the reason you ask for a revert on my commit ? Did my commit wasn't clear enough or maybe not useful ? I know many people that were stuck on the installation of the wizard for that reason and that's why I asked for an enhancement.
> If needed I'm also ok to adapt the advices of this commit.
> I think that would be great to show a error message when needed in the installation wizard but I don't know enough of the nextcloud core programing for do it myself for the moment.

Sorry for this. It seemed to be a GitHub bug. I just clicked "revert" in the PR #932 and didn't checked again. Somehow it reverted two commits instead only the one. Let me bring it back to master. Thanks for noticing.

MorrisJobke:
Weird thing: the commit is actually also in master:

bildschirmfoto 2018-11-28 um 15 02 19

So it seems to be a glitch, because https://github.com/nextcloud/documentation/pull/936/commits/e8d6aed8e9c00de5870e7f2f04287fd41ecdebf4 is also empty.

PR #944, opened 2018-11-26
Add Apache modules to OAuth 2 docs

Dagefoerde:
IIRC, `mod_headers` should be the most important one... not sure if the others are required but they don't hurt either


rullzer:
@MorrisJobke you know more about apache than me

PR #946, opened 2018-11-27
Remove legacy theming documentation

MariusBluem:
As discussed! @MorrisJobke @skjnldsv 


MariusBluem:
> Is this broken or was it moved somewhere else?

It is simply deprecated and may cause some unexpected behavior. We are not putting any efforts fixing this feature and may remove this with a future release (of course this would be noticed in the future versions changelog). 

With the new theming-app and its advanced settings for favicons and our [custom css capabilities](https://apps.nextcloud.com/apps/theming_customcss) we cover 99% of this old and hacky way of theming. Advanced customizations which need more than our custom css or theming-app is providing already should be done with an app. Examples for that are the [LSD](https://github.com/jancborchardt/lsd) and [Unsplash](https://apps.nextcloud.com/apps/unsplash) apps.

MorrisJobke:
> Is this broken or was it moved somewhere else?

The goal is to let the theming app handle as much as possible. Themes just caused so many problems in the past that it is hard to keep them compatible with updates to the frontend.

In general we should try to replace this a bit more gradually. For example the section about favicon and logos can be replaced by a link to the theming app. The theming app also could be used together with a theme. In this way people can migrate step by step to the theming app.

J0WI:
How about redirecting to [admin_manual/configuration_server/theming.rst](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/theming.rst)?

deusoz:
Hi, bows and respect, Just wading in here.  As for the legacy folder theming, this is, from a developer point of view (mine anyway), having the folder method IS the creme de la theming methods.  Being able to just switch a folder with images and css is the best scenario one could ask for in this respect.   I see it being called "hackish", but I'm thinking that having to create a "theme hacking app" to theme Nextcloud, with very few choices of image manipulation used by other apps, i.e. file icons, is much more "hackish" than having the choice of integrated theme folders where one can store and point to images easily and change the look of a nextcloud fork with the change of a config var.  I was tickled pink when I read the code comments and figured out I could swap out the file icons and thought that whoever thought of all of this was truly "brilliant" (a utility app searching theme for used mimetype icons so that it could make a new config to instruct app to look only for those type icons in theme folder whenever it constructs an image URL using other core OC classes of the app).  I can actually make a theme folder a submodule and completely keep the core Nextcloud software untouched and still manipulate the icon graphics throughout the app.  I hope you "un" deprecate the legacy code.  If you need any assistance as much as I am capable, I volunteer to assist.  As for the argument that the folders theming has broken upgrades in the past, well, just warn people who choose to using theming folders and leave it at that. I really don't care if some folders changed version to version and that my fork had to adjust.  Just my thoughts, I don't pretend to know the whole picture.

deusoz:
OK this one just for laughs ok, can't contain myself...

You want to throw away this beautiful legacy folder theming method, and then tell me I should use a theme-hacking-app???  Are you guys tripping???

MariusBluem:
> You want to throw away this beautiful legacy folder theming method, and then tell me I should use a theme-hacking-app??? Are you guys tripping???

I just want to make absolutely clear that removing and deprecating this method does not mean we are removing it for now. We want to explicitly communicate that our future efforts go into a theming-app which is way more stable and matches 99% of the use-cases. At some point we all need to focus and not implement everything twice. The classic way of theming has it issues - this is why this decision was made regarding the documentation. There are NO changes in the code yet. Please have a look at my comment above linking to the custom CSS app and the LSD and Unsplash app. You can still modify everything by building custom apps - even if we would remove theming from our code (which is again not on the roadmap currently 😉).

jancborchardt:
I see no issue here? Functionality is still there but as it's not something we work on or support, we remove the docs.

I'll leave the decision to @rullzer @MorrisJobke @juliushaertl  whether to support the actual folder-based theming or drop it.

deusoz:
@MariusBluem Ok great, and thanks.  And like I mentioned, just for laughs.  I know you have the best roadmap in mind.  I had to make the statement a bit overstated to my knowledge of your project just because I couldn't let the pun go by unmentioned!

PR #968, opened 2018-11-29
Updated the "3rd party APPS" upgrade instructions

hanserasmus:
So there was a comment about "Needs better instruction".


MorrisJobke:
master 30a07608

PR #970, opened 2018-11-29
Added point 9 to remind user to copy theme

mnheia:
Added point 9 to remind user to copy theme


mnheia:
> @mnheia Thanks for this :) It seems that you are the first that used the new "Edit on Github" feature in our docs that is there since yesterday 🎉 Welcome in the community 😄

@MorrisJobke Thanks, happy to be a part of community. 😄

PR #980, opened 2018-11-30
Admin manual polishing

MorrisJobke:
I went over the first half of the admin manual and did a proof-read. I also restructured a bit some parts. We should most likely port this also to stable15.

Review this per commit to have it a bit more structured.

Remaining items are moved to https://github.com/nextcloud/documentation/issues/983


admin_manual/configuration_server/background_jobs_configuration.rst
```diff
@@ -64,17 +64,17 @@ Using the operating system cron feature is the preferred method for executing
...
 
   # crontab -u www-data -e
-  */15  *  *  *  * php -f /var/www/nextcloud/cron.php
+  */5  *  *  *  * php -f /var/www/nextcloud/cron.php
 
 You can verify if the cron job has been added and scheduled by executing::
 
   # crontab -u www-data -l
-  */15  *  *  *  * php -f /var/www/nextcloud/cron.php
+  */5  *  *  *  * php -f /var/www/nextcloud/cron.php
```
MorrisJobke: @rullzer @nickvergessen @schiessle @blizzz I used this quite often at instances and it just makes the overall speed of email deliveries for example just faster. Thus I changed it here as well. Just that you are aware of it ;)

MorrisJobke:
There is another idea in my head to better organize the admin manual:

* move "Server configuration"/"Server tuning" to "Installation"
* move "Server configuration"/"Hardening and security guidance" to "Installation"
* rename "Installation" to "Installation and server configuration" because it holds only the installation and configuration of the stack
* rename "Server configuration" to "Nextcloud configuration" because it holds then the configuration of Nextcloud itself


@rakekniven @tflidd @danielkesselberg @michag86 @hanserasmus @jasonbayton Does that make sense? I'm for myself often confused where to find what part. Especially the mix of Nextcloud config vs stack config (PHP, DB, webserver, ...) is confusing me.

MorrisJobke:
This is how the two sections look like right now (in this branch):

bildschirmfoto 2018-11-30 um 19 00 54
bildschirmfoto 2018-11-30 um 19 00 59


Obviously also feedback from others like @nickvergessen @MariusBluem @jospoortvliet @rullzer @skjnldsv or @blizzz is welcome. (just to mention a few)

admin_manual/configuration_server/background_jobs_configuration.rst
blizzz: Do the same on my instance :+1:

nickvergessen: I have 5 minutes since ever.
Maybe we should also recommend this in the admin settings/whats new

MorrisJobke:
I moved the remaining items over to two new tickets to not overload one PR. Please review 🚀 👩‍💻 👨‍💻

nickvergessen:
Nice work, the only disadvantage is that you can't simply change the version number in the URL to go to another version anymore. Can we readd empty files with the old name which are not included in the menus which simply say "Was moved to …" ?

MorrisJobke:
> Nice work, the only disadvantage is that you can't simply change the version number in the URL to go to another version anymore. Can we readd empty files with the old name which are not included in the menus which simply say "Was moved to …" ?

True, but is this really needed? The version switcher on the bottom left will give you the landing page and there you can search in the worst case. I would prefer a nicely structured manual over a backwards compatible URLs one.

hanserasmus:
Second that Morris. With new version switcher it makes a lot more sense to keep the structure as is. +1 from me.

hanserasmus:
The only thing that might worth looking at as well, is the following scenario:

When searching for a specific thing in Google/DuckDuckGo, for example theming, one is sometimes taken to NC13's page for theming. That was the search result given by the search engine. Now, if I click there I am taken to the theming for NC13, but when I use the navigator to move to 14 or stable, I am taken to their "landing page". It would be nice if it could "stick" and deliver the theming page for 14. Will make navigation easier and make it easier to track how stuff changed between version (if it did).

Does this make sense?

PR #996, opened 2018-12-03
set title to Configuration file

korelstar:
The current title ("app config") is misleading.


MorrisJobke:
This page should go completely because it is covered by the server docs already: https://docs.nextcloud.com/server/15/admin_manual/configuration_server/config_sample_php_parameters.html

Let me do that.

MorrisJobke:
See #997

PR #1007, opened 2018-12-04
Migrate release schedule to docs

MorrisJobke:
Migrates the release schedule from the wiki (https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) to the docs. In this way it's a bit more visible and people can find it. IMO the wiki is quite hidden and yet another place where to look. I would then replace the wiki page with a link to the docs.

localhost_8081__build_html_com_release_schedule html 1

@jospoortvliet @skjnldsv @rullzer @nickvergessen Does that make sense for you too?


MorrisJobke:
I updated according to the feedback by @jospoortvliet:

* removed the specific dates for future releases
* changed the text to include a link to the wiki with specific dates

localhost_8081_admin_manual__build_html_com_release_schedule html 1

PR #1008, opened 2018-12-05
Stable15, behaviour with empty LDAP user-attributes undocumented

victorbw:
As mentioned here [#12582](https://github.com/nextcloud/server/issues/12582#issue-383448404), there's no documentation about what happens when LDAP user-attributes are left empty:

grafik

Sugestion for somewhere in the top chapter of LDAP-documentation:

**Note**

When a LDAP object's name or surname is left empty, Nextcloud will treat it as an empty object and will not return any results from this user or AD-Object.

//optional: picture
grafik


MorrisJobke:
@blizzz This looks more like a issue, than a PR.

@victorbw Could you open a new ticket with your comment in here: https://github.com/nextcloud/documentation/issues/new

victorbw:
moved to #1010

PR #1011, opened 2018-12-05
Fix "erver" -> "server" typo

ChristophWurst:
(yes, I know this targets stable15, but when I change the merge base I'd pull in tons of unrelated commits)


MorrisJobke:
> (yes, I know this targets stable15, but when I change the merge base I'd pull in tons of unrelated commits)

No problem at all ;)

PR #1024, opened 2018-12-07
/.well-known/webfinger nginx rule for Social app

violoncelloch:
cc @nextcloud/social 

Signed-off-by: Jonas Sulzer <jonas@violoncello.ch>


jancborchardt:
Nice nice! Would any harm be done by just having it in there by default and saying "comment / remove the line if you don’t plan on using it"?

violoncelloch:
I'm not sure if the redirect would just lead to a 404 if there is nothing behind it or if it could become a potential security issue? @juliushaertl or @daita Do you know what happens if this redirect exists even tough Social is not installed? (How do we do it for the .htaccess? Is the line only added on the install of social or is it added by default for NC15 upwards?)

PR #1031, opened 2018-12-10
Update contacts.rst

bakkegaard:
Add note explaining that the Contacts app only supports vCard version 3.0 and 4.0


skjnldsv:
@bakkegaard thanks a lot!

PR #1032, opened 2018-12-11
New File Sharing order and configuration options

MariusBluem:


tflidd:
What happened to the mail notification for shared files, is this not available any more, and if not, is it enabled/disabled?

skjnldsv:
@tflidd I haven't seen this checkbox in a long time! I don't think this was an option on the mast nextcloud versions.

MorrisJobke:
> What happened to the mail notification for shared files, is this not available any more, and if not, is it enabled/disabled?

It is now on by default and respects the settings of the recipient. It just didn't made sense to define if a notification email should be send on the sender side but it should be handled on the recipient side.

PR #1037, opened 2018-12-16
Add PublicPage documentation

rullzer:


developer_manual/app/index.rst
```diff
@@ -7,7 +7,6 @@ App development
 
    intro
    tutorial
-   npm
```
skjnldsv: Hum?

rullzer: mmm indeed

PR #1051, opened 2018-12-19
Add documentation for the locale setting

tcitworld:
See https://github.com/nextcloud/server/pull/5623

### DIFF
--- a/admin_manual/configuration_server/language_configuration.rst
+++ b/admin_manual/configuration_server/language_configuration.rst
@@ -60,10 +60,10 @@ If you want to force a specific locale, users will no longer be able to change
 their locale in the personal settings. You can set a **force_locale** parameter
 in the :file:`config/config.php`.
 
-'::
+::
 
     "force_locale" => "en_US",
 
-.. note:: Please check `Transifex language codes
-   <https://www.transifex.com/explore/languages/>`_ for the list of valid locale
-   codes (aside the language, between parentheses).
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/language_configuration.rst
```diff
@@ -36,3 +36,34 @@ this value can be set to ``true`` instead of a language code.
...
+
+If you want to force a specific locale, users will no longer be able to change
+their locale in the personal settings. You can set a **force_locale** parameter
+in the :file:`config/config.php`.
+
+'::
+
+    "force_locale" => "en_US",
+
+.. note:: Please check `Transifex language codes
```
georgehrke: Copy and Paste leftover? ;) 

What locale should one check? The moment js locales?

skjnldsv: Well, the data are provided by moment indeed, so we should probably check moment instead, you're right!

tcitworld: Anything better than https://github.com/moment/moment/tree/develop/locale ?

georgehrke: I think i would prefer to link to the stable, used version: https://github.com/moment/moment/tree/2.18.1/locale

But then we would have to update the link every time we update moment.js

tcitworld: Does it matter much in the end ? Maybe remove the whole note ?

PR #1055, opened 2018-12-19
Move recommendation to 18.04 #1030

### LINKED ISSUE #1030: Move recommendation to 18.04
We still recommend 16.04 in the docs. 18.04 shoud be used.

fgsl:
Fix #1030

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -9,7 +9,7 @@ install `Snap Packages <http://snapcraft.io/docs/core/install/>`_. See
 In case you prefer installing from the source tarball, you can setup Nextcloud
 from scratch using a classic LAMP stack (Linux, Apache, MySQL/MariaDB, PHP).
 This document provides a complete walk-through for installing Nextcloud on
-Ubuntu 16.04 LTS Server with Apache and MariaDB, using `the Nextcloud .tar
+Ubuntu 18.04 LTS Server with Apache and MariaDB, using `the Nextcloud .tar
 archive <https://nextcloud.com/install/>`_.
 
 .. note:: Admins of SELinux-enabled distributions such as CentOS, Fedora, and
@@ -169,10 +169,10 @@ If ``mod_webdav`` is enabled you must disable it for Nextcloud. (See
 
 .. _ubuntu_installation_label:
 
-Example installation on Ubuntu 16.04 LTS server
+Example installation on Ubuntu 18.04 LTS server
 -----------------------------------------------
 
-On a machine running a pristine Ubuntu 16.04 LTS server, you have two options:
+On a machine running a pristine Ubuntu 18.04 LTS server, you have two options:
 
 You can either install the Nextcloud `Snap Package <http://snapcraft.io/>`_, just run the
 following command in a terminal::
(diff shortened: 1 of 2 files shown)


rullzer:
But are the installation steps actually the same on 18.04 ?

J0WI:
I think we should switch to version independent packages in the installation guide (e.g. `libapache2-mod-php` instead of `libapache2-mod-php7.2`, `php-curl` instead of `php7.2-curl` etc.).
There are much more of these in `source_installation.rst`.

fgsl:
@J0WI, I suggest to keep section "Example installation on Ubuntu 16.04 LTS server" and add right after "Example installation on Ubuntu 18.04 LTS server", but in other issue.  What do you think?

J0WI:
I fine with this. But after switching to version independent packages names there shouldn't be a big difference between the installation steps on 16.04 and 18.04.

skjnldsv:
@fgsl let's keep it like that. Simpler is cleaner :)
Could you make sure the various path are the same from 16.04 to 18.04?

fgsl:
@skjnldsv, in Ubuntu 18.04, PHP packages are in version 7.2 (for instance, libapache2-mod-php7.2). However, the virtual package, without version, is the same. If suffix "7.0" is removed from current Ubuntu 16.04 installation section, it's enough to change title to "Example installation on Ubuntu 16.04 LTS or 18.04 LTS server".

skjnldsv:
@fgsl Well, I think we should only suggest 18.04 and php7.2 then.
Let's update all the path and references ? :)

fgsl:
@rullzer, @J0WI, what do you think? 

1) I merge this pull request, only with recommendation change?
2) I update this PR, including change in Ubuntu install steps?
3) I change Ubuntu install steps replacing packages of 16.04 by 18.04 (like to move php 7.0 to php 7.2) or to let generic steps to both versions, using virtual packages?

skjnldsv:
Please change Ubuntu install steps replacing packages of 16.04 by 18.04 (like to move php 7.0 to php 7.2).

fgsl:
@skjnldsv, I have changed Ubuntu install steps.

admin_manual/installation/source_installation.rst
```diff
@@ -183,9 +183,9 @@ Or you can use .deb packages to install the required and recommended modules for
...
 
-    apt-get install apache2 mariadb-server libapache2-mod-php7.0
-    apt-get install php7.0-gd php7.0-json php7.0-mysql php7.0-curl php7.0-mbstring
-    apt-get install php7.0-intl php7.0-mcrypt php-imagick php7.0-xml php7.0-zip
+    apt-get install apache2 mariadb-server libapache2-mod-php7.2
+    apt-get install php7.2-gd php7.2-json php7.2-mysql php7.2-curl php7.2-mbstring
+    apt-get install php7.2-intl php-imagick php7.2-xml php7.2-zip
 
 * This installs the packages for the Nextcloud core system.
   ``libapache2-mod-php7.0`` provides the following PHP extensions: ``bcmath bz2
```
J0WI: This should also be changed to `libapache2-mod-php7.2`

fgsl: Right. I changed it.

Issue #899, opened 2018-10-12
redirect /.well-known/ to correct port in nginx


macmeck:
Currently, the documentation for service discovery in nginx proposes to redirect /.well-known/ to proper paths. 

```
    location = /.well-known/carddav {
      return 301 $scheme://$host/remote.php/dav;
    }
    location = /.well-known/caldav {
      return 301 $scheme://$host/remote.php/dav;
    }

```
This only works for the standard ports but fails if Nextcloud runs on a different port. A dynamic port parameter should be added to the redirect but I did not yet find out how.

MorrisJobke:
@josh4trunks Do you have an idea for this? This is basically then the case when there is a reverse proxy infront of Nginx as well :/

josh4trunks:
I'm confused what a proxy in front of NGINX has to do with return 301. Isn't a return 301 pretty much handled by the client doing a new request for the updated URI.

@macmeck does this work for your scenario? ``return 301 $scheme://$host:PORT/remote.php/dav;``

MorrisJobke:
> I'm confused what a proxy in front of NGINX has to do with return 301. Isn't a return 301 pretty much handled by the client doing a new request for the updated URI.

It is because the Nginx port is a different one, than the one of the reverse proxy in front of it. Thus the redirect goes to the port of Nginx instead of the reverse proxy. See https://github.com/nextcloud/server/issues/11789 for more details.

josh4trunks:
@macmeck if there is a reverse proxy in front of NGINX, NGINX would not know what port that reverse proxy is using. There is the PROXY Protocol, which allows the reverse proxy to forward additional information about the client connection including client IP:port and server (reverse proxy) IP:port.
https://www.haproxy.com/blog/haproxy/proxy-protocol/

NGINX does support this, but currently does not save the server IP or port to a variable. See a feature request here.
https://trac.nginx.org/nginx/ticket/1206

There are ways to hack this, like forwarding an HTTP header from the reverse proxy, then mapping that in NGINX. This is protocol specific, and not as efficient as if we had full PROXY Protocol support.
```
map $http_x_forwarded_proto $host_port {
	default 80;
	https 443;
}
```

macmeck:
Hm, I just realized that the setup is even worse. Port forwarding is already done by the router upfront. So the machine running nginx and Nextcloud already gets the request on port 443. 

So to fix this, we need to find out the port the client used in its initial request.

josh4trunks:
[...]

PR #1114, opened 2019-01-04
Hardcode https redirects

PriceChild:
Other https settings are already selected (redirecting 80 to 443) so why not hardcode the redirects too?

The existing $scheme breaks "System Overview" checks if nextcloud is placed behind an ssl terminating proxy, as it'll return http links which see a further redirect.

Who doesn't enable https?


MorrisJobke:
See the discussion in #899

PR #1115, opened 2019-01-04
Add redirect for admin-bigint-conversion

danielkesselberg:
Used here https://github.com/nextcloud/server/blob/17b2827bbf20691dc59721a9c61a225c5fb4e0de/core/js/setupchecks.js#L350 but redirect is missing.


MorrisJobke:
> I guess backport to stable15 should be enough. Check for missing bigint conversions has been added at 2018-12-04.

Fair enough - at least it doesn't hurt xD

PR #1113, opened 2019-01-04
Remove outdated ArchLinux documentation

MorrisJobke:
cc @skjnldsv @ChristophWurst @juliushaertl Or do you want to add updated docs?

PR #1116, opened 2019-01-04
Remove outdated ArchLinux steps

MorrisJobke:
Same as #1113


MorrisJobke:
> Is it really outdated?
> 
> We should favour pip install I think: `pip install -r requirements.txt`
> So it still requires some deps?

Yes - It didn't worked back in early 2017 - I doubt that it will work again ;) I wrote them quite some time ago and it wasn't updated (and also wasn't used, because we didn't got any complaints in the past 2 years).

skjnldsv:
It worked for me :woman_shrugging:

PR #1121, opened 2019-01-06
Centos php72

### LINKED ISSUE #1058: Update RHEL/CentOS docs to use PHP 7.1+
PHP 7.0 is EOL:
https://docs.nextcloud.com/server/stable/admin_manual/installation/php_70_installation.html

michag86:
Fixes #1058

@J0WI @skjnldsv @hanserasmus Some who can test this please?

I've tested it on CentOS 7.6

@MorrisJobke are doc changes also backported? I think backporting for stable15 make sense.


MorrisJobke:
> @MorrisJobke are doc changes also backported? I think backporting for stable15 make sense.

Sure - makes sense 👍

MorrisJobke:
Waiting for some feedback from @hanserasmus

hanserasmus:
Will try to create a quick vm tomorrow with these instructions and report
back. Give me 24h.

hanserasmus:
Lastly, after adding those packages mentioned above, I get the `OPcache settings` warning. It might be useful to the users if we just added the path where these settings should be configured to the CentOS install. These settings seems to be distro specific, and based on the community forum is something that has created a LOT of confusion. Maybe just drop a hint right at the end stating:

If you find a warning on the Admin -> Settings page about OPcache warnings, you could check your settings in 
> /etc/opt/rh/rh-php72/php.d/10-opcache.ini

MorrisJobke:
Thanks for the feedback @hanserasmus! It makes a lot of sense 👍 @michag86 Could you update your PR and then we can get this into the documentation 🙌

michag86:
Does it really mage sense to add rh-php72-php-opcache and sclo-php72-php-pecl-imagick?
This is just a how-to for upgrading php. 

rh-php72-php-intl was just added to avoid the annoying errors on the log.

And I added a general note to configuration changes to the new installed php version.

@hanserasmus your opinion?

MorrisJobke:
intl is required, but the other two don't need to be included in here IMO.

hanserasmus:
> Does it really mage sense to add rh-php72-php-opcache and sclo-php72-php-pecl-imagick?
This is just a how-to for upgrading php.

I feel that it is. As you guys stated earlier, you don't like the Webtatic repo. So if that doesn't bother you anymore in the installtion procedure, then no, this does not have to be included. BUT, if you want to convert the CentOS installation back to use the SCL repository instead of Webtatic, then we need to add this. As the install and the upgrade procedure should be consistent.

MorrisJobke:
Let's go with this now and iterate on it. Thanks @michag86 👍

MorrisJobke:
@hanserasmus @michag86 How do you install redis with this SCL?php-redis seems to be only available in the php72 packages.

michag86:
@MorrisJobke I use sclo-php72-php-pecl-redis4

PR #923, opened 2018-11-06
Groupware Clients Part 1

MariusBluem:
@georgehrke @nextcloud/documentation 

Signed-off-by: Marius Blüm <marius@lineone.io>

PR #1126, opened 2019-01-10
Update sync_android.rst

wgroenewold:
Changed name of DAVDroid and added reference to Google Play Store

### DIFF
--- a/user_manual/pim/sync_android.rst
+++ b/user_manual/pim/sync_android.rst
@@ -5,7 +5,7 @@ Synchronize Android
 Address book and Calendar
 -------------------------
 
-#. Install DAVdroid (possibly through F-Droid)
+#. Install DAVx⁵ (possibly through F-Droid or Google Play Store (latter wil set you back $3,99)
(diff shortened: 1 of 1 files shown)


user_manual/pim/sync_android.rst
```diff
@@ -5,7 +5,7 @@ Synchronize Android
 Address book and Calendar
 -------------------------
 
-#. Install DAVdroid (possibly through F-Droid)
+#. Install DAVx⁵ (possibly through F-Droid or Google Play Store (latter wil set you back $3,99)
```
MorrisJobke:
```suggestion
#. Install DAVx⁵ (possibly through F-Droid or Google Play Store (latter will set you back $3,99)
```

MariusBluem:
Thanks for your Pull Request! We fixed this by merging #923. We are now linking to DAVx⁵ direct website and let people choose whether they want to download the app from Google Play Store, F-Droid or directly via an APK file :)

PR #1127, opened 2019-01-10
Fix indefinite article before Nextcloud (an->a)

uusijani:


uusijani:
@juliushaertl Is backporting for docs also done [as per the manual](https://docs.nextcloud.com/server/stable/developer_manual/general/backporting.html)? It seems pretty straightforward, so I think I can produce those pull requests too, if needed.

skjnldsv:
@uusijani not for the documentation.
And I think the backporting docs should also be updated. Thought @MorrisJobke @rullzer ?
We do not mention Frank every time now, do we?

MorrisJobke:
We need to update the docs about our new bot and also that @rullzer and me now decide about the backports

PR #1181, opened 2019-01-20
Add another table not converted from sqlite to mysql

Yvan-Masson:
Hi,

I just converted an SQLite database to MySQL, and the tool reported that `oc_privatedata` table would not be converted: I suggest to add it to the documentation.

Regards,
Yvan

### DIFF
--- a/admin_manual/configuration_database/db_conversion.rst
+++ b/admin_manual/configuration_database/db_conversion.rst
@@ -69,5 +69,6 @@ Here is a list of known old tables:
 * oc_media_songs
 * oc_media_users
 * oc_permissions
+* oc_privatedata
 * oc_queuedtasks
 * oc_sharing
--- a/admin_manual/configuration_database/db_conversion.rst
+++ b/admin_manual/configuration_database/db_conversion.rst
@@ -69,6 +69,6 @@ Here is a list of known old tables:
 * oc_media_songs
 * oc_media_users
 * oc_permissions
-* oc_privatedata
+* oc_privatedata - this table was later added again by the app `privatedata` (https://apps.nextcloud.com/apps/privatedata) and is safe to be removed if that app is not enabled
 * oc_queuedtasks
 * oc_sharing


admin_manual/configuration_database/db_conversion.rst
```diff
@@ -69,5 +69,6 @@ Here is a list of known old tables:
 * oc_media_songs
 * oc_media_users
 * oc_permissions
+* oc_privatedata
```
MorrisJobke: It's added by the private data app: https://github.com/nextcloud/privatedata/blob/master/lib/Migration/Version10000Date201707202147.php#L43

MorrisJobke: We removed it from the server core, but added it to an app. Fine by me with stating it here, but maybe add a hint, that it could be added by the privatedata app?

Yvan-Masson: I do not fully understand this point, I must let you improve or reject this pull modification. Thanks for all your work!

MorrisJobke:
```suggestion
* oc_privatedata - this table was later added again by the app `privatedata` (https://apps.nextcloud.com/apps/privatedata) and is safe to be removed if that app is not enabled
```

skjnldsv: @Yvan-Masson any news here?

Yvan-Masson: @skjnldsv I do not understand if you are waiting something especially from me or not, feel free to ask precisely if I need to do something.

MorrisJobke: I used the "suggested change" feature of GitHub here. I guess there is a button to accept this suggested change. Then this can go in 👍 

But as this is a beta feature it is maybe not visible to you. Do you see this above @skjnldsv's reply?

bildschirmfoto 2019-02-15 um 10 24 14

Yvan-Masson:
@MorrisJobke You are right, there was a button to accept this commit. Thanks for the clarification.

Yvan-Masson:
Woups, closed before merge…

MorrisJobke:
> @MorrisJobke You are right, there was a button to accept this commit. Thanks for the clarification.

Don't worry :) Thanks for your help here :)

PR #1186, opened 2019-01-22
Suppress confirmation

jzdm:
Shared hostings without shell access make the confirmation step impossible. Therefore you can add named argument and suppress the confirmation message. Then the conversion works even from one time cronjobs.
See: https://github.com/nextcloud/server/issues/12976

### DIFF
--- a/admin_manual/configuration_database/bigint_identifiers.rst
+++ b/admin_manual/configuration_database/bigint_identifiers.rst
@@ -20,5 +20,11 @@ or otherwise ask for confirmation, before performing the heavy actions::
     This can take up to hours, depending on the number of files in your instance!
     Continue with the conversion (y/n)? [n]
 
+in a shared hosting environment you can add
+
+    --no-interaction
+
+as an argument to suppress the confirmation.
+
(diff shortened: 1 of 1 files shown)


jzdm:
There are hosters [1] which do not give you shell access (or no PHP on the shell) but enable you to setup a cronjob. The cronjob is then able to call a PHP script as for example `occ`.
But using a cronjob, there is no way for the user to further interact with the script.
In that case you need to disable interaction.

[1] For example bitpalast: https://preiswerter-webserver-de.bitpalast.net

nickvergessen:
wow, so you set it up as a cron job one time? That's quite a nice trick. :tada: 

Never the less, the description you add is wrong and confusing. I'm on a shared host and can run the terminal just fine.
I think we should advertise your workaround in a different way, but not sure what would be the best.

jzdm:
Yes, using a cronjob in such a setup is quite a nice workaround 😉

> the description you add is wrong and confusing

You were just a few minutes faster than me. I just wanted to add this as it's truly not that clear.
Maybe the sentence can be something like:
_The confirmation message can be suppressed adding the following command …_

PR #594, opened 2017-11-07
Fix nginx comment syntaxwise

MorrisJobke:
Fixes #592 

@oclair thanks for the report.

PR #1200, opened 2019-01-25
Nginx config: Remove line-breaks from STS header

jacksgt:
This patch removes the incorrect line-breaks in the Strict-Transport-Security Headers.
Just removing the comment at the beginning of the line (and therefore having a line-break in the header field) leads to various kinds of failures across different clients and browsers.

While this syntax error had already been fixed in https://github.com/nextcloud/documentation/pull/594 , it apparently snuck its way back in.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -61,8 +61,7 @@ webroot of your nginx installation. In this example it is
       # Add headers to serve security related headers
       # Before enabling Strict-Transport-Security headers please read into this
       # topic first.
-      # add_header Strict-Transport-Security "max-age=15768000;
-      # includeSubDomains; preload;";
+      # add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload;";
       #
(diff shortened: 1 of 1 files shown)


skjnldsv:
I think it was here because people would want to use `Strict-Transport-Security "max-age=15768000` without the `includeSubDomains; preload` which can lead to undesirable issues :)

jacksgt:
@skjnldsv  The way it's currently written is very ambiguous and not only in-experienced users struggle with it.

skjnldsv:
@jacksgt agree, but it's on purpose! That's what I was saying. :)
Anyway, let's wait for the other reviewers :wink:

MorrisJobke:
> I think it was here because people would want to use `Strict-Transport-Security "max-age=15768000` without the `includeSubDomains; preload` which can lead to undesirable issues :)

Then it should be written there two times - one time with the subdomains included and one time without.

cc @josh4trunks for veto right. I merge this in the meantime.

PR #1250, opened 2019-02-12
rh-php72-php-pecl-apcu is required on RH/Centos

rullzer:


MorrisJobke:
cc @michag86 @hanserasmus

PR #1254, opened 2019-02-14
Changed CentOS install to SCL from webtatic

hanserasmus:
Fixed the repo issue as previously discussed. Please note I added `imagick` package now as default. IMHO it is of no use to warn a user about it on the admin page if it is not to be used. This how-to should act as a one-stop page for them to get a basic Nextcloud up and running, without issues. (Excluding SSL etc which is more advanced).


MorrisJobke:
cc @michag86 😉

PR #1260, opened 2019-02-18
add exception for ocm-provider end-point

schiessle:
update documentation according to https://github.com/nextcloud/server/pull/14255


MorrisJobke:
Please add a setup check for this, because otherwise nobody will notice that there was an update to the Nginx config.

PR #1270, opened 2019-02-19
Only render gif in HTML

MorrisJobke:
Alternative approach to #1266 

@skjnldsv Maybe this is a better and more explicit way of dealing with this problem.

cc @MariusBluem


MorrisJobke:
We should go for this way, because the other one causes this:

```
! Package pdftex.def Error: File `calendar_create-gif-converted-to.png' not found: using draft setting.
```

MorrisJobke:
stable13 6a0315fe

PR #1273, opened 2019-02-21
Added undocumented 'jpeg_quality' setting to the docs

iGamerHD:
This option is really useful for people like me, who like the preview folder to be half as large as the orginal image folder.
I found this setting by coincidence and added it to the documentation.

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -1067,6 +1067,15 @@ Defaults to ``50`` megabytes
 
 ::
 
+	'jpeg_quality' => 90,
+	
+By default Nextcloud generates previews with a JPEG quality of 90 %.
+If you want to save disk space and network traffic, lower this value to about 70-50.
+
+Defaults to ``90`` %
+
+::
+
 	'preview_libreoffice_path' => '/usr/bin/libreoffice',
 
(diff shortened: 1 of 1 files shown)


iGamerHD:
It would also be great to inform the user, that by default previews are always generated by power of 4 and which sizes are needed/ requested from the different client applications.
But I'm not enough into this topic to make this clear.

skjnldsv:
@MorrisJobke aren't this doc auto-generated?

MorrisJobke:
> @MorrisJobke aren't this doc auto-generated?

Usually - let me check it, but the monitoring didn't triggered anything.

MorrisJobke:
Should work: https://github.com/MorrisJobke/nextcloud-config-converter/blob/master/updateConfig.sh#L14

MorrisJobke:
```
...
error: gpg failed to sign the data
fatal: failed to write commit object
```

🤔

MorrisJobke:
Working again: 0a5977d58c52e472ed9ad27529f82ed52be9df0f

MorrisJobke:
I just looked at the code and `jpeg_quality` is not there as system setting at all. Where do you have this from?

juliushaertl:
It actually is an app value so setting in config.php shouldn't work at all.

https://github.com/nextcloud/server/blob/master/lib/private/legacy/image.php#L384

PR #1282, opened 2019-03-02
Fix a link to point to correct version and remove TODO entry

frlan:
This PR corrects a link to point to most recent version of documentation and removes a leftover TODO-marker


MorrisJobke:
Thanks for finding this 👍

PR #1296, opened 2019-03-10
add expected database, user and password for unit tests

chiefbrain:
see https://github.com/nextcloud/server/issues/13803

### DIFF
--- a/developer_manual/core/unit-testing.rst
+++ b/developer_manual/core/unit-testing.rst
@@ -117,7 +117,22 @@ adjust your php.ini and file rights.
 
 Running unit tests for the Nextcloud core project
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


developer_manual/core/unit-testing.rst
```diff
@@ -117,7 +117,7 @@ adjust your php.ini and file rights.
 
 Running unit tests for the Nextcloud core project
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-The core project provides a script that runs all the core unit tests using different database backends like sqlite, mysql, pgsql, oci (for Oracle)::
+The core project provides a script that runs all the core unit tests using different database backends like sqlite, mysql, pgsql, oci (for Oracle). A database oc_testuser which can be accessed with user oc_testuser and password ownloud is expected::
```
nickvergessen: isnt the database name oc_autotest?

MorrisJobke: Yes - the password is also `owncloud` (instead of `ownloud`)

https://github.com/nextcloud/server/blob/c32c91d1da11444619f545e18c2c39732326c8a8/autotest.sh#L16-L17

chiefbrain: new words: 15
errors: 3 (since the user is also oc_autotest)
Good quota.

PR #1298, opened 2019-03-11
add instructions to use revised filetype icons

deusoz:


J0WI:
Please note that the theming feature is deprecated: #946

skjnldsv:
Only done this for backports.
#946 will take over

PR #1317, opened 2019-03-16
Add another common problem

luvarqpp:
After following two different tutorials on Internet, I have encountered login loop after fresh install. It had taken me some time to fix this problem, so I am proposing this troubleshooting part here.


skjnldsv:
I'm guessing it was merged into stable16 at the time

PR #1337, opened 2019-03-27
add server_port variable to well-known redirects

mrplow87:
if Nextcloud is used on a custom HTTPS port, the well-known redirects do not work, because the server port is not included in the redirect.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -42,7 +42,7 @@ webroot of your nginx installation. In this example it is
       listen [::]:80;
       server_name cloud.example.com;
       # enforce https
-      return 301 https://$server_name$request_uri;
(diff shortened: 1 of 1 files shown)


t123yh:
I believe changing `$host` to `$http_host` is better because it avoids adding a redundant port number when you are using a standard port.

MorrisJobke:
cc @josh4trunks

josh4trunks:
To be consistent we may want to explicitly state the ports in the two http location redirects.

``
return 301 https://$server_name:443$request_uri;
``

MorrisJobke:
> To be consistent we may want to explicitly state the ports in the two http location redirects.

@mrplow87 Do you want to integrate this in this PR as well?

skjnldsv:
@mrplow87 Bump? :)

PR #1338, opened 2019-03-28
LDAP configuration: clarify effect of "Turn off SSL certificate verification" checkbox

jacksgt:
This patch adds an alternate possibility to disable ssl certificate
verification in LDAP connections for a particular host.

As discussed in https://github.com/nextcloud/server/issues/14821

For reference: https://stackoverflow.com/a/13593766/4159145


jacksgt:
@blizzz Yes, you're right. I had another look at [ldap.conf(5)](https://manpages.debian.org/stretch/libldap-common/ldap.conf.5.en.html).

I have updated the patch accordingly.

blizzz:
thanks @jacksgt !

PR #1339, opened 2019-03-29
Modify host variable in NGINX config

t123yh:
`$host` doesn't contain port field, while `$http_host` contains port which is useful when you use a non-standard port.


skjnldsv:
Duplicate of https://github.com/nextcloud/documentation/pull/1337
Though you both have differents approachs. If you think yours have some advantages, please discuss it over there :)
Thanks for your help :v:

PR #1374, opened 2019-04-13
Add section on automatically updating Nextcloud via cron

Viveur:


tflidd:
Not sure if that is a good idea. Depending on the branches, you can quickly push a major upgrade that are more likely to fail. We can put it like this without warning and also propose a way that this works only on minor upgrades. Also the result should be mailed to the admin in case something goes wrong.

MorrisJobke:
> Not sure if that is a good idea. Depending on the branches, you can quickly push a major upgrade that are more likely to fail. We can put it like this without warning and also propose a way that this works only on minor upgrades. Also the result should be mailed to the admin in case something goes wrong.

Shouldn't cron automatically mail to the root account and thus the admin?

In general: I have this in mind for a very long time. Also I know of some people already doing this successfully. I would add a big warning there for now, that this is totally on own risk, but it shouldn't break. For the Nextcloud 17 cycle I plan to add more testing for this so that it is save to run it that way at some point in time.

Viveur:
Apologies about the delay and thanks for the comments.

Have added a warning about the upgrade process. It should hopefully adopt the same CSS used for other documentation warnings.

With regards to email notification cron will automatically mails the root account for every action. 
I did contemplate including a MAILTO=admin@example.com line; however, this would result in cron sending an email with the text below every time the script is run, regardless of whether an update is available or not.

```
Nextcloud Updater - version: v14.0.2RC2-7-g57268cb
Current version is 15.0.7.
No update available.
Nothing to do. 
```

Happy to amend should there be any suggestions

Viveur:
Update: Installed Sphinx to ensure the page looks as it should and also reformatted the instructions to better match the style used on other pages. I also added a reference to systemd.

Happy with the format and styling, though as before any suggestions are welcome.

skjnldsv:
Bump! :)

kesselb:
I would not do that hence not recommending it. On failure it will leave your instance in a broken state. Having such a section (even with the warnings) will leave the impression that this is a common / supported setup.

skjnldsv:
Let's close then :)

PR #1393, opened 2019-04-18
Fix npm dev build instructions

skjnldsv:


skjnldsv:
Ah, 15 doesn't have the page, that's fine.

PR #1434, opened 2019-05-06
updated release schedule for 17

wiswedel:


admin_manual/installation/source_installation.rst
```diff
@@ -100,7 +100,7 @@ If you get a result, the module is present.
 
 Required:
 
-* PHP (>= 7.0, 7.1 or 7.2)
+* PHP (7.0, 7.1, 7.2 or 7.3)
```
nickvergessen:
```suggestion
* PHP (7.1, 7.2 or 7.3)
```

nickvergessen: 7.0 is not supported in master

admin_manual/release_schedule.rst
```diff
@@ -7,6 +7,8 @@ This page gives an overview of the currently supported and released versions as
 +--------+-------------------+------------------+--------------------------+---------------------------+
 |        | release date      | end of life      | latest version           | next version              |
 +========+===================+==================+==========================+===========================+
+| **17** | 2019-08-08        | 2020-08          |                       -  | *17.0.0 (2019-08-08)*     |
++--------+-------------------+------------------+--------------------------+---------------------------+
```
MorrisJobke: This was on purpose - we don't want to put major release dates into the manual, as this is a document that is also shipped in the release and thus easily out of date.

wiswedel: > This was on purpose - we don't want to put major release dates into the manual, as this is a document that is also shipped in the release and thus easily out of date.

Ah ok. didn't know that.

PR #1457, opened 2019-05-21
Update mysql_4byte_support.rst

cyBea:
I tried to follow the manual but my case (empty query set because INFORMATION_SCHEMA.INNODB_SYS_TABLES is empty) was not covered.  (Happend with debian/stretch mariadb-server-10.1)
All following steps and emoji in filename worked just fine.


skjnldsv:
@nextcloud/server-triage

blizzz:
@cyBea so that's from before installing nextcloud? and thus no tables were created yet.

cyBea:
No, it is a many times updated nextcloud. The original installation was an owncloud.

blizzz:
It's strange, according to the doc it should be Antelope or Barracuda, but not empty. https://mariadb.com/kb/en/library/xtradbinnodb-file-format/

cyBea:
OK - now I checked again on a different installation. The reason for my queryset to be empty was that I used my unprivileged "nextcloud" database user.  So my pull request should be rejected but a pull request "pivileged user needed" would be nice?

PR #1460, opened 2019-05-23

grueneedv:
Added "dummy" line to get proper Syntax Highlightning for Section Redis and more important get rid of typographic quotation marks in Apache Config.

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -311,6 +311,8 @@ After you have done this, make sure you create a database with a username and pa
 
 **Redis**
 
+Install as usual:
+
     yum install -y redis
     systemctl enable redis.service
     systemctl start redis.service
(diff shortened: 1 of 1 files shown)


admin_manual/installation/source_installation.rst
```diff
@@ -438,23 +440,18 @@ Then enable the newly created site::
 
 On CentOS/RHEL, create a virtualhost :file:`/etc/httpd/conf.d/nextcloud.conf`and add the following content to it:
```
MorrisJobke: I guess the following block should be a code block. To trigger this `::` is needed like that:

```suggestion
On CentOS/RHEL, create a virtualhost :file:`/etc/httpd/conf.d/nextcloud.conf`and add the following content to it::
```

skjnldsv:
Is this still relevant? It looks good to me?
https://docs.nextcloud.com/server/16/admin_manual/installation/source_installation.html#apache-web-server-configuration

Please answer to reopen the pr if you think we should still have this in :)

Issue #1069, opened 2018-12-24
Rewrite Rule sample typo error


JochenHWezel:
At https://docs.nextcloud.com/server/14/admin_manual/issues/general_troubleshooting.html#service-discovery there is a typo
```
RewriteRule ^\.well-known/host-meta /nextcloud/public.php?service=host-meta [QSA,L]
RewriteRule ^\.well-known/host-meta\.json /nextcloud/public.php?service=host-meta-json [QSA,L]
RewriteRule ^\.well-known/webfinger /nextcloud/public.php?service=webfinger [QSA,L]
RewriteRule ^\.well-known/carddav /nextcloud/remote.php/dav/ [R=301,L]
RewriteRule ^\.well-known/caldav /nextcloud/remote.php/dav/ [R=301,L]
```
needs to be
```
RewriteRule ^/\.well-known/host-meta /nextcloud/public.php?service=host-meta [QSA,L]
RewriteRule ^/\.well-known/host-meta\.json /nextcloud/public.php?service=host-meta-json [QSA,L]
RewriteRule ^/\.well-known/webfinger /nextcloud/public.php?service=webfinger [QSA,L]
RewriteRule ^/\.well-known/carddav /nextcloud/remote.php/dav/ [R=301,L]
RewriteRule ^/\.well-known/caldav /nextcloud/remote.php/dav/ [R=301,L]
```

Only with the / in the beginning, the rules are working.

MorrisJobke:
@JochenHWezel Could you directly change it in https://github.com/nextcloud/documentation/edit/master/admin_manual/issues/general_troubleshooting.rst. Then we can directly take the changes and don't need to redo them. Thanks :)

MorrisJobke:
The link to this is also at the top of every documentation page 😉

JochenHWezel:
@MorrisJobke sorry - I don't find the reuqired spare time for this handling with sources and forks etc. - could you please do it?

MorrisJobke:
> @MorrisJobke sorry - I don't find the reuqired spare time for this handling with sources and forks etc. - could you please do it?

My intention was to guide you, because it isn't really that hard anymore. You literally click that link I posted (or in the documentation at the top), then you do the change and you press "Save" together with a message why this change was made. No need to handle forks or check out files locally. :wink

MorrisJobke:
Fix is in #1100

JochenHWezel:
[...]

PR #1100, opened 2019-01-02
Add slashes to rewrite rules

MorrisJobke:
Fixes #1069

PR #1480, opened 2019-06-08
Correct error in .htaccess, revert #1069

stecklars:
The forward slashes are only needed if the rewrite rules are being put directly into the config file of the apache virtual host. They should not be added to the .htaccess file, as in this case the redirect is not working (this has been tested by me on multiple systems).

You can also see in the discussion of the bug report #1069 after it was merged, that the reporter actually put the rules into his apache conf, instead of a .htaccess and that somebody else was reporting, that it was only working without the forward slashes in a .htaccess file.

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -271,11 +271,11 @@ document root of your Web server and add the following lines::
 
     <IfModule mod_rewrite.c>
       RewriteEngine on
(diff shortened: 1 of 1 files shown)


stecklars:
I think this needs to be backported to be visible in the current versions of the documentation? @MorrisJobke @skjnldsv

skjnldsv:
13 and 14 were not backported on #1100

Issue #216, opened 2016-10-19
Documentation update and translation with transifex


privatereese:
Would appreciate an update of the documentation and translation via transifex.

Major goal is an update of the documentation.
Minor goal is transifex translation.

Would start with an update of the english documentation and then at least create a german translation via transifex.
As I am new to nextcloud i would appreciate every help i can get.
I assume that this should be handled with high priority as it helps spreading nextcloud to users and could ease handle and  supporting effort.

esprit-libre:
Hello,
also interested for creating and contributing to french translation of user manual (admin and dev manual doesn't need to be translated in french).
My need is to share this doc to my customers, mostly french speaking users.
Any chance to have this?
Best regards.

dhebert91:
Hello,
I'm also interested with a french translation of the user manual and i can contribute to it.
Regards

guidtz:
@esprit-libre @dhebert91 cela vous intéresserait de mettre en place cette traduction et une fois qu'elle est en place faire un pull request pour la proposer ? On pourrait partir sur la version 12 et se répartir le travail de traduction / relecture. 
Je suis dans le même besoins de fournir une doc utilisateur en français à mes clients et pas obligatoirement le temps de tout réécrire.
J'ai commencé à regarder un peu, il faut modifier les fichiers rst et générer le pdf, voir adapter les screenshots si besoins. 
guidtz 
@esprit-libre on c'est déjà croisé lors des RRLL à Nantes il me semble : Guillaume Chéramy, société Hisyl faisant partie d'Alliance Libre

pierreozoux:
@jancborchardt how can we get started on that?

I'm seeing [that](https://howto.disroot.org/en/nextcloud/introduction) also, and I'm a bit sad that this doc is not on Nextcloud to collaborate upstream :)

I think we could do a lot better job here, and would be nice to start to work on that!

I have people asking on how to help, and I think here is a good place!

jancborchardt:
@nextcloud/documentation and @muppeth from Disroot: Get together please :)

esprit-libre:
and also [this](https://fr.wikibooks.org/wiki/Manuel_utilisateur_Nextcloud)

frju365:
any update ?

MorrisJobke:
cc @jospoortvliet

muppeth:
[...]

PR #1482, opened 2019-06-10
adding synch for Android

dacoex:
It should link to:
https://docs.nextcloud.com/android/android_app.html#sync-calendar-and-contacts

### DIFF
--- a/user_manual_de/pim/index.rst
+++ b/user_manual_de/pim/index.rst
@@ -11,6 +11,7 @@ können diese Apps aber mit einem Klick auf "Aktivieren" unter Apps -> Produktiv
    contacts
    calendar
    sync_ios
+   sync_android
    sync_osx
    sync_thunderbird
    troubleshooting


dacoex:
Sorry this is not ready for release.
Do we actually want a dedicated sub-page for android?
I guess linking to the doc for the android would be sufficient

skjnldsv:
@dacoex ah yes you're right!
Sorry I checked the user_manual but did not realized we were on the de section.
Yes, ideally everything should be translated :)

dacoex:
Oh, it seems I should have created a issue first.
But I could not find the repo for the documentation as there are so many for Nextcloud & I did not want to have my issue forgotten.

To explain:
* the English page needs to be updated to reflect token-based access & auto-discovery from the Nextcloud app (it automatically enters the server id).
* then, the English page needs to be translated.

Reference:
https://github.com/nextcloud/documentation/blob/master/user_manual/pim/sync_android.rst

skjnldsv:
Sure all fine :)
Can you do it? DO you need help?

dacoex:
> Can you do it? DO you need help?
Yes, but need some time. Maybe in July.

But is this translation always manual?
This is how new pages from the English master get forgotten.

Can this be done more systematic?
There's this issue as well:
https://github.com/nextcloud/documentation/issues/216

PR #1486, opened 2019-06-10
Update release_schedule.rst

tflidd:
This is the NC16 admin manual, it makes no sense to keep old releases from ownCloud. Even the version overview could be limited to supported versions. Or directly state from which versions updates can be run directly.


nickvergessen:
I would prefer to move 13-9 to that table too, as they are not supported anymore.

And it should be somewhere on our websites, so you don't have to check 3rd party websites to find out which version was released when, just like https://php.net/eol.php

tflidd:
I mostly refer to this page: https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule

It would be great to have a central location for this information and others can get this information from there. The servers already query some update servers for new version, not sure if they know if the version itself is EOL.

skjnldsv:
So what is the status here?

MorrisJobke:
I would keep the history in here. That page was added there to replace the more static part of the wiki page. I kept the history in there because sometimes you want to look up some date ranges and how it evolved. Thus having this as a nice page with the even not that useful information is something that is good to be preserved.

skjnldsv:
Closing then

PR #1505, opened 2019-06-20
Update Windows solution and alternate connection address.

kuipumu:
Based on multiple error trying to connect to multiple Nextcloud 16 servers thought WebDav, the correct address to get access from File Explorer or CLI is \\domain.comSSL\DavWWWRoot\remote.php\dav\files\username or if a subdomain is present (has in the current manual example)  \\domain.comSSL\DavWWWRoot\subdomain\remote.php\dav\files\username. Also add another solution to WebDav connection problems based on the "WebClient" service from Windows not beign started.

A part of the problem it's described in this next Nextcloud forum topic: https://help.nextcloud.com/t/webdav-problems-in-windows-7-8-10

### DIFF
--- a/user_manual/files/access_webdav.rst
+++ b/user_manual/files/access_webdav.rst
@@ -275,7 +275,7 @@ For example: ``net use Z: https://example.com/nextcloud/remote.php/dav/files/USE
 
 An alternative command syntax is::
 
-  net use Z: \\example.com@ssl\nextcloud\remote.php\dav /user:youruser
+  net use Z: \\example.com@ssl\DavWWWRoot\nextcloud\remote.php\dav\files\youruser /user:youruser
   yourpassword
 
 Mapping drives with Windows Explorer
@@ -288,7 +288,7 @@ To map a drive using the Microsoft Windows Explorer:
    the drop-down menu.
 3. Choose a local network drive to which you want to map Nextcloud.
 4. Specify the address to your Nextcloud instance, followed by
-   **/remote.php/dav/files/USERNAME/**.
+   **\DavWWWRoot\nextcloud\remote.php\dav\files\USERNAME**.
 
   For example::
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
Wow, how weird is that!
https://superuser.com/questions/494203/how-to-access-webdav-using-windows-explorer-on-windows-7

tflidd:
Interesting. Is this working properly, meaning is the connection stable an fast? Currently in the forums, we rather recommend to use 3rd party software and it would be cool to have a native implementation that works well. And which windows versions are working with that solution (if it works for one environment, we can ask the community to help with other versions)?

user_manual/files/access_webdav.rst
```diff
@@ -275,7 +275,7 @@ For example: ``net use Z: https://example.com/nextcloud/remote.php/dav/files/USE
 
 An alternative command syntax is::
 
-  net use Z: \\example.com@ssl\nextcloud\remote.php\dav /user:youruser
+  net use Z: \\example.com@ssl\DavWWWRoot\nextcloud\remote.php\dav\files\youruser /user:youruser
```
tflidd, line 278: That is if nextcloud is installed in a subfolder. For other reasons, this is not the preferred way, but it should be clear that the nextcloud folder is only required in this case.

tflidd (changes requested):
I just tried on Windows 10,
I had to adjust the registry, then start the Service manually, after that it was working!

We should clear up the thing, if Nextcloud is used in a subfolder or not and we are ready to go.

Before I did the modifictions, it tried to call the resource `/login` from the webserver which failed with a 405 error. Might help others where the service is not started and/or registry not yet fixed.

edit: Tested with NC 22.2.6 & NC 24 RC1

susnux:
This seems to be stale as there is no activity for over two years.
With Window 10 and 11 it seems to work fine just using `\\domain.com\subdir/remote.php/dav/files/admin`.

Please feel free to re-open if you want to work on it.

PR #1514, opened 2019-06-27
Bump minium pqsql version to 9.5

kesselb:
Ref: https://github.com/nextcloud/server/issues/15613

Nextcloud 16+ uses upsert for some queries on postgresql. Upsert is available since postgres 9.5.


skjnldsv:
@kesselbabot merge

PR #1515, opened 2019-06-27
Add example of client which doesn't support 2FA

pjrobertson:
More info, see https://github.com/nextcloud/twofactor_gateway/issues/249#event-2443339194

It was confusing/difficult for me to set up my CalDav client on macOS, as I was completely unaware of app passwords. This is an attempt to improve documentation to help other people down the line.

### DIFF
--- a/user_manual/user_2fa.rst
+++ b/user_manual/user_2fa.rst
@@ -78,6 +78,7 @@ Using client applications with two-factor authentication
 
 Once you have enabled 2FA, your clients will no longer be able to connect with
 just your password unless they also have support for two-factor authentication.
+One example of a client which don't support 2FA include the Calendar app on MacOS.
 To solve this, you should generate device specific passwords for them. See 
 :doc:`session_management` for more information on how to do this.
 


pjrobertson:
I agree with the thoughts here in principle: that we should list the clients that *do* support 2FA.

However the purpose of my commits wasn't to state which clients do/don't work, but to give people who were in my situation a solution. I was unaware that after setting up 2FA I wouldn't be able to use my other apps (although in hindsight it is obvious) so I hope we can change the documentation to add a reminder that not *all* clients accept 2FA. I guess my issue was more with the `twofactor_gateway` not providing information in this case, so this PR can be closed and the solution is to add a reminder about 'Using client applications with two-factor authentication' (which I have already created a PR for)

ChristophWurst:
Got it. Thanks! They idea was to one day only have app passwords for clients. That would of course take a lot of confusion from users. Unfortunately the login flow used by official clients is a non-standard one, hence 3rd party apps fail to connect with a little helpful authentication error.

PR #1517, opened 2019-07-01
Fix some code-block warnings

rullzer:


rullzer:
> Was it the reason of the docs failure?

No clue. I'm just slowly making progress there...

PR #1549, opened 2019-07-03
Add hash module to required list

bramceulemans:
Added the PHP hash module to the list of required modules. 

Nextcloud won't work without the PHP module for hashing.

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -105,6 +105,7 @@ Required:
 * PHP module curl
 * PHP module dom
 * PHP module GD
+* PHP module hash
 * PHP module iconv
 * PHP module JSON
 * PHP module libxml (Linux package libxml2 must be >=2.7.0)
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -105,7 +105,7 @@ Required:
 * PHP module curl
 * PHP module dom
 * PHP module GD
-* PHP module hash
+* PHP module hash (only on FreeBSD)
 * PHP module iconv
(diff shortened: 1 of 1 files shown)


kesselb:
>  As of PHP 5.1.2, the Hash extension is bundled and compiled into PHP by default.
> 
> It may be explicitly disabled by using the --disable-hash switch to configure. Earlier versions of PHP may incorporate the Hash extension by installing the » PECL module.
> 
> As of PHP 7.4.0, the Hash extension is a core PHP extension, so it is always enabled. 

https://www.php.net/manual/en/hash.installation.php

I'm voting against this change because the module is enabled by default.

@BramCeulemans https://github.com/nextcloud/documentation/pull/1549/checks?check_run_id=161376025 would you mind to sign off your commits?

skjnldsv:
Yeah, let's close then :)

Thanks for the help @BramCeulemans and @kesselb !

BramCeulemans:
It somehow is not bundled on FreeBSD 11, had to install it manually.

kesselb:
They disable all modules: https://svnweb.freebsd.org/ports/head/lang/php72/Makefile?view=markup#l26

I think it can't hurt. Mind to sign off your commits?

BramCeulemans:
Done. @kesselb

admin_manual/installation/source_installation.rst
```diff
@@ -105,6 +105,7 @@ Required:
 * PHP module curl
 * PHP module dom
 * PHP module GD
+* PHP module hash
```
kesselb:
```suggestion
* PHP module hash (only on freebsd)
```

PR #1559, opened 2019-07-16
move OCS API in one place

jospoortvliet:
So the OCS API was spread over several pages. I created one page (under client API) where, from what I can tell, the whole external OCS API can be found.

I did not re-arrange much more than this, so it still has room for improvement, but I wanted to keep the change minimal.

The existing situation is that you can find OCS pieces at these links:
https://docs.nextcloud.com/server/16/developer_manual/core/index.html  which links to:
https://docs.nextcloud.com/server/16/developer_manual/core/ocs-share-api.html
https://docs.nextcloud.com/server/16/developer_manual/core/ocs-sharee-api.html
(plus 2 more things which are, indeed, relevant for core development)

Then there is:
https://docs.nextcloud.com/server/16/developer_manual/client_apis/index.html
Which has a OCS API sub-page:
https://docs.nextcloud.com/server/16/developer_manual/client_apis/OCS/index.html

I re-arranged it that the above page is now an index which has 3 sub pages, one with the content of above, plus 2 with the pages that used to be under core, so all OCS is in one place. It might make sense to move it top-level, if you consider the other pages under [client APIs](https://docs.nextcloud.com/server/16/developer_manual/client_apis/index.html) to simply also be OCS - but I can't judge that.

I'd prefer to do a separate PR then, anyhow.

Signed-off-by: Jos Poortvliet <jospoortvliet@gmail.com>


MorrisJobke:
stable16 6b03e897

PR #1560, opened 2019-07-18
note on the redis socket file permissions

jospoortvliet:
This can take forever to debug ;-)


tflidd:
I still avoid having 777 permissions. So I put the webserver-user in the redis-group and use 770 socketpermissions.

blizzz:
> I still avoid having 777 permissions. So I put the webserver-user in the redis-group and use 770 socketpermissions.

Agreed, I am doing the same.

jospoortvliet:
Perfectly valid point, I modified the instructions to advise to put the web server user in the redis group instead of changing the permissions.

jospoortvliet:
Thanks @MorrisJobke !!!

PR #1564, opened 2019-07-19
Add webdav comments api

jospoortvliet:
based on: https://github.com/nextcloud/server/issues/7315#issuecomment-347468998

### DIFF
--- a/developer_manual/client_apis/WebDAV/comments.rst
+++ b/developer_manual/client_apis/WebDAV/comments.rst
@@ -1,0 +1,151 @@
+=========
+Comments
+=========
+
+Comments have a PHP API within OCP and also via WebDAV. Here is a very basic description of the WebDAV API.
+
+The Comments resource has an endpoint::
+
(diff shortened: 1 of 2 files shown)


developer_manual/client_apis/WebDAV/comments.rst
```diff
@@ -0,0 +1,151 @@
...

+* DELETE to delete it

+* PROPFIND to list all properties

+

+

+Here is a full list of current properties::
```
blizzz: ah, i misunderstood what you meant with leaving away things. In fact, the whole namespace is of importance. Like in https://docs.nextcloud.com/server/16/developer_manual/client_apis/WebDAV/search.html 

Also the `self::PROPERTY_NAME_FOO`` needs to be resolved as it does not help here. self is a reference to a php class of which we do not know here and the other part is the name of a constant.

MorrisJobke:
Fixed by #1968

Issue #1275, opened 2019-02-23
Add information about E2E encryption


tflidd:
The client-side encryption is not really documented, there is nothing about the current 2.5 desktop client. Without documentation it is hard to answer them without trying each option yourself (still it would be hard to say if this is default behavior, some mis-understanding function or a bug). In the forum, there are already questions piling up:
https://help.nextcloud.com/tags/e2e

MorrisJobke:
cc @tobiasKaminsky @schiessle @rullzer 

There is currently the technical part in https://github.com/nextcloud/end_to_end_encryption_rfc/blob/master/RFC.md

tobiasKaminsky:
On android side we do have some intro while setting up E2E, but no dedicated help/info.
As with other parts I would rather go with a better UI/UX than having a real documentation.
After all E2E should work seamless without knowing technical parts.
If people cannot set it up properly, we failed at UX.

But in general, please note that there are currently know bugs, especially in Desktop.

joshtrichards:
The README for the E2EE repository was (significantly) expanded recently. Even added basic troubleshooting steps, etc. :-)

https://github.com/nextcloud/end_to_end_encryption/blob/master/README.md

I suggest directing people there for now.

skjnldsv:
E2EE is now fully documented at https://docs.nextcloud.com/server/latest/user_manual/en/files/using_e2ee.html — covers setup, encrypting folders, adding devices, and mnemonic management for desktop, Android, iOS, and web.

PR #1571, opened 2019-07-29
added the encryption details page which decribes the server-side encr…

yahesh:
…yption implementation of the default encryption module

As was advised in the nextcloud/server issue [#16419](https://github.com/nextcloud/server/issues/16419#issuecomment-515935285) we provide the description of the server-side encryption as a ReStructuredText document. As this document will likely have another target audience than the existing "Encryption configuration" page we created a separate "Encryption details" page.


jknockaert:
I had a look at the text. It's definitely a good idea to document this topic, so thank you for your effort. My knowledge about the encryption mechanism in nextcloud is limited to the encryption wrapper, and as far as that mechanism is concerned the text seems correct, with just one omission that you may want to adress.
Very early versions of owncloud had a different encryption schedule. If I remember correctly there was no header in the first block of the encrypted file. The header only came upon the first revision of the encryption format. All later versions of owncloud/nextcloud still support that initial (and all later) versions of the encryption format, as encrypted files never get updated when a new version of the format is introduced. After that revision there have been later updates to the encryption format, but I don't remember exactly what they were about. There has been a vulnerability adressed, and also file versioning required some tweaking in the encryption mechanism, but I cannot recall the fine details.
Perhaps you should just mention in your document that you are discussing a specific version of the encrypted file format, and that older versions are still around and supported by nextcloud.

yahesh:
@jknockaert Thanks for the feedback. A added a note to the introduction that describes that previous versions of Nextcloud implemented slightly different encryption schemes and that these files may till be around.

yahesh:
@rullzer @MorrisJobke @schiessle @jknockaert Hi, this PR is now stalled for a over a month. Is there something we can help with to get this PR approved?

J0WI:
cc @nextcloud/encryption

J0WI:
Maybe you could also add a reference to E2E to fix #1275.

J0WI:
Oh and maybe a note that app data (calendar, contacts, chats etc.) are stored in the database and **not** encrypted.

yahesh:
> Maybe you could also add a reference to E2E to fix #1275.

The server-side encryption is a totally different concept compared to the client-side E2E encryption. So "referencing" it just to fix that issue doesn't seem like a viable solution.

yahesh:
> Oh and maybe a note that app data (calendar, contacts, chats etc.) are stored in the database and **not** encrypted.

IMHO the document already goes into great detail **what** gets encrypted by the server-side encryption. From the details it should become obvious that data within the database do **not** get encrypted. :)

yahesh:
@rullzer @MorrisJobke @schiessle @jknockaert Another month has gone by.

yahesh:
@rullzer @MorrisJobke @schiessle @jknockaert Hi there, anonther month has passed. Did you find the time to review this pull request? 🙂

skjnldsv:
@yahesh unfortunately Morris is unavailable and everyone is completely booked!

Maybe @J0WI and @jknockaert can help and review? :)

admin_manual/configuration_files/encryption_details.rst
```diff
@@ -0,0 +1,356 @@
...
+
+Key type: public sharing key
+----------------------------
+
+The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that can be generated by the server administrator. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
+
+Key type: recovery key
+----------------------
+
+The recovery key is used to provide a restore mechanism in cases where the user key encryption is enabled and users have lost their passwords. The recovery key is protected by a recovery password that the server administrator should store securely. The advantage of the recovery key is that files can be recovered but has the disadvantage that the server administrator is able to decrypt user files without knowing any user password.
```
J0WI: The recovery key is only used if user keys are activated and no master key is used. Both, the admin and the user must enable the recovery key:
https://docs.nextcloud.com/server/stable/admin_manual/configuration_files/encryption_configuration.html#enabling-users-file-recovery-keys

yahesh:
@J0WI I introduced your recommended changes.

kesselb:
Thanks @yahesh :+1:

PR #1575, opened 2019-08-03
Link to help on oc_activity migration failure

gagbo:
I tried to do the migration and the solution in the linked issue is exactly what I missed to be able to finish the process. Since I am not the only one with the issue apparently, it would be nice to have it in the documentation just in case.


MorrisJobke:
@nickvergessen Any comment on this documentation note?

nickvergessen:
I guess this is something that happened, when the activity app was installed on OC7 or before and disabled on an update to 8 or 8.1, before stuff was done with migrations.

I'm not sure we should add it in general like this, as there could be much more other issues. But no strong feelings.

MorrisJobke:
Let's close it then and if we get more reports we can get it in.

PR #1580, opened 2019-08-11
Update to rsync command for backups

Wehzie:
Firstly, I removed -A because the man page prohibits the use of -A and -a together. The need for -a for archiving is clear. The need for ACL in this command provided by the documentation is not clear. In this sense the need for -a wins over -A.

Secondly, I removed -x as users may have mounted drives or semantic links to other drives. The explicit prohibition of rsync accessing those drives in this documentation page isn't clear.

Lastly, I added the requirement of sudo as the Nextcloud folder in /var/www/nextcloud is owned by the Apache www-data group. Without sudo a privilege related error occurs.

### DIFF
--- a/admin_manual/maintenance/backup.rst
+++ b/admin_manual/maintenance/backup.rst
@@ -33,7 +33,7 @@ Backup folders
 Simply copy your config, data and theme folders (or even your whole Nextcloud install and data folder) to a place outside of
 your Nextcloud environment. You could use this command::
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
Please have a look at  https://github.com/nextcloud/documentation/pull/1377

3 months ago we change it on purpose :)

Wehzie:
I see, thank you for pointing this out.
Is there anything to be said or done regarding the other two changes?

skjnldsv:
No idea really, i'm really not an expert on the subject :)
FOr the additional info regarding the user, it's always a good info to add

PR #1581, opened 2019-08-16
Bob kircher patch 2

BobKircher:
Updated Doku with respect to SSL limitations starting with iOS 12
If the reference to letsencrypt is not OK, than delete it.

### DIFF
--- a/user_manual/pim/sync_ios.rst
+++ b/user_manual/pim/sync_ios.rst
@@ -16,7 +16,7 @@ Calendar
 
 Your calendar will now be visible in the Calendar application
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
@marinofaggiana please have a look :)

BobKircher:
I get a message "DCO Action required"
Should I close the patch, add a "Signed-off-by: ... " and upload a new pull request ?

juliushaertl:
@BobKircher Yes that would be highly appreciated. For more details see the description at https://probot.github.io/apps/dco/

PR #1582, opened 2019-08-17
Update tutorial to match the repo

### LINKED ISSUE #1501: App Tutorial Doc references different table in NoteMapper.php
Following the tutorial on https://docs.nextcloud.com/server/15/developer_manual/app/tutorial.html#database it states for NoteMapper.php:
>     public function __construct(IDbConnection $db) {
>         parent::__construct($db, 'ownnotes_notes', Note::class);
>     }

results for me in "notes could not be loaded"

looks like that table never has been defined within the tutorial documentation?
the only table that has been instantiated within the migration-steps was "notes_meta"..

Getting a bit desperate while trying to follow this tutorial..

juliushaertl:
Requires https://github.com/nextcloud/app-tutorial/pull/5
fix #1501

- [x] Adjust naming to match the app repo
- [x] Simplify frontend description to point to the source code, so we don't need to maintain this in two places
- [x] Add links to the actual files in the git repo
- [x] Add some more details on webpack/vue maybe with helpful links

### DIFF
--- a/developer_manual/app/tutorial.rst
+++ b/developer_manual/app/tutorial.rst
@@ -1061,7 +1061,7 @@ Since the **NoteApiController** is basically identical to the **NoteController**
 Building the frontend
 ---------------------
 
-To create a modern webapp you need to write :doc:`JavaScript<view/js>`. You can use any JavaScript framework, but this tutorial focusses on a simple frontend using Vue.js.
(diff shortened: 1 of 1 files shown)


skjnldsv:
Should we add some testing too?
e2e cypress should be simple enough?

juliushaertl:
Feel free to do that to the notestutorial app, but I would prefer to keep the documentation pointing to the actual source so we don't need to maintain it in two locations.

skjnldsv:
Agreed!

PR #1588, opened 2019-08-21
Docu: Change background job cycle from 5 min to 15 min

### LINKED ISSUE #1587: Documentation: backgroud jobs 15 min vs. 5 min
The background jobs configuration page [1] of my installation offers three options [german]: 
"AJAX
Eine Aufgabe bei jedem Laden einer Seite ausführen

Webcron
cron.php ist als Webcron-Dienst registriert, der die cron.php alle 15 Minuten per HTTP aufruft.

Cron
Nutzen Sie den System-Cron-Service um cron.php alle 15 Minuten aufzurufen. Die cron.php muss durch den Systemnutzer "wwwrun" ausgeführt werden. "

Both cron options mention a 15 minute cycle.
The corresponding documentation [2] mentions in the examples always a 5 minute cycle.
(e.g. "To run a cron job on a *nix system, every 5 minutes, under the default Web server user (often, www-data or wwwrun), you must set up the following cron job to call the cron.php script:")

I suggest to change the documentation to 15 minutes (6 hits on that page).

[1] https://<myhomepage>/index.php/settings/admin
[2] https://docs.nextcloud.com/server/16/admin_manual/configuration_server/background_jobs_configuration.html

koppd:
The NC configuration page /index.php/settings/admin recommends a 15 min cycle for background jobs.
Change the documentation accordingly (from 5 min to 15 min).

fix  #1587

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -64,17 +64,17 @@ Using the operating system cron feature is the preferred method for executing
 regular tasks.  This method enables the execution of scheduled jobs without the
 inherent limitations the Web server might have.
 
-To run a cron job on a \*nix system, every 5 minutes, under the default Web
+To run a cron job on a \*nix system, every 15 minutes, under the default Web
 server user (often, ``www-data`` or ``wwwrun``), you must set up the following
 cron job to call the **cron.php** script::
 
   # crontab -u www-data -e
-  */5  *  *  *  * php -f /var/www/nextcloud/cron.php
+  */15  *  *  *  * php -f /var/www/nextcloud/cron.php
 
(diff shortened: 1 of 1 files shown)


juliushaertl:
Thanks for your suggestion. Running the cronjob every 5 minutes is just fine and will make sure that any background job is executed quicker. We probably should rather adjust the webcron message in the server repo although it is not a big issue I guess.

skjnldsv:
@juliushaertl should we do this then?

nickvergessen:
5 Minutes is actually what it should be, so we should fix the admin settings.

skjnldsv:
Let's still fix the documentation :)

skjnldsv:
Ah, I misread this one!
I thought it was 15 and we were changing it to 5! Sorry :)

PR #1598, opened 2019-08-24
Gitignore against stable16

kotp:


kotp:
This is what #1600 should have been.

Issue #1584, opened 2019-08-20
nextcloudcron.service file fails produces error on reboot.


Steve8291:
After installing `nextcloudcron.service` and `nextcloudcron.timer` and enabling them in systemd they create an error on reboot. It appears that `nextcloudcron.service` is starting up before `mysql.service`  
I've added the line `After=mysql.service` to the service file. Would it be prudent to make this change in the documentation?
Referencing this file: [https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/background_jobs_configuration.rst](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/background_jobs_configuration.rst)
```
Description=Nextcloud cron.php job
After=mysql.service

[Service]
User=www-data
ExecStart=/usr/bin/php -f /var/www/nextcloud/cron.php

[Install]
WantedBy=basic.target
```
Also noticed what appears to be a typo:  
`When the option --now is used with enable, the resp. unit will also be started.`  
Is "resp." supposed to read `.service` ?

skjnldsv:
@Steve8291 thanks for the suggestion :)
If you think it make sense, please provide a patch for the documentation! :rocket: 

I think it does make sense to wait for the dependencies services to be ready

kesselb:
> When the option --now is used with enable, the resp. unit will also be started.
Is "resp." supposed to read .service

Unit is  the official naming: https://www.freedesktop.org/software/systemd/man/systemd.unit.html

Steve8291:
Ah. "resp." is the abbreviation for "respective". I get it now. Sorry. I guess I had never seen it abbreviated before.

kesselb:
I think this has been fixed by #1603

PR #1603, opened 2019-08-24
Update background_jobs_configuration.rst

Steve8291:
Adjust `nextcloudcron.service` to wait for `mysql.service` dependency before starting. This prevents errors during a reboot.  
This references: https://github.com/nextcloud/documentation/issues/1584#issuecomment-524545159

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -96,6 +96,7 @@ This approach requires two files: **nextcloudcron.service** and **nextcloudcron.
 
   [Unit]
   Description=Nextcloud cron.php job
+  After=mysql.service
   
   [Service]
   User=www-data
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -96,17 +96,13 @@ This approach requires two files: **nextcloudcron.service** and **nextcloudcron.
 
   [Unit]
   Description=Nextcloud cron.php job
-  After=mysql.service
   
   [Service]
   User=www-data
   ExecStart=/usr/bin/php -f /var/www/nextcloud/cron.php            
-  
-  [Install]
-  WantedBy=basic.target
 
 Replace the user ``www-data`` with the user of your http server and ``/var/www/nextcloud/cron.php`` with the location of **cron.php** in your nextcloud directory.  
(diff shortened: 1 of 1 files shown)


Steve8291:
On Ubuntu both mariadb and MySQL are covered by the mysql.service so there is no need to add anything additional. I don’t know about pgsql but I’ll try to find out. Either way I’ll adjust with a comment to change the code based on your database.

skjnldsv:
Thanks a lot @Steve8291 :)

Steve8291:
@skjnldsv I've made the changes. I don't need to do another pull request do I?

kesselb:
Thanks :+1: 

I'm not an systemd expert but adding a dependency to mysql or pqsql is wrong way to fix it. 

> The important parts in the timer-unit are OnBootSec and OnUnitActiveSec. OnBootSec will start the timer 5 minutes after boot, otherwise you would have to start it manually after every boot. OnUnitActiveSec will set a 5 minute timer after the service-unit was last activated.

Why is this not working? I think the Install section within the service definition is the problem here. It says the nextcloud cron is requested to be started during boot. 

> For each .timer file, a matching .service file exists (e.g. foo.timer and foo.service). The .timer file activates and controls the .service file. The .service does not require an [Install] section as it is the timer units that are enabled. If necessary, it is possible to control a differently-named unit using the Unit= option in the timer's [Timer] section. 

From https://wiki.archlinux.org/index.php/Systemd/Timers. Removing the Install section from the service definition should do the trick. On very slow instances the OnBootSec needs a higher value but 5min should work fine for most systems.

Steve8291:
Thanks @kesselb that sounds like a better way to fix the errors.  
I just tested this on my server and it works. No errors after a reboot and the timer is doing its job.  
nextcloudcron.service now looks like this:  
```
[Unit]
Description=Nextcloud cron.php job

[Service]
User=www-data
ExecStart=/usr/bin/php -f /var/www/nextcloud/cron.php
```
Might need to add a note in the documentation that you don't run `systemctl enable nextcloudcron.service` anymore but only enable the timer.

skjnldsv:
> Might need to add a note in the documentation

Could you add this? :) :bowing_woman:

Steve8291:
Should I create a new pull request or just modify this one?

kesselb:
Modify this one. Could you add another note like "heads up the service file does not need the Install section. Please check your setup because we recommended it in earlier versions". I guess the wording needs some tweaking.

kesselb:
Thanks @Steve8291 :rocket:

PR #1609, opened 2019-09-03
Enable browser cache for thumbnails

stefaang:
As suggested in https://github.com/nextcloud/docker/issues/848

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -179,6 +179,12 @@ webroot of your nginx installation. In this example it is
 
       location ~ \.(?:png|html|ttf|ico|jpg|jpeg|bcmap)$ {
           try_files $uri /index.php$request_uri;
+          
+          # Optional: enable browser caching for thumbnails
+          # Makes page loads really snappy for low power servers
+          # but potentially slightly outdated
+          add_header Cache-Control "public, max-age=86400";
+          
           # Optional: Don't log access to other assets
           access_log off;
       }


stefaang:
I've been reading into cache control a bit.. I guess private is more appropriate here so prevent privacy leaks.

rullzer:
Which thumbnails do you mean here?
Because there should already be caching headers set on the previews you get back. If not then that is a bug.

stefaang:
With browser cache - page reload = 1.3s
nc-with-browser-cache
Without browser cache - page reload = 9s
nc-no-browser-cache

As said, only applies to instances running on low-end hardware.

stefaang:
Ideally, the file browser uses browser cached icons when available on load, then updates them once any non-304 GETs come in.

admin_manual/installation/nginx.rst
```diff
@@ -179,6 +179,12 @@ webroot of your nginx installation. In this example it is
 
       location ~ \.(?:png|html|ttf|ico|jpg|jpeg|bcmap)$ {
           try_files $uri /index.php$request_uri;
+          
+          # Optional: enable browser caching for thumbnails
+          # Makes page loads really snappy for low power servers
+          # but potentially slightly outdated
+          add_header Cache-Control "public, max-age=86400";
```
skjnldsv: If it's optional, it should commented :)

rullzer:
> Ideally, the file browser uses browser cached icons when available on load, then updates them once any non-304 GETs come in.

Again, which request. Because the 304 on the previews should only happen after a day. Else something else if fishy.

So which requests (url please) are you trying to speed up here.
Because on all 4 setups I have access to icons are properly cached.

stefaang:
> So which requests (url please) are you trying to speed up here.

Ah... I've changed my setup from lighttpd to nginx and it seems to resolve the problems.
No more need for this MR. 

FYI, I'm working on an Entware (openwrt based) setup and there was only a guide for a lighttpd based installation that worked more or less out of the box. Unfortunately performance was lacking.

Thanks anyway for your time.

PR #1640, opened 2019-10-01
Add note about known to work hardware tokens

alex-nitrokey:


skjnldsv:
Cc @christophWurst

jans23:
@rullzer What do you think?

alex-nitrokey:
> If we want to have this it should be at https://github.com/nextcloud/twofactor_u2f#u2f-second-factor-provider-for-nextcloud, similar to the apps at https://github.com/nextcloud/twofactor_totp

Do you mean adding this information to said repos additionally or instead of adding it here?

tflidd:
FIDO/U2F is not enough of a standard that we can suppose in general that it works?

For TOTP based, I'd like to see more so we do not just promote one product (then perhaps point out devices with open design).

ChristophWurst:
> Do you mean adding this information to said repos additionally or instead of adding it here?

We can add it as you proposed, but it's not 100% fitting IMO as the manual does not cover u2f otherwise. Hence I suggested adding it to the twofactor_u2f repo readme. We can also do both if you want :)

jans23:
Looking at our partnership to promote the usage of 2FA, adding both would be great.

jospoortvliet:
@jans23 - if you modify the PR to add it to both, if that's OK with @ChristophWurst  - we can merge ;-)

alex-nitrokey:
[OTP README PR](https://github.com/nextcloud/twofactor_totp/pull/720)
[U2F README PR](https://github.com/nextcloud/twofactor_u2f/pull/569)

Is this what you were thinking of?

PR #1646, opened 2019-10-02
Obtain CSRF token

usselite:
Explanation for external clients on how to retrieve the CSRF token.

### DIFF
--- a/developer_manual/client_apis/LoginFlow/index.rst
+++ b/developer_manual/client_apis/LoginFlow/index.rst
@@ -159,3 +159,41 @@ This will return a 404 until authentication is done. Once a 200 is returned it i
 
 Use the server and the provided credentials to connect.
 Note that the 200 will only be returned once.
+
+CSRF Check
+-------------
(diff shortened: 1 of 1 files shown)


rullzer:
Client should never fetch the CSRF token.
What are you trying to do that you need this in a client?

usselite:
Creating guest accounts requires a CSRF check. Within the Outlook-addin I am trying to facilitate this option. See https://github.com/nextcloud/guests/blob/57062708251b418d09e03ee7bac4794c3c013f5a/lib/Controller/UsersController.php#L70

rullzer:
> Creating guest accounts requires a CSRF check. Within the Outlook-addin I am trying to facilitate this option. See https://github.com/nextcloud/guests/blob/57062708251b418d09e03ee7bac4794c3c013f5a/lib/Controller/UsersController.php#L70

So that is actually invalid and not supported. Because without a session your CSRF token won't be valid anyways.

The real fix here is providing a real API endpoint in the guest app that clients can call ;)
I'll see what we can do there.

usselite:
I actually do have a session in this case. But granted, the addition

PR #1674, opened 2019-10-24
Harden NGINX config

rullzer:

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -119,7 +119,7 @@ webroot of your nginx installation. In this example it is
       #pagespeed off;
 
       location / {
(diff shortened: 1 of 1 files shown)


jknockaert:
I think a feature of nginx was overlooked in this PR. By adding `try_files` the variable `$fastcgi_path_info` is reset:
https://trac.nginx.org/nginx/ticket/321
So the fastcgi parameter `PATH_INFO` will get a different value because of adding `try_files` to the config.
This can be solved by adding `set $path_info $fastcgi_path_info;` before `try_files`, and then assigning `$path_info` to `PATH_INFO`.

rullzer:
@jknockaert could you submit a PR?

jknockaert:
@rullzer #1679

PR #1679, opened 2019-10-24
fix PATH_INFO

jknockaert:
`try_files` was added in #1674, but that resets `$fastcgi_path_info`. This PR makes sure PATH_INFO gets the original value.


jknockaert:
@rullzer Would there be any need for a backport to the blog post?
https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/

rullzer:
Yes will take care tomorrow
@jospoortvliet

Starbix:
I think this should also be added to the blog entry.

rullzer:
Yeslike I said it will be added. However, having said that. For me my config works just fine without it as well.

jknockaert:
@jknockaert I've been wondering if and where the php paramater `PATH_INFO` is used in nextcloud. But provided it is there in the nextcloud config file, we should make sure it doesn't change.

PR #1610, opened 2019-09-04
Remove APC config line from Redis example

szepeviktor:

PR #1687, opened 2019-11-03
Corrected the memcache.local

IPv777:
if we want to use Redis, why specify APCu here ?


skjnldsv:
This is done on purpose as far as I know :)
You're not the first one to open a pull request about it ;)
https://github.com/nextcloud/documentation/pull/1610

Thanks nonetheless!

skjnldsv:
https://docs.nextcloud.com/server/15/admin_manual/configuration_server/caching_configuration.html#recommendations-based-on-type-of-deployment

PR #1594, opened 2019-08-23
update user manual PIM iOS

BobKircher:
iOS documentation was updated with respect to SSL requirements starting with iOS 12.

PR #1693, opened 2019-11-12
Update synchronizing with MacOS

heaterC:
If service discovery on server side is configured correctly (which it should be anyhow), then "manual" instead of "advanced" is the easiest way to add account to MacOS.


BernieO:
See also #1594 , which is apparently the according pull request for iOS devices.

skjnldsv:
Ah right
@heaterC can you see if #1594 fit your needs? :)

heaterC:
@skjnldsv Well not really, since one is iOS and the other MacOS. They still have different dialogues and slightly different procedures for account creation. For that reason, there are two different documentation pages, one for each, in the first place.

skjnldsv:
@BernieO we should merge both then?

BernieO:
> @BernieO we should merge both then?

@skjnldsv Yes, imho both should be merged:
- this one for Mac OS X
-  #1594 for iOS

Additionally maybe a new section could be created due to the more restrictive certificate handling in Mac OS X 10.15 (Catalina) and iOS 13 (see here: https://support.apple.com/en-us/HT210176 and here: https://help.nextcloud.com/t/ios-13-syncs-calendar-and-contacts-but-not-reminders/62158/12). I assume that most selfsigned certificates do not meet the new conditions for a trustworthy certificate. But that is another story. Maybe I‘ll do that when I have some spare time...

PR #1703, opened 2019-11-14
Remove SAMEORIGIN header from nginx config doc

ogasser:
Removes the SAMEORIGIN header from the nginx configuration documentation as this is handled by PHP directly.

Fixes #1701

PR #1707, opened 2019-11-20
Removed `X-Frame-Options: SAMEORIGIN` header

dvdgsng:
Some headers don't need to be set in nginx anymore, since they are hardcoded into Nextcloud. This causes problems with the header `X-Frame-Options: SAMEORIGIN`, which should not be set twice.

See:
* https://docs.nextcloud.com/server/17/admin_manual/installation/harden_server.html#serve-security-related-headers-by-the-web-server
* https://docs.nextcloud.com/server/12/admin_manual/release_notes.html#updates-to-nginx-configuration
* https://github.com/nextcloud/server/issues/4764
* https://blog.qualys.com/securitylabs/2015/10/20/clickjacking-a-common-implementation-mistake-that-can-put-your-websites-in-danger

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -69,7 +69,6 @@ webroot of your nginx installation. In this example it is
       add_header Referrer-Policy "no-referrer" always;
       add_header X-Content-Type-Options "nosniff" always;
       add_header X-Download-Options "noopen" always;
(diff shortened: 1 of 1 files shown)


skjnldsv:
Please see https://github.com/nextcloud/documentation/pull/1703

dvdgsng:
missed that, thanks.

PR #1714, opened 2019-11-22
add Content-Security-Policy header

morgankevinj:
Add "add_header Content-Security-Policy "default-src 'self'" always;" and remove redundant headers

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -69,10 +69,9 @@ webroot of your nginx installation. In this example it is
       add_header Referrer-Policy "no-referrer" always;
       add_header X-Content-Type-Options "nosniff" always;
       add_header X-Download-Options "noopen" always;
-      add_header X-Frame-Options "SAMEORIGIN" always;
+      add_header Content-Security-Policy "default-src 'self'" always;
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hi,

Our csp settings are managed within nextcloud directly.


> remove redundant headers

What do you mean by that?

morgankevinj:
Setting frame-ancestors to 'none' should be roughly equivalent to X-Frame-Options: DENY.
script-src limits scripts so it can replace X-XSS-Protection.
When frame-ancestors and script-src are not decaled the policy fallsback on default-src.
according to https://content-security-policy.com/#source_list

skjnldsv:
cc @rullzer

rullzer:
Nope. We set the csp from nextcloud. Because we need to also set the nonce etc.

kesselb:
Thanks @morgankevinj :+1: 

Unfortunately IE11 is still supported https://github.com/nextcloud/browserslist-config/blob/master/browserlist.config.js

and does not support the frame-ancestors directive https://caniuse.com/#feat=mdn-http_headers_csp_content-security-policy_frame-ancestors

skjnldsv:
Closing then :)

PR #1715, opened 2019-11-23
Fix MySQLCMD query

KenLewis:


kesselb:
Thanks @KenLewis :+1: 

What is the purpose of this change?

KenLewis:
The query as it was written wouldn’t work as intended. Notice in the query just before this one that it uses “LIKE ‘nextcloud%’;” to find all the tables that start with nextcloud. The query I changed supposed to do create a list of ALTER queries for all the nextcloud tables but fails to do so since it’s using = instead of LIKE. With the change I submitted, the list of ALTER queries is created as intended. 

(Sorry for the formatting of this comment as I’m on mobile)

kesselb:
OK. The actual query works for me. It will create the alter queries for all tables in the nextcloud database. Your query will create the alter queries for every table in a database starting with nextcloud (eg. nextcloud_prod, nextcloud_test). Why would we do that?

juliushaertl:
I'd say we should keep the documentation generic as it is here. You should always be careful when running commands and you need to adjust it to your database naming schema anyway.

PR #1725, opened 2019-11-28
add a section about GDPR relevant information

schiessle:
start with documentation about cookies, more can be added later. E.g. it might make sense to also have a few sentence about our GDPR tools

@NinaNC what do you think?
@rullzer can you double-check the technical details?

### DIFF
--- a/admin_manual/gdpr/cookies.rst
+++ b/admin_manual/gdpr/cookies.rst
@@ -1,0 +1,32 @@
+=======
+Cookies
+=======
+
+.. sectionauthor:: Björn Schießle <bjoern@nextcloud.com>
+.. _cookies:
+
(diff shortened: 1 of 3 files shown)


admin_manual/gdpr/cookies.rst
```diff
@@ -0,0 +1,30 @@
...
+.. _`only data which contain personal data are relevant`: https://gdpr-info.eu/recitals/no-26/
+
+
+Cookies stored by Nextcloud
+===========================
+
+====================  ====================================  ================
+ Cookie                  Data Stored                             Lifetime
+====================  ====================================  ================
+ Session cookie        - session ID                          24 minutes
```
rullzer: So we store more cookies. But it is roughtly this list. Let me add some more info later. Also with examples what is stored and why.

juliushaertl:
:ping_pong:

PR #1726, opened 2019-11-28
Remove Same-Origin header from Nginx configuration (duplicate with PHP)

louis-berruyer:
The Same-Origin header is handled in PHP and keeping it in Nginx configuration will result in a wrong header.


kesselb:
Thanks +1

For Nextcloud 17 the X-Frame-Options is required: nextcloud/server#16179. The release notes are outdated.

PR #1743, opened 2019-12-04
Remove old theme and unused dependencies

skjnldsv:
Capture d’écran_2019-12-04_14-50-50


developer_manual/Makefile
```diff
@@ -47,7 +47,7 @@ html:   html-com
 html-all: html-release html-com
 
 html-release:
-	$(SPHINXBUILD) -b html -D html_theme='nextcloud_release' $(ALLSPHINXOPTS) $(BUILDDIR)/html/release
+	$(SPHINXBUILD) -b html -D html_theme='sphinx_rtd_theme' $(ALLSPHINXOPTS) $(BUILDDIR)/html/release
```
nickvergessen: this is now the same as `html-com` just a different path at the end, so I guess we can delete one of them?

skjnldsv: Yes, but I did not wanted to try my luck ;)
ALso it would means fixing the release script and/or the publish script

nickvergessen: Ah, nevermind then

PR #1748, opened 2019-12-15
admin_manual/installation/nginx: Added tip for allowing uploads while blocking dot files

Zocker1999NET:
Adds hint about solution for issue nextcloud/server#8802.


kesselb:
Thanks @Zocker1999NET :rocket:

PR #1755, opened 2019-12-29
document option --no-interaction for occ db:convert-type

BernieO:
See nextcloud/server#18576


BernieO:
I suggest to backport this to stable 17 and stable 16.

PR #1757, opened 2020-01-02
Cleanup installation manual

juliushaertl:
- Move example installation guides for Ubuntu/CentOS out of the general installation
- Update guide for CentOS 8 (recommend remi packages)
- Create mysql database with utf8mb4 by default
- Try clarifying that snap/vm are community projects


skjnldsv:
Awesome!!

wiswedel:
Awesome work, @juliushaertl. Thanks a lot for that.
However, you're recommending PHP 7.4 with RHEL but 7.2 with Ubuntu 18.04 LTS.

The default Ubuntu apt repo only provides 7.2, but I think we should also be open for 3rd party repos here. PHP 7.2 might be acceptable for now (not sure if for 19 as well), but for example Apache ships with 2.4.29 (Oct 2017) in the Ubuntu repos, Mariadb with 10.1.43 whichi will have eol in october 2020 (current stable major is 10.4.x).

I suggest we open up for
- ppa:ondrej/php
- ppa:ondrej/apache2
- a repo of choice from https://downloads.mariadb.org/mariadb/repositories

juliushaertl:
> However, you're recommending PHP 7.4 with RHEL but 7.2 with Ubuntu 18.04 LTS.

No, the first instructions are with the default provided packages for 7.2, but we also state that there is a 3rdparty repo for 7.4. I'd say we should do the same for Ubuntu. Keep the main instructions for the current release shipped with the distro but maybe add a section about repositories with more recent versions.

> The default Ubuntu apt repo only provides 7.2, but I think we should also be open for 3rd party repos here. PHP 7.2 might be acceptable for now (not sure if for 19 as well), but for example Apache ships with 2.4.29 (Oct 2017) in the Ubuntu repos, Mariadb with 10.1.43 which will have eol in october 2020 (current stable major is 10.4.x).

Should be fine until we update the docs for Ubuntu 20.04

PR #1772, opened 2020-01-15
Windows 7 is EOL

rullzer:


tflidd:
It's just EOL for consumers, there are enterprise users (of Windows) that still receive security updates. Such links can be interesting because problems are often similar between versions in Windows and there might be links more information if you have similar problems (even with other versions).

MorrisJobke:
Let's keep it for now.

PR #1777, opened 2020-01-18
Remove server php docs build

juliushaertl:


juliushaertl:
CI also fails on master

rullzer:
Just forcing the old svgexport now :P

PR #1779, opened 2020-01-18
Document (new) JS APIs

ChristophWurst:
Ref https://github.com/nextcloud/server/issues/15932
Ref https://github.com/nextcloud/server/issues/18677

### DIFF
--- a/developer_manual/app/javascript-apis.rst
+++ b/developer_manual/app/javascript-apis.rst
@@ -1,0 +1,159 @@
+===============
+JavaScript APIs
+===============
+
+Nextcloud apps can use existing JavaScript APIs to ease the development of front-end components and simple scripts.
+
+The APIs used to be provided via global variables, available on most pages of Nextcloud. To smoothen the development experience with modern development tools, this method is in the process of being deprecated and removed. Existing APIs are being migrated to npm packages and new APIs will only be available that way. The bottom of the page covers the basics of the global variable method, if you want to develop an app for old versions of Nextcloud.
+
+npm packages
+------------
+
+`@nextcloud npm packages <https://www.npmjs.com/org/nextcloud>`_ provide the current front-end APIs for Nextcloud apps.
+
+
+Usage
+^^^^^
+
+The idea is that apps install these packages via `npm` and bundle the code with tools like `Babel <https://babeljs.io/>`_, `Webpack <https://webpack.js.org/>`_ or `Parcel <https://parceljs.org/>`_. This ensures that an app runs the exact same code independent of the Nextcloud version and also reduces the chances of running into conflicts with other apps.
+
+For more details on the design considerations see `the discussion on Github <https://github.com/nextcloud/server/issues/15932>`_.
+
+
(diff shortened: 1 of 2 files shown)


developer_manual/app/javascript-apis.rst
```diff
@@ -0,0 +1,157 @@
...
+
+This package provides helpers to generate URLs, e.g. to access assets and REST APIs of your app or the Nextcloud server.
+`Read the full documentation <https://nextcloud.github.io/nextcloud-router/>`_
+
+
+``@nextcloud/vue``
+^^^^^^^^^^^^^^^^^^
+
+This package provides a component library.
+`Read the full documentation <https://github.com/nextcloud/nextcloud-vue/#documentation>`_
```
juliushaertl: I'd say we can link directly to https://nextcloud-vue-components.netlify.com/

ChristophWurst: sure! I wasn't sure if this was the final domain. but we of course always update the docs if the domain changes :)

skjnldsv:
I disagree with the OCA part. Every app that register is fine. But any app that uses them needs to wait for domcontentloaded, this works fine and it is really the only way to expose proper APIs like the viewer or the sidebar without too much annoyance I think :)

ChristophWurst:
> I disagree with the OCA part. Every app that register is fine. But any app that uses them needs to wait for domcontentloaded, this works fine and it is really the only way to expose proper APIs like the viewer or the sidebar without too much annoyance I think :)

I know, thus

> Except for edge cases with inter-app communication, this use of this approach is discouraged as well

Apps used to assign all their Backbone models and views to OCA. That should go away. For the edge case like with viewer it's fine, though we could hide the global variable in a tiny wrapper package.

skjnldsv:
> > Except for edge cases with inter-app communication, this use of this approach is discouraged as well
> 
> Apps used to assign all their Backbone models and views to OCA. That should go away. For the edge case like with viewer it's fine, though we could hide the global variable in a tiny wrapper package.

Yeah, I thought is was not very clear. We should maybe be more specifics?

ChristophWurst:
> Yeah, I thought is was not very clear. We should maybe be more specifics?

Please see updated text.

developer_manual/app/javascript-apis.rst
```diff
@@ -0,0 +1,159 @@
...
+^^^^^^^^^^^^^^^^^^^^^
+
+This package provides helpers to generate URLs, e.g. to access assets and REST APIs of your app or the Nextcloud server.
+`Read the full documentation <https://nextcloud.github.io/nextcloud-router/>`_
+
+
+``@nextcloud/vue``
+^^^^^^^^^^^^^^^^^^
+
+This package provides a component library.
```
skjnldsv:
```suggestion
This package provides lots of nextcloud components allowing you to quickly build UIs.
```
? Or something?

ChristophWurst: sure :)

ChristophWurst:
:pray:

PR #1808, opened 2020-01-31
Add info about activating a cron job

Moini:
Changes:

* a cron job needs to be activated to work, this MR adds the necessary command.
* the Nextcloud web interface suggests to repeat the job every 15 minutes. The documentation is consistent with this suggestion now.
* added info that the line may need to be adjusted (because that info is only mentioned later, and people who copy-paste might see it too late)

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -64,15 +64,15 @@ Using the operating system cron feature is the preferred method for executing
 regular tasks.  This method enables the execution of scheduled jobs without the
 inherent limitations the Web server might have.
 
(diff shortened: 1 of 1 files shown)


tcitworld:
There's no need to restart the cron service.

Moini:
I had to restart it - before that, the web interface would tell me that there's no cron job set up. Maybe it varies?

Moini:
It does vary, according to comments at https://stackoverflow.com/questions/10193788/restarting-cron-after-changing-crontab-file

tcitworld:
> I had to restart it - before that, the web interface would tell me that there's no cron job set up.

Because you need to wait for the first time it triggers.

kesselb:
Thanks :+1: 5 minutes is the correct value as per https://github.com/nextcloud/documentation/issues/1587

PR #1813, opened 2020-02-04
typographic fixes, remove redundancy, fix for TB 60 added

vasyugan:
also removed some colloquial language..


vasyugan:
@skjnldsv Could you maybe review this request? It has been in the queue for 16 days no. I am not sure how quick these things are supposed to be, though...

MorrisJobke:
The other back ports are in 1976

PR #1817, opened 2020-02-06
Add instructions for docker Nextcloud 18 image

BartolomeSintes:
I would say that the suggested occ command does not work in the official docker image of Nextcloud 18. The url of occ is different and sudo can not be used inside docker. These two commands have worked for me instead:
$ sudo docker exec -u www-data -it nextcloud /bin/bash
www-data@:~/html$ php /var/www/html/occ user:resetpassword admin

### DIFF
--- a/admin_manual/configuration_user/reset_admin_password.rst
+++ b/admin_manual/configuration_user/reset_admin_password.rst
@@ -24,3 +24,15 @@ learn more about using the ``occ`` command.
  
 If your Nextcloud username is not ``admin``, then substitute your Nextcloud 
 username.
+
+In the official Docker image of Nextcloud, the ``occ`` command would be:
+
+::
+
+ $ sudo docker exec -u www-data -it nextcloud /bin/bash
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hey! Which image are you using?
The commands are working fine for me :)

kesselb:
Had a similar request a nextcloud/server from a docker user. I recall that `occ` is used as placeholder. How to execute occ depends on your environment. The usage of occ with docker is documented at: https://github.com/nextcloud/docker#using-the-nextcloud-command-line-interface. I guess there is nothing wrong with adding a example for docker and then someone will add a example for snap, and the vm, .....

skjnldsv:
Well, then it would be harder to maintain the docs.
I would rather have everything on the docker repo, duplicating data would just lead to out of sync ressources :)

juliushaertl:
Yes, we should keep the docs generic in this regard. Otherwise we end up with unmaintainable long lists of examples for different setups as @kesselb mentioned.

PR #1691, opened 2019-11-09
Add service discovery example for HAProxy

antoneliasson:

PR #1829, opened 2020-02-17
Add HAProxy .well-known redirect

juliushaertl:


kesselb:
If you merge this one please close https://github.com/nextcloud/documentation/pull/1691. Your version is a easier to understand. I thought that `dav` would be something haproxy related but it's just the name of a variable set if the request matches a given path. `url_discovery` as name is better :+1:

kesselb:
@juliushaertl could you rebase the branch or restart sphinx check?

PR #1833, opened 2020-02-20
Serve webm, mp4 files directly in nginx

juliushaertl:


skjnldsv:
We should have some tests for those :)
In the nextcloud admin setup tests page

MorrisJobke:
> We should have some tests for those :)
> In the nextcloud admin setup tests page

Yep .. this caused https://github.com/nextcloud/server/issues/21180

PR #1837, opened 2020-02-23
corrected section 'disabling encryption'

phonon112358:
If maintenance mode is enabled (like the current documentation recommends to do!), disabling encryption results in a complete loss of data since the encryption module is not loaded in maintenance mode (see https://github.com/nextcloud/server/issues/8311)!!!
I added some additional warnings about encryption, too.

### DIFF
--- a/admin_manual/configuration_files/encryption_configuration.rst
+++ b/admin_manual/configuration_files/encryption_configuration.rst
@@ -6,6 +6,8 @@ The primary purpose of the Nextcloud server-side encryption is to protect users'
 files on remote storage, such as Dropbox and Google Drive, and to do it easily 
 and seamlessly from within Nextcloud.
 
+However, there are many downsides of the server-side encryption in Nextcloud, e.g., many useful apps may not work (properly) and the speed of synchronisation between clients and server may be very slow. Hence, think twice before you enable encryption, especially since disabling encryption can cause a loss of all your data.
+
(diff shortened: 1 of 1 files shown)


kesselb:
How do you run decrypt-all if the maintenance mode was enabled? https://github.com/nextcloud/server/pull/12114 added a warning for that case.

phonon112358:
I ran `occ encryption:disable` as it stands in the current documentation. 
the execution wasn't blocked.
2020-02-23 at 22-13-42

kesselb:
OK. I guess we need some advice to decrypt all your data before disabling the encryption. That's the issue you run into. Disable encryption without decrypting the files before.

cc @nextcloud/encryption @rullzer for input how to clarify.

phonon112358:
> OK. I guess we need some advice to decrypt all your data before disabling the encryption. That's the issue you run into. Disable encryption without decrypting the files before.
> 
> cc @nextcloud/encryption @rullzer for input how to clarify.

Yes, that would be awesome for a future release...!

However, I guess this will take a lot more effort than adjusting the documentation. Hence, this pull request might be seen as an intermediate step. Why not edit the doc and then look for more general and certainly much better implementations for the app?!

joshtrichards:
This appears no longer applicable today and the section has been since expanded:

https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/encryption_configuration.html#disabling-encryption

Okay to close this PR?

PR #1850, opened 2020-02-27
Add example Swift v3 scope

ded7:
Add an example of Swift V3 with a configuration scope following this commit :
https://github.com/nextcloud/server/pull/9280


MorrisJobke:
> /backport to stable17

See #2006 (also for the 16)

PR #1852, opened 2020-02-27
Use more inclusive pronouns

ahstro:
This replaces instances of gendered pronouns (e.g. 'she' and 'he') with an equivalent variant of singular 'they'.

I was reading the docs and _he_ was being used to refer to _the developer_, which seems counter intuitive, what with wanting more diversity in tech and all that 😅

I used a script to convert the pronouns, and when I went through the changes with `git add --patch`, I noticed that some of the pronouns referred to actual people. Now - if you don't want to be as radical as I - I see the case for keeping those the way they were, but I personally think it's good praxis to use gendered pronouns as little as possible 🤷

### DIFF
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -466,7 +466,7 @@ This example creates a new calendar for molly::
 
  sudo -u www-data php occ dav:create-calendar molly mollycal
  
-Molly will immediately see these on her Calendar and Contacts pages.
+Molly will immediately see these on their Calendar and Contacts pages.
 
 ``dav:lists-calendars [user]`` will display a table listing the calendars for an given user. 
 This example will list all calendars for user annie::
@@ -1045,7 +1045,7 @@ memberships with the ``user:add`` command. The syntax is::
 
 The ``display-name`` corresponds to the **Full Name** on the Users page in your 
 Nextcloud Web UI, and the ``uid`` is their **Username**, which is their 
-login name. This example adds new user Layla Smith, and adds her to the 
+login name. This example adds new user Layla Smith, and adds them to the 
 **users** and **db-admins** groups. Any groups that do not exist are created:: 
  
  sudo -u www-data php occ user:add --display-name="Layla Smith" 
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -466,7 +466,7 @@ This example creates a new calendar for molly::
 
  sudo -u www-data php occ dav:create-calendar molly mollycal
  
-Molly will immediately see these on their Calendar and Contacts pages.
+Molly will immediately see these in the Calendar and Contacts apps.
 
 ``dav:lists-calendars [user]`` will display a table listing the calendars for an given user. 
 This example will list all calendars for user annie::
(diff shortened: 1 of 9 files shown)


skjnldsv:
Hey @ahstro!
Thank you very much for taking care of this! :hugs: 

cc @jancborchardt

admin_manual/configuration_server/occ_command.rst
```diff
@@ -466,7 +466,7 @@ This example creates a new calendar for molly::
 
  sudo -u www-data php occ dav:create-calendar molly mollycal
  
-Molly will immediately see these on her Calendar and Contacts pages.
+Molly will immediately see these on their Calendar and Contacts pages.
```
tflidd: For all the sharing-related topics, I see a lot of potential for confusion: Molly wants to see her calendar on her Calendar and Contacts page and she doesn't want to see her calendar on the Calendar or Contacts page of anyone else (unless she shared it, or is it shared by default?).

admin_manual/configuration_user/user_configuration.rst
```diff
@@ -186,13 +186,13 @@ Disable and enable users
 
 .. figure:: ../images/users-actions.png
 
-Sometimes you may want to disable a user without permanently deleting his
+Sometimes you may want to disable a user without permanently deleting their
```
tflidd: Why not use the plural of user in the first place?

tflidd:
For me these inclusive pronouns make texts harder to understand (non-native speaker). For certain things you might get used to it. However, everything related to sharing, it can get a bit tricky with more inclusive pronouns: It is a huge difference if you share something to one person, to a group or if you keep it to yourself. For that reason, I think the female users are a good idea and it keeps everything clear and understandable.

admin_manual/configuration_user/user_configuration.rst
ahstro: What do you mean? Could you give an example sentence? 😊

ahstro:
Hmm, yeah, I'm so used to using gender neutral pronouns that I don't really register when it might be confusing for people not used to it, so these are definitely valid concerns. Could you point out the specific pages which you find confusing? I could try to reword them to both be less confusing and use more inclusive pronouns, or if people would rather use gendered ones, I could revert them.

admin_manual/configuration_user/user_configuration.rst
nickvergessen:
```suggestion
Sometimes you may want to disable users without permanently deleting them
```

nickvergessen:
```suggestion
Sometimes you may want to disable users without permanently deleting them
```

ahstro: Oh, I see. Well, it didn't really strike me as something to do. If y'all want to change it that way, I don't mind; should still be `their` though, to make sense with the rest of the sentence: `settings and files. The user can be activated any time again, without data-loss.` 😊

admin_manual/configuration_server/occ_command.rst
jancborchardt: Here’s a better sentence to remove confusion:
```suggestion
Molly will immediately see these in the Calendar and Contacts apps.
```

admin_manual/configuration_user/user_configuration.rst
jancborchardt: This change seems totally fine as is. :)

PR #1860, opened 2020-03-04
Update Primary Storage doc for OpenStack Swift v3

cjlano:
Add scope/project with TenantName, as discussed in #11264 comment https://github.com/nextcloud/server/issues/11264#issuecomment-457525893 
This is necessary for OVH Object Storage. Not sure for other providers.

### DIFF
--- a/admin_manual/configuration_files/primary_storage.rst
+++ b/admin_manual/configuration_files/primary_storage.rst
@@ -91,12 +91,20 @@ V3 Authentication:
 		'arguments' => array(
 			'autocreate' => true,
 			'user' => [
-				'name' => 'swift',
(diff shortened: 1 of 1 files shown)


kesselb:
Looks similar to https://github.com/nextcloud/documentation/pull/1850. I would prefer this one because it only changes the relevant lines.

MorrisJobke:
> Looks similar to #1850. I would prefer this one because it only changes the relevant lines.

I fixed the other PR 🙈

MorrisJobke:
Let us use the username and tenant name variables from this PR and then we are good :) 


Thanks @cjlano anyways and sorry for this duplicate work needed here.

PR #1867, opened 2020-03-06
HAProxy example: Improved formating

Fonata:
It was not obvious that it must be two lines.


Fonata:
I amended my commit with a --signoff to make DCO happy.

PR #1868, opened 2020-03-07
Add example for apps store url

kesselb:
Recently someone at `nextcloud/server` used `https://api.nextcloud.com/v1` as app store url and reported a issue about a broken certificate. It took a while to convince him that not the certificate is the issue but the wrong app store url. 

Question: `The appstore server must use OCS (Open Collaboration Services).` is that still true?


nickvergessen:
Doesn't look like it

kesselb:
> Doesn't look like it

I removed the sentence.

PR #1872, opened 2020-03-09
Update Traefik config for v2

fengkaijia:
Add Traefik v2 config as the grammar is slightly different than v1.


skjnldsv:
Let's ignore then

PR #1896, opened 2020-03-23
clarified trash bin retention by storage shortage override

wiswedel:
Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>


wiswedel:
note: There are internal links which are pointing to /latest/. @juliushaertl shall we adjust that in backports or just leave it that way? I'd say let's leave it so we don't have to update it with each release.

skjnldsv:
> note: There are internal links which are pointing to /latest/. @juliushaertl shall we adjust that in backports or just leave it that way? I'd say let's leave it so we don't have to update it with each release.

fine to leave latest I think

PR #1909, opened 2020-03-28

bkmgit:
Indicate can also use PHP 7.4

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -34,7 +34,7 @@ If you get a result, the module is present.
 
 Required:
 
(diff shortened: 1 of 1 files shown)


admin_manual/installation/source_installation.rst
```diff
@@ -34,7 +34,7 @@ If you get a result, the module is present.
 
 Required:
 
-* PHP (7.1, 7.2 or 7.3)
+* PHP (7.1, 7.2, 7.3 or 7.4)
```
skjnldsv:
```suggestion
* PHP (7.2, 7.3 or 7.4)
```
Actually php 7.1 is EOL, so we should also remove it :)

PR #1908, opened 2020-03-28
Update example_centos.rst

bkmgit:
Update reference location for listing of PHP modules

PR #1913, opened 2020-03-28
Update example_centos.rst

bkmgit:
Slight update of #1908

Indicate for PHP 74 in modules
Use dnf instead of yum


bkmgit:
Performing further checks after finding an error

juliushaertl:
> Performing further checks after finding an error

Anything left to do from your side?

bkmgit:
I checked, it seems to work, so can merge.

bkmgit:
Seems like signoff needs to be done on each commit through web interface, https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330 hopefully can still approve.

bkmgit:
Would it be worth adding some information about signing off in the README https://github.com/nextcloud/documentation/blob/master/README.rst
See also https://probot.github.io/apps/dco/

PR #1920, opened 2020-03-28
updated a large bunch of screenshots

wiswedel:
Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>


wiswedel:
backport no further than 18, since some screenshots would be from the future then

PR #1931, opened 2020-03-31
Fix database file name in backup documentation

lentzi90:
I tried following the documentation and realized the file name was wrong (owncloud instead of nextcloud).


kesselb:
The default value for the database name is stil owncloud: https://github.com/nextcloud/server/blob/28f8eb5dba60a75f7e22debbdc26f1d3164deb18/lib/private/DB/ConnectionFactory.php#L44

PR #1936, opened 2020-04-02
Update documentation and readability

clementduveau:
Hi,

I propose this small documentation update.

As my reverse proxy can change its internal IP address (self-healing, rolling update, etc) I resolve it only using internal DNS. I just tried and my Nextcloud install (officiel docker, no modifications) is able to resolve it.

Options for the array have been updated for readability

### DIFF
--- a/admin_manual/configuration_server/reverse_proxy_configuration.rst
+++ b/admin_manual/configuration_server/reverse_proxy_configuration.rst
@@ -14,8 +14,12 @@ trust. Connections from trusted proxies will be specially treated to get the
 real client information, for use in access control and logging. Parameters are
 configured in :file:`config/config.php`
 
-Set the **trusted_proxies** parameter as an array of IPv4 addresses, 
-IPv4 ranges in CIDR notation or IPv6 addresses to define the
+Set the :file:`trusted_proxies` parameter as an array of:
+* IPv4 addresses, 
+* IPv4 ranges in CIDR notation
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/reverse_proxy_configuration.rst
```diff
@@ -14,8 +14,12 @@ trust. Connections from trusted proxies will be specially treated to get the
...
-IPv4 ranges in CIDR notation or IPv6 addresses to define the
+Set the :file:`trusted_proxies` parameter as an array of:
+* IPv4 addresses, 
+* IPv4 ranges in CIDR notation
+* IPv6 addresses
+* host to resolve
+to define the
 servers Nextcloud should trust as proxies. This parameter provides protection
 against client spoofing, and you should secure those servers as you would your
 Nextcloud server.
```
skjnldsv, line 25: Maybe also adjust the lines after so the code is a bit more clear?

```suggestion
to define the servers Nextcloud should trust as proxies.
This parameter provides protection against client spoofing,
and you should secure those servers as you would your
Nextcloud server.
```

MorrisJobke: Let's merge this for now and do this later.

MorrisJobke: see #2000

PR #1944, opened 2020-04-09
Fix resource-accessability

manuth:
Hey guys
This PR is related to following issue in the https://github.com/nextcloud/docker repository:
https://github.com/nextcloud/docker/issues/1055
A pull request concerning this issue is already present:
https://github.com/nextcloud/docker/pull/1056
I was facing the issue that sometimes when loading ressources an HTTP 302 code was returned which redirected me to the main page.

Rather than solving this issue by adding `.mp4` and `.webm` to the location-directive I'd rather recommend to pass all requests to the files directly.

This change can be implemented by accepting this PR.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -118,10 +118,6 @@ webroot of your nginx installation. In this example it is
       # This module is currently not supported.
       #pagespeed off;
 
-      location / {
-          rewrite ^ /index.php;
-      }
-
       location ~ ^\/(?:build|tests|config|lib|3rdparty|templates|data)\/ {
           deny all;
       }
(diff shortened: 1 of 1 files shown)


kesselb:
> Rather than solving this issue by adding .mp4 and .webm to the location-directive I'd rather recommend to pass all requests to the files directly.

That is missing for the docker example. Update the example there and everything is all right.

manuth:
@kesselb so does that mean, that this change doesn't cause any sort of security violation?

kesselb:
I have no idea what you are talking about :) My suggestion is to update the nginx configuration at nextcloud/docker to include mp4 and webm. The configuration for nginx here is already updated to include them.

skjnldsv:
Let's close then:wink:

manuth:
Well what I find pretty confusing about the current settings is that it seems to me like custom apps only canm provide resources with the specified file-extension listed here:
https://github.com/nextcloud/documentation/blob/cda627b7c8896f3bac8bfde420c0c8b5d1806190/admin_manual/installation/nginx.rst#L182

But what if an extension, let's say, serves an `.xml`- or a `.json`-file?
In this case nginx wouldn't be able to serve this document as the file-extension is not present in this list.

That's why I set up this PR. And also to ask whether this PR would cause any security issues.

kesselb:
For now apps have to serve any other file with a custom controller. I'm not against changing that but the apache (.htaccess rules) and nginx setup must be in sync. Otherwise apps might behave different causing a lot of questions / confusion at help.nextcoud.com or GitHub.

manuth:
Alright, I got it!
Thanks for the explanation ^^

I'll go update the docker-repo asap

PR #1949, opened 2020-04-10
Update deprecated globals in 18

ChristophWurst:
Ref https://github.com/nextcloud/server/pull/17402
Ref https://github.com/nextcloud/server/pull/17392


korelstar:
This should be added to https://github.com/nextcloud/eslint-plugin/blob/master/lib/rules/no-deprecations.js , too.

ChristophWurst:
Will do today :)

ChristophWurst:
The file is not in stable18 :shrug:

PR #1959, opened 2020-04-12
Add NGINX reverse proxy code block

mat-m:
Even if trivial, code block was missing for nginx as reverse proxy.


admin_manual/configuration_server/reverse_proxy_configuration.rst
```diff
@@ -90,6 +90,18 @@ HAProxy
   acl url_discovery path /.well-known/caldav /.well-known/carddav
   http-request redirect location /remote.php/dav/ code 301 if url_discovery
 
+NGINX
+^^^^^^^
```
gary-kim, line 94:
```suggestion
^^^^^
```
One super tiny nitpick.

PR #1964, opened 2020-04-14
fix some issues in upgrade-guide

korelstar:


developer_manual/app/upgrade-guide.rst
```diff
@@ -30,7 +32,7 @@ Deprecated global variables
 * ``OC.linkTo``: use ``linkTo`` from https://www.npmjs.com/package/@nextcloud/router
 * ``OC.linkToOCS``: use ``generateOcsUrl`` from https://www.npmjs.com/package/@nextcloud/router
 * ``OC.linkToRemote``: use ``generateRemoteUrl`` from https://www.npmjs.com/package/@nextcloud/router
-* ``OC.set``: use ``generateFilePath`` from use https://lodash.com/docs#set
+* ``OC.set``: use https://lodash.com/docs#set
```
skjnldsv, line 35: Can't be recommend Object.assign instead?

ChristophWurst: no, `_.set` works with "namespaces" like `OC.A.B`. And it merges. See the examples of https://lodash.com/docs/4.17.15#set. Object.assign would overwrite, right?

PR #2037, opened 2020-04-30
Do not (yet) recommended PHP 7.4

J0WI:
https://github.com/nextcloud/3rdparty/pull/435#issuecomment-622107360
https://github.com/nextcloud/3rdparty/pull/432#issuecomment-615914859
https://github.com/nextcloud/server/issues/18996

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -31,7 +31,7 @@ For best performance, stability and functionality we have documented some recomm
 +------------------+-----------------------------------------------------------------------+
 | PHP Runtime      | - 7.2                                                                 |
 |                  | - **7.3** (*recommended*)                                             |
-|                  | - **7.4** (*recommended*)                                             |
+|                  | - 7.4                                                                 |
 +------------------+-----------------------------------------------------------------------+
 
 See :doc:`source_installation` for minimum PHP-modules and additional software for installing Nextcloud.


MorrisJobke:
Most of this seem to be fixed in https://github.com/nextcloud/server/pull/20794

J0WI:
Let's see if we still need it on 18-stable or if it gets backported.

tflidd:
What is the status here? The new release comes out soon, so people will come check for updates/upgrades and work on their systems.

ChristophWurst:
> Most of this seem to be fixed in [nextcloud/server#20794](https://github.com/nextcloud/server/pull/20794)

@tflidd doesn't that fix the issues?

J0WI:
Superseded by #2117

abysso2:
Just installed 18.0.5 ...

root@nextcloud:~# grep VER /var/www/nextcloud/3rdparty/leafo/scssphp/src/Version.php
const VERSION = 'v0.7.6';

So there ist no update to 0.8.4 as mentioned here: nextcloud/3rdparty#361

Is the php7.4-error still existant????

J0WI:
Yes, as you can read in the comments above the fix is only in Nextcloud 19 and has not been backported to 18.

kesselb:
Actually there are no stupid questions but @abysso2 comment is very close 😉

PR #2103, opened 2020-05-26
Extend whitelist for php entrypoints

juliushaertl:


timar:
@juliushaertl you wrote "apps" in the patch not that what you sent me.

juliushaertl:
Aw, good catch. I copied the wrong line there. Fixed that to `.+` so we cover all different app directories.

PR #2117, opened 2020-05-30
Do not (yet) recommended PHP 7.4

J0WI:


rullzer:
Why not?

ChristophWurst:
https://github.com/nextcloud/documentation/pull/2037#issuecomment-623436839 but I do not think this is necessarily true anymore.

kesselb:
There are some warnings with 7.4.

rullzer:
Warnings are not really a reason not to recommend something.
We should just fix them. I am running nextcloud for some months now with php7.4 and it is working just fine.

kesselb:
> I am running nextcloud for some months now with php7.4 and it is working just fine.

You are a developer. That's a different story ;) Admin checks the logs, see warnings, searches for a reason. It's often hard to judge for them if the log message is critical or not. 

https://github.com/nextcloud/groupfolders/issues/907
https://github.com/nextcloud/server/issues/18996

ChristophWurst:
But you fixed those, right?

kesselb:
> But you fixed those, right?

The patch with SCSS is not backported yet. As the original patch is for scssphp 1.03 and we still have scssphp 0.7.x with stable18 not sure if it will.

rullzer:
ok then lets do it for stable18.

J0WI:
18 is EOL

PR #2136, opened 2020-06-07
Update sync_ios.rst

maxgrass:
verified with the nextcloud 19


BernieO:
This makes adding a caldav-account to iOS more complicated as it actually is as long as service-discovery is working, which should be the case if Nextcloud is installed according to the documentation.

Is there a certain reason for changing the easier setup to a more complicated one?

maxgrass:
Ok, thanks for the clarification. The only reason was the "default URL" without full path didn't work in my case at all and I thought that the documentation was obsolete.

PR #2138, opened 2020-06-08
Document current bootstrap mechanism

ChristophWurst:
As I intend to change this, I want to document the current state and afterwards draft a docs page for the new stuff. This shall highlight the difference and show the advantages of https://github.com/nextcloud/server/pull/20865.

app.php and Application.php were briefly mentioned, but the details that matter were not. This tiny updated clarifies on which file gets loaded for every request and what to do in order to get your Application.php loaded properly.

Ref https://github.com/nextcloud/server/issues/20573

### DIFF
--- a/developer_manual/app/init.rst
+++ b/developer_manual/app/init.rst
@@ -60,38 +60,3 @@ purpose there are several events emitted that an app can act upon.
             \OCP\Util::addStyle('myapp', 'style');
         }
     );
-
-Best practice
--------------
-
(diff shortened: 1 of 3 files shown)


developer_manual/app/init.rst
```diff
@@ -61,37 +61,7 @@ purpose there are several events emitted that an app can act upon.
...
-
-.. code-block:: php
-
-    <?php
-
-    $app = new \OCA\MyApp\AppInfo\Application();
-    $app->registerHooks();
-
 
 lib/AppInfo/Application.php
```
juliushaertl, line 65: Can be removed as well

PR #2139, opened 2020-06-08
Document the need to symlink apps outside server root

zwets:
Document that any apps path outside the server root must be symlinked from inside the server root.

See https://github.com/nextcloud/server/issues/21292.


MorrisJobke:
@zwets Mind to resolve the conflicts?

zwets:
@MorrisJobke rebased on master and fixed.

PR #2163, opened 2020-06-21
Add @PublicPage to get() method tags

skaramicke:
I could not get this example to work without adding @PublicPage to the PublicAPIController->get() method. In fairness, my implementation only validates token, and returns false in isPasswordProtected.

### DIFF
--- a/developer_manual/app/publicpage.rst
+++ b/developer_manual/app/publicpage.rst
@@ -74,6 +74,7 @@ As said the PublicShareController is a very basic controller. You need to implem
 
 		/**
 		 * Your normal controller function
+		 * @PublicPage
 		 */
 		public function get() {
 			// Work your magic


MorrisJobke:
@ChristophWurst Should we also adjust the text around this? cc @skjnldsv

skjnldsv:
cc @juliushaertl

ChristophWurst:
> @ChristophWurst Should we also adjust the text around this? cc @skjnldsv

Which part specifically?

MorrisJobke:
Adding a section that describes that the `get` method needs to be annotated like this to allow unauthenticated access to it.

ChristophWurst:
Sure that won't hurt.

nickvergessen:
Rebased in https://github.com/nextcloud/documentation/pull/8092

PR #2165, opened 2020-06-22
Restructure the dev manual

ChristophWurst:
This moves lots of pages around. The high-level changes are

* Better main sections, so it's more *general*, *into*, *basics* and *details*
* Move more general topics to a *Basic* section, which are not
  app-specific
* Remove app docs to the stuff that is likely used, anything else goes
  into "Digging deeper"
* Move general guides into a prologue
* Try to *compress*/combine some pages with similar content
* Try to have better consistencs on level ob abstraction across pages
* Split app development and maintenance pages into two sections
* Integrate bugtracker info into prologue
* Integrate Android pages into client APIs section

There is a lot of changes that I can't recall (this was quite a lot of back and forth until I was satisfied with the structure). I don't think there is a specific ticket for this but I've mentioned a few times that I was not happy with this. I hope this also helps us see the spots where we lack some details.

This is heavily inspired by https://laravel.com/docs/7.x, hence the similar section titles.

Let me know what you think. I'm happy to discuss further improvements :pray: 

---

### Before
Screenshot_2020-06-22 Nextcloud developer documentation — Nextcloud latest Developer Manual latest documentation(1)

### After
Screenshot_2020-06-22 Nextcloud developer documentation — Nextcloud latest Developer Manual latest documentation


developer_manual/digging_deeper/index.rst
```diff
@@ -0,0 +1,24 @@
...
+   :maxdepth: 2
+
+   api
+   changelog
+   debugging
+   classloader
+   flow
+   javascript-apis
+   npm
+   notifications
```
juliushaertl, line 15: Maybe we should move those kind of integrations like notifications, flow, 2fa to a separate sub-category?

developer_manual/getting_started/debugging.rst
```diff
@@ -0,0 +1,89 @@
...
+- The Easiest XDebug (Firefox): https://addons.mozilla.org/en-US/firefox/addon/the-easiest-xdebug/
+- XDebug Helper (Chrome): https://chrome.google.com/extensions/detail/eadndfjplgieldjbigjakmdgkmoaaaoc
+
+For debugging scripts on the command line, like ``occ`` or unit tests, set the ``XDEBUG_CONFIG`` environment variable.
+
+
+Debugging JavaScript
+--------------------
+
+By default all JavaScript files in Nextcloud are minified (compressed) into a single file without whitespace. To prevent this, see Debug mode.
```
juliushaertl, line 77: Maybe add a note about source maps here, since debug mode doesn't help for any bundled js

ChristophWurst: I did not actually change this file. I think I just renamed or moved it :see_no_evil:

developer_manual/digging_deeper/index.rst
ChristophWurst: Just for integrations? I guess that would make sense as well. My consideration was that *Digging deeper* contains the specific stuff you don't need for the average app, but will look up if really needed. Integrations are part of this in a way, I think.

ChristophWurst: I was more thinking of even lowering the number of sections. Like combining *Getting started* and *The basics* :thinking:

ChristophWurst: So I'm really on the fence with this :speak_no_evil: 

Let's think about it. We can move things around later on.

MorrisJobke:
Thanks @ChristophWurst 👍

MorrisJobke:
Somehow this is not deployed 🤔 cc @rullzer

nickvergessen:
https://docs.nextcloud.com/server/latest/developer_manual/ yes it is.
But you need to go to latest instead of stable

MorrisJobke:
Thanks :)

marcelklehr:
Thank you @ChristophWurst for all your efforts!! :heart:

PR #2164, opened 2020-06-22
Document new bootstrap process

ChristophWurst:
For https://github.com/nextcloud/server/pull/20865

PR #2166, opened 2020-06-23
Document unified search API

ChristophWurst:
- [x] Blocked by the API changes from https://github.com/nextcloud/server/pull/20916
- [x] Requires https://github.com/nextcloud/documentation/pull/2164

### DIFF
--- a/developer_manual/app_development/bootstrap.rst
+++ b/developer_manual/app_development/bootstrap.rst
@@ -1,3 +1,5 @@
+.. _Bootstrapping:
+
 =============
 Bootstrapping
 =============
--- a/developer_manual/digging_deeper/index.rst
+++ b/developer_manual/digging_deeper/index.rst
@@ -18,6 +18,7 @@ Digging deeper
    publicpage
    repair
    rest_apis
+   search
    settings
    testing
    two-factor-provider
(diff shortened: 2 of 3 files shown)


kesselb:
cc @marcelklehr @raimund-schluessler does that make sense to you? ;)

kesselb:
Also @korelstar for notes ;)

marcelklehr:
LGTM.

Currently, a lot of apps override the default search bar in the menu bar to hook in to their custom search functionality. How do we deal with that? Is that functionality expected to be removed in favor of this new search API?

ChristophWurst:
> Currently, a lot of apps override the default search bar in the menu bar to hook in to their custom search functionality. How do we deal with that? Is that functionality expected to be removed in favor of this new search API?

Most likely, yeah. But the UI part is still to do, so there's a lot that might change. :eyes: @skjnldsv @jancborchardt 

Ref https://github.com/orgs/nextcloud/projects/38

jancborchardt:
> Currently, a lot of apps override the default search bar in the menu bar to hook in to their custom search functionality. How do we deal with that? Is that functionality expected to be removed in favor of this new search API?

Yes, so the idea is that the search bar in the header will become the unified search and work the same anywhere.

In-app search would then move into each app, ideally to the top of the AppNavigation to filter content as-you-type. A great example there is Talk, and the same would be possible design-wise for Files, Mail, Contacts, Activity, Bookmarks, Notes, etc. :)

PR #2168, opened 2020-06-25
Add docs for new dashboard api

juliushaertl:
For https://github.com/nextcloud/server/pull/21346

Let me know if we should describe something more verbose, I'm always not sure how much knowledge we should assume with such docs.

### DIFF
--- a/developer_manual/digging_deeper/dashboard.rst
+++ b/developer_manual/digging_deeper/dashboard.rst
@@ -1,0 +1,178 @@
+=========
+Dashboard
+=========
+
+The dashboard app aims to provide the user with a general overview of their
+Nextcloud and shows information that is currently important.
+
+App developers can integrate into the dashboard app and provide their own widgets.
+
+
+Register a dashboard widget
+---------------------------
+
+A dashboard widget is represented by a class implementing the `OCP\\Dashboard\\IWidget`
+interface. This class will be instantiated whenever the dashboard is loaded.
+Any bootstrap code that is needed for the widget can be implemented inside
+of the `load` method and will be called when the dashboard is loaded.
+
+.. code-block:: php
+
+    <?php
+
+    namespace OCA\MyApp\Dashboard;
+
+    use OCP\Dashboard\IWidget;
+    use OCP\IInitialStateService;
+    use OCP\IL10N;
+    use OCP\IURLGenerator;
+    use OCP\IUserSession;
+
+    class MyAppWidget implements IWidget {
+
+        public function __construct(
+            IInitialStateService $initialStateService,
+            IL10N $l10n,
+            IURLGenerator $urlGenerator
+        ) {
+            $this->initialStateService = $initialStateService;
+            $this->l10n = $l10n;
+            $this->urlGenerator = $urlGenerator;
+        }
+
+        /**
+         * @return string Unique id that identifies the widget, e.g. the app id
+         * @since 20.0.0
+         */
+        public function getId(): string {
+            return 'myappwidgetid';
+        }
+
+        /**
+         * @return string User facing title of the widget
+         * @since 20.0.0
+         */
+        public function getTitle(): string {
+            return $this->l10n->t('My app');
+        }
+
+        /**
+         * @return int Initial order for widget sorting
+         *   in the range of 10-100, 0-9 are reserved for shipped apps
+         * @since 20.0.0
+         */
+        public function getOrder(): int {
+            return 0;
+        }
+
+        /**
+         * @return string css class that displays an icon next to the widget title
+         * @since 20.0.0
+         */
+        public function getIconClass(): string {
+            return 'icon-class';
+        }
+
+        /**
+         * @return string|null The absolute url to the apps own view
+         * @since 20.0.0
+         */
+        public function getUrl(): ?string {
+            return $this->urlGenerator->linkToRouteAbsolute('myapp.view.index');
+        }
+
+        /**
+         * Execute widget bootstrap code like loading scripts and providing initial state
+         */
+        public function load(): void {
+            $initialStateService->provideInitialState('myapp', 'myData', []);
+            \OCP\Util::addScript('myapp', 'dashboard');
+        }
+    }
+
+
+The class needs to be registered during the :ref:`app bootstrap<Bootstrapping>`.
+
+.. code-block:: php
+
+    <?php
+
+    declare(strict_types=1);
+
+    namespace OCA\MyApp\AppInfo;
+
+    use OCP\AppFramework\App;
+    use OCP\AppFramework\Bootstrap\IBootContext;
+    use OCP\AppFramework\Bootstrap\IBootstrap;
+    use OCP\AppFramework\Bootstrap\IRegistrationContext;
+    use OCA\MyApp\Dashboard\MyAppWidget;
+
+    class Application extends App implements IBootstrap {
+
+        public const APP_ID = 'myapp';
+
+        public function __construct(array $urlParams = []) {
+            parent::__construct(self::APP_ID, $urlParams);
+        }
+
+        public function register(IRegistrationContext $context): void {
+            $context->registerDashboardWidget(MyAppWidget::class);
+        }
+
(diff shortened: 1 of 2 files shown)


developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,150 @@
...
+
+App developers can integrate into the dashboard app and provide their own panels.
+
+
+Register a dashboard panels
+===========================
+
+A dashboard panel is represented by a class implementing the `OCP\\Dashboard\\IPanel` 
+interface. This class will be instantiated whenever the dashboard is loaded, so 
+scripts and initial state that your panel needs can be added in the constructor.
```
ChristophWurst, line 16: I guess this should be raised in the PR rather than here, but I personally don't like seeing this kind of *active code* in a constructor. In my understanding, constructors should only do the things necessary to initialize the newly constructed object.

This pattern makes it also a bit harder/weirder to unit test. As soon as you create an object, a side effect happens.

An alternative would be do have like a dedicated method ``load`` or something that is invoked when apps should run these tasks (initial state, scripts, ...). Would that make sense?

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,150 @@
...
+
+        public function getOrder(): int {
+            return 0;
+        }
+
+        public function getIconClass(): string {
+            return 'icon-class';
+        }
+
+        public function getUrl(): string {
```
ChristophWurst, line 60: what should this URL point to?

developer_manual/digging_deeper/dashboard.rst
juliushaertl: Adjusted the server PR and docs here

developer_manual/digging_deeper/dashboard.rst
juliushaertl: I've added a description in phpdoc for each method now

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,176 @@
...
+        ) {
+            $this->initialStateService = $initialStateService;
+            $this->l10n = $l10n;
+            $this->urlGenerator = $urlGenerator;
+        }
+
+        /**
+         * @return string Unique id that identifies the panel, e.g. the app id
+         * @since 20.0.0
+         */        public function getId(): string {
```
ChristophWurst, line 46: the whitespaces are a bit off. should be a line break

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,176 @@
...
+        /**
+         * @return string User facing title of the panel
+         * @since 20.0.0
+         */
+        public function getTitle(): string {
+            return $this->l10n->t('My app');
+        }
+
+        /**
+         * @return int Initial order for panel sorting
```
ChristophWurst, line 59: initial? is there a way for the user to re-order?

ChristophWurst: can we define some soft constraints on the value range? lik 0-10/1-100/whatever so it's easier for devs to pick something reasonable?

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,176 @@
...
+        /**
+         * @return string css class that displays an icon next to the panel title
+         * @since 20.0.0
+         */
+        public function getIconClass(): string {
+            return 'icon-class';
+        }
+
+        /**
+         * @return string|null The absolute url to the apps own view
```
ChristophWurst, line 75: what about apps that don't have this? should we consider those and make this nullable?

developer_manual/digging_deeper/dashboard.rst
juliushaertl: Yes, reordering can be performed by the user. I'm not sure about the range, maybe just 1-100 as that seems reasonable and we reserve 1-10 for shipped apps?

developer_manual/digging_deeper/dashboard.rst
juliushaertl: Fixed, I forgot to also add the nullable here, it is already in place in the interface:
https://github.com/nextcloud/server/blob/ea46d1a3a2e99b77cb35dd5eefa895406cbceb58/lib/public/Dashboard/IPanel.php#L62

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
+=========
+Dashboard
+=========
+
+The dashboard app aims to provide the user with a general overview of their
```
ChristophWurst, line 5: nit: wanna link to the app on the app store for those who want to have a quick look at what that is?

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
...
+         * @return string User facing title of the panel
+         * @since 20.0.0
+         */
+        public function getTitle(): string {
+            return $this->l10n->t('My app');
+        }
+
+        /**
+         * @return int Initial order for panel sorting
+         *   in the range of 10-100, 0-10 are reserved for shipped apps
```
ChristophWurst, line 61: nit: is 10 now a for shipped apps or the lower allowed value for others? :upside_down_face:

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
...
+         * Execute panel bootstrap code like loading scripts and providing initial state
+         */
+        public function load(): void {
+            $initialStateService->provideInitialState('myapp', 'myData', []);
+            \OCP\Util::addScript('myapp', 'dashboard');
+        }
+    }
+
+
+The class needs to be registered during the app bootstrap.
```
ChristophWurst, line 94: could you link to the bootstrapping page for those that read *app bootstrap* for the first time? There is a reference label `Bootstrapping` that you can use.

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
...
+                    $event->registerPanel(MyAppPanel::class);
+            });
+        }
+    }
+
+
+Provide a user interface
+------------------------
+
+The user interface can be registered though the public `OCA.Dashboard.register`
```
ChristophWurst, line 154:
```suggestion
The user interface can be registered through the public `OCA.Dashboard.register`
```

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
...
+JavaScript method. The first parameter represents the panel id that has already
+been specified in the `IPanel` implementation. The callback parameter will be
+called to render the panel in the frontend. The user interface can be added to
+the provided DOM element `el`.
+
+The following example shows how a Vue.js component could be used to render the
+panel user interface, however this approach works for any other framework as well
+as plain JavaScript as well:
```
ChristophWurst, line 162:
```suggestion
as plain JavaScript:
```

one *as well* too many :P

ChristophWurst:
Erm I mean to request changes … you get it :)

developer_manual/digging_deeper/dashboard.rst
juliushaertl: It will be shipped, I'll extend with screenshots once we have something nice ;)

juliushaertl:
Thanks, pushed a fixup

developer_manual/digging_deeper/dashboard.rst
```diff
@@ -0,0 +1,178 @@
...
+Dashboard
+=========
+
+The dashboard app aims to provide the user with a general overview of their
+Nextcloud and shows information that is currently important.
+
+App developers can integrate into the dashboard app and provide their own panels.
+
+
+Register a dashboard panels
```
splitt3r, line 11:
```suggestion
Register a dashboard panel
```

juliushaertl:
Also reflects the API naming change we did, ready to merge.

PR #2197, opened 2020-07-11
Update nginx.rst -- Reworked the Nginx configs

jivanpal:
After inspecting the Nginx configs currently seen at [ https://docs.nextcloud.com/server/19/admin_manual/installation/nginx.html ], I have notice that they are very redundant in some areas, and handle specific paths in special ways which are not needed, and which run the risk of needing to become more specific down the road if NextCloud adds new features, as well as not handling all static filetypes. I have reworked the configs from scratch by directly adapting the `.htaccess` files, and this is the result. Comments are included in the file to give rationale/explanation for why things are written in this new way, and if the maintainers decide to accept this pull request, I will leave it up to them to decide whether to include these expository comments upstream.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -105,13 +105,17 @@ webroot of your nginx installation. In this example it is
       # `/updater`, `/ocm-provider`, `/ocs-provider`), and thus
       # `try_files $uri $uri/ /index.php$request_uri`
       # always provides the desired behaviour.
-      index index.php index.html index.htm /index.php$request_uri;
+      index index.php index.html /index.php$request_uri;
       
       # Default Cache-Control policy
       expires 1m;
       
       # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
-      if ( $http_user_agent ~ DavClnt ) { rewrite ^/$ /remote.php/webdav/ redirect; }
+      location = / {
+          if ( $http_user_agent ~ DavClnt ) {
+              return 302 /remote.php/webdav/$is_args$args;
+          }
+      }
       
       location = /robots.txt {
           allow all;
@@ -120,8 +124,9 @@ webroot of your nginx installation. In this example it is
       }
       
       # Make a regex exception for `/.well-known` so that clients can still
-      # access it despite the existence of the regex rule `location ~ /\.`
-      # which would otherwise handle requests for `/.well-known`.
+      # access it despite the existence of the regex rule
+      # `location ~ /(\.|autotest|...)` which would otherwise handle requests
+      # for `/.well-known`.
       location ^~ /.well-known {
           # The following 6 rules are borrowed from `.htaccess`
       
@@ -136,11 +141,7 @@ webroot of your nginx installation. In this example it is
       
       # Rules borrowed from `.htaccess` to hide certain paths from clients
       location ~ ^/(build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
-      location ~ ^/(autotest|occ|issue|indie|db_|console)                 { return 404; }
-      
-      # Respond with HTTP 404 to *all* requests for hidden files;
-      # not just hidden files that reside in the root directory.
-      location ~ /\. { return 404; }
+      location ~ ^/(\.|autotest|occ|issue|indie|db_|console)              { return 404; }
       
       # Ensure this block, which passes PHP files to the PHP process, is above the blocks
       # which handle static assets (as seen below). If this block is not declared first,
@@ -166,13 +167,13 @@ webroot of your nginx installation. In this example it is
       }
       
       location ~ \.(css|js|svg|gif)$ {
-          try_files $uri $uri/ /index.php$request_uri;
+          try_files $uri /index.php$request_uri;
           expires 6M;         # Cache-Control policy borrowed from `.htaccess`
           access_log off;     # Optional: Don't log access to assets
       }
       
       location ~ \.woff2?$ {
-          try_files $uri $uri/ /index.php$request_uri;
+          try_files $uri /index.php$request_uri;
           expires 7d;         # Cache-Control policy borrowed from `.htaccess`
           access_log off;     # Optional: Don't log access to assets
       }
@@ -195,6 +196,7 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
 - The string ``/nextcloud`` is prepended to all prefix paths.
 - The URI ``/nextcloud`` is *aliased* to ``/var/www/nextcloud``, rather than the domain itself being *rooted* at ``/var/www/nextcloud``.
 - The blocks that handle requests for paths outside of ``/nextcloud``, i.e. ``/robots.txt``, ``/.well-known``, and hidden files, are pulled out of the ``location ~ ^/nextcloud($|/)`` block.
+- The block which handles `/.well-known` doesn't need a regex exception, since the rule which prevents users from accessing hidden folders at the root of the NextCloud installation no longer matches that path.
 
 .. code-block:: nginx
 
@@ -237,11 +239,8 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
           log_not_found off;
           access_log off;
       }
-      
-      # Make a regex exception for `/.well-known` so that clients can still
-      # access it despite the existence of the regex rule `location ~ /\.`
-      # which would otherwise handle requests for `/.well-known`.
-      location ^~ /.well-known {
+
+      location /.well-known {
           # The following 6 rules are borrowed from `.htaccess`
 
           rewrite ^/\.well-known/host-meta\.json  /nextcloud/public.php?service=host-meta-json    last;
@@ -254,11 +253,6 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
 
           try_files $uri $uri/ /nextcloud/index.php$request_uri;
       }
-      
-      # Respond with HTTP 404 to *all* requests for hidden files;
-      # not just hidden files that reside in the root directory of
-      # the NextCloud installation.
-      location ~ /\. { return 404; }
       
       location ~ ^/nextcloud($|/) {
           # set max upload size
@@ -303,17 +297,21 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
           # `/updater`, `/ocm-provider`, `/ocs-provider`), and thus
           # `try_files $uri $uri/ /nextcloud/index.php$request_uri`
           # always provides the desired behaviour.
-          index index.php index.html index.htm /nextcloud/index.php$request_uri;
+          index index.php index.html /nextcloud/index.php$request_uri;
           
           # Default Cache-Control policy
           expires 1m;
 
           # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
-          if ( $http_user_agent ~ DavClnt ) { rewrite ^/nextcloud/?$ /nextcloud/remote.php/webdav/ redirect; }
+          location = /nextcloud {
+              if ( $http_user_agent ~ DavClnt ) {
+                  return 302 /nextcloud/remote.php/webdav/$is_args$args;
+              }
+          }
           
           # Rules borrowed from `.htaccess` to hide certain paths from clients
           location ~ ^/nextcloud/(build|tests|config|lib|3rdparty|templates|data)($|/)    { return 404; }
-          location ~ ^/nextcloud/(autotest|occ|issue|indie|db_|console)                   { return 404; }
+          location ~ ^/nextcloud/(\.|autotest|occ|issue|indie|db_|console)                { return 404; }
           
           # Ensure this block, which passes PHP files to the PHP process, is above the blocks
           # which handle static assets (as seen below). If this block is not declared first,
@@ -339,13 +337,13 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
           }
           
           location ~ \.(css|js|svg|gif)$ {
-              try_files $uri $uri/ /nextcloud/index.php$request_uri;
+              try_files $uri /nextcloud/index.php$request_uri;
               expires 6M;         # Cache-Control policy borrowed from `.htaccess`
               access_log off;     # Optional: Don't log access to assets
           }
           
           location ~ \.woff2?$ {
-              try_files $uri $uri/ /nextcloud/index.php$request_uri;
+              try_files $uri /nextcloud/index.php$request_uri;
               expires 7d;         # Cache-Control policy borrowed from `.htaccess`
               access_log off;     # Optional: Don't log access to assets
           }
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -76,11 +76,11 @@ webroot of your nginx installation. In this example it is
       gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
       gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
 
-      # Pagespeed is not supported by NextCloud, so if your server is built
+      # Pagespeed is not supported by Nextcloud, so if your server is built
       # with the `ngx_pagespeed` module, uncomment this line to disable it.
       #pagespeed off;
       
-      # HTTP response headers borrowed from NextCloud `.htaccess`
+      # HTTP response headers borrowed from Nextcloud `.htaccess`
       add_header Referrer-Policy                      "no-referrer"   always;
       add_header X-Content-Type-Options               "nosniff"       always;
       add_header X-Download-Options                   "noopen"        always;
@@ -188,7 +188,7 @@ Nextcloud in a subdir of the nginx webroot
 
 The following config should be used when Nextcloud is placed within a subdir of
 the webroot of your nginx installation.
-In this example the NextCloud files are located at
+In this example the Nextcloud files are located at
 ``/var/www/nextcloud`` and the Nextcloud instance is accessed via ``http(s)://cloud.example.com/nextcloud/``.
 The configuration differs from the "Nextcloud in webroot" configuration above in the following ways:
 
@@ -196,7 +196,7 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
 - The string ``/nextcloud`` is prepended to all prefix paths.
 - The URI ``/nextcloud`` is *aliased* to ``/var/www/nextcloud``, rather than the domain itself being *rooted* at ``/var/www/nextcloud``.
 - The blocks that handle requests for paths outside of ``/nextcloud``, i.e. ``/robots.txt``, ``/.well-known``, and hidden files, are pulled out of the ``location ~ ^/nextcloud($|/)`` block.
-- The block which handles `/.well-known` doesn't need a regex exception, since the rule which prevents users from accessing hidden folders at the root of the NextCloud installation no longer matches that path.
+- The block which handles `/.well-known` doesn't need a regex exception, since the rule which prevents users from accessing hidden folders at the root of the Nextcloud installation no longer matches that path.
 
 .. code-block:: nginx
 
@@ -267,11 +267,11 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
           gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
           gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
 
-          # Pagespeed is not supported by NextCloud, so if your server is built
+          # Pagespeed is not supported by Nextcloud, so if your server is built
           # with the `ngx_pagespeed` module, uncomment this line to disable it.
           #pagespeed off;
           
-          # HTTP response headers borrowed from NextCloud `.htaccess`
+          # HTTP response headers borrowed from Nextcloud `.htaccess`
           add_header Referrer-Policy                      "no-referrer"   always;
           add_header X-Content-Type-Options               "nosniff"       always;
           add_header X-Download-Options                   "noopen"        always;
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -41,67 +41,29 @@ webroot of your nginx installation. In this example it is
       listen 80;
       listen [::]:80;
       server_name cloud.example.com;
-      # enforce https
-      return 301 https://$server_name:443$request_uri;
+      
+      # Enforce HTTPS
+      return 301 https://$server_name$request_uri;
   }
 
   server {
-      listen 443 ssl http2;
+      listen 443      ssl http2;
       listen [::]:443 ssl http2;
       server_name cloud.example.com;
 
       # Use Mozilla's guidelines for SSL/TLS settings
       # https://mozilla.github.io/server-side-tls/ssl-config-generator/
-      # NOTE: some settings below might be redundant
-      ssl_certificate /etc/ssl/nginx/cloud.example.com.crt;
+      ssl_certificate     /etc/ssl/nginx/cloud.example.com.crt;
       ssl_certificate_key /etc/ssl/nginx/cloud.example.com.key;
 
-      # Add headers to serve security related headers
-      # Before enabling Strict-Transport-Security headers please read into this
-      # topic first.
-      #add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload;" always;
-      #
+      # HSTS settings
       # WARNING: Only add the preload option once you read about
       # the consequences in https://hstspreload.org/. This option
       # will add the domain to a hardcoded list that is shipped
       # in all major browsers and getting removed from this list
       # could take several months.
-      add_header Referrer-Policy "no-referrer" always;
-      add_header X-Content-Type-Options "nosniff" always;
-      add_header X-Download-Options "noopen" always;
-      add_header X-Frame-Options "SAMEORIGIN" always;
-      add_header X-Permitted-Cross-Domain-Policies "none" always;
-      add_header X-Robots-Tag "none" always;
-      add_header X-XSS-Protection "1; mode=block" always;
-
-      # Remove X-Powered-By, which is an information leak
-      fastcgi_hide_header X-Powered-By;
-
-      # Path to the root of your installation
-      root /var/www/nextcloud;
-
-      location = /robots.txt {
-          allow all;
-          log_not_found off;
-          access_log off;
-      }
-
-      # The following 2 rules are only needed for the user_webfinger app.
-      # Uncomment it if you're planning to use this app.
-      #rewrite ^/.well-known/host-meta /public.php?service=host-meta last;
-      #rewrite ^/.well-known/host-meta.json /public.php?service=host-meta-json last;
-
-      # The following rule is only needed for the Social app.
-      # Uncomment it if you're planning to use this app.
-      #rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
-
-      location = /.well-known/carddav {
-        return 301 $scheme://$host:$server_port/remote.php/dav;
-      }
-      location = /.well-known/caldav {
-        return 301 $scheme://$host:$server_port/remote.php/dav;
-      }
-
+      #add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload;" always;
+      
       # set max upload size
       client_max_body_size 512M;
       fastcgi_buffers 64 4K;
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
       gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
       gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
 
-      # Uncomment if your server is build with the ngx_pagespeed module
-      # This module is currently not supported.
+      # Pagespeed is not supported by NextCloud, so if your server is built
+      # with the `ngx_pagespeed` module, uncomment this line to disable it.
       #pagespeed off;
-
-      location / {
-          rewrite ^ /index.php;
+      
+      # HTTP response headers borrowed from NextCloud `.htaccess`
+      add_header Referrer-Policy                      "no-referrer"   always;
+      add_header X-Content-Type-Options               "nosniff"       always;
+      add_header X-Download-Options                   "noopen"        always;
+      add_header X-Frame-Options                      "SAMEORIGIN"    always;
+      add_header X-Permitted-Cross-Domain-Policies    "none"          always;
+      add_header X-Robots-Tag                         "none"          always;
+      add_header X-XSS-Protection                     "1; mode=block" always;
+      
+      # Remove X-Powered-By, which is an information leak
+      fastcgi_hide_header X-Powered-By;
+      
+      # Path to the root of your installation
+      root /var/www/nextcloud;
+      
+      # Specify how to handle directories -- specifying `/index.php$request_uri`
+      # here as the fallback means that Nginx always exhibits the desired behaviour
+      # when a client requests a path that corresponds to a directory that exists
+      # on the server. In particular, if that directory contains an index.php file,
+      # that file is correctly served; if it doesn't, then the request is passed to
+      # the front-end controller. This consistent behaviour means that we don't need
+      # to specify custom rules for certain paths (e.g. images and other assets,
+      # `/updater`, `/ocm-provider`, `/ocs-provider`), and thus
+      # `try_files $uri $uri/ /index.php$request_uri`
+      # always provides the desired behaviour.
+      index index.php index.html index.htm /index.php$request_uri;
+      
+      # Default Cache-Control policy
+      expires 1m;
+      
+      # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
+      if ( $http_user_agent ~ DavClnt ) { rewrite ^/$ /remote.php/webdav/ redirect; }
+      
+      location = /robots.txt {
+          allow all;
+          log_not_found off;
+          access_log off;
       }
-
-      location ~ ^\/(?:build|tests|config|lib|3rdparty|templates|data)\/ {
-          deny all;
+      
+      # Make a regex exception for `/.well-known` so that clients can still
+      # access it despite the existence of the regex rule `location ~ /\.`
+      # which would otherwise handle requests for `/.well-known`.
+      location ^~ /.well-known {
+          # The following 6 rules are borrowed from `.htaccess`
+      
+          rewrite ^/\.well-known/host-meta\.json  /public.php?service=host-meta-json  last;
+          rewrite ^/\.well-known/host-meta        /public.php?service=host-meta       last;
+          rewrite ^/\.well-known/webfinger        /public.php?service=webfinger       last;
+          rewrite ^/\.well-known/nodeinfo         /public.php?service=nodeinfo        last;
+          
+          location /.well-known/carddav   { return 301 /remote.php/dav/; }
+          location /.well-known/caldav    { return 301 /remote.php/dav/; }
       }
-      location ~ ^\/(?:\.|autotest|occ|issue|indie|db_|console) {
-          deny all;
-      }
-
-      location ~ ^\/(?:index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy)\.php(?:$|\/) {
-          fastcgi_split_path_info ^(.+?\.php)(\/.*|)$;
+      
+      # Rules borrowed from `.htaccess` to hide certain paths from clients
+      location ~ ^/(build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
+      location ~ ^/(autotest|occ|issue|indie|db_|console)                 { return 404; }
+      
+      # Respond with HTTP 404 to *all* requests for hidden files;
+      # not just hidden files that reside in the root directory.
+      location ~ /\. { return 404; }
+      
+      # Ensure this block, which passes PHP files to the PHP process, is above the blocks
+      # which handle static assets (as seen below). If this block is not declared first,
+      # then Nginx will encounter an infinite rewriting loop when it prepends `/index.php`
+      # to the URI, resulting in a HTTP 500 error response.
+      location ~ \.php($|/) {
+          fastcgi_split_path_info ^(.+?\.php)(/.*|)$;
           set $path_info $fastcgi_path_info;
+          
           try_files $fastcgi_script_name =404;
+          
           include fastcgi_params;
           fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
           fastcgi_param PATH_INFO $path_info;
           fastcgi_param HTTPS on;
-          # Avoid sending the security headers twice
-          fastcgi_param modHeadersAvailable true;
-          # Enable pretty urls
-          fastcgi_param front_controller_active true;
+          
+          fastcgi_param modHeadersAvailable true;         # Avoid sending the security headers twice
+          fastcgi_param front_controller_active true;     # Enable pretty urls
           fastcgi_pass php-handler;
+          
           fastcgi_intercept_errors on;
           fastcgi_request_buffering off;
       }
-
-      location ~ ^\/(?:updater|oc[ms]-provider)(?:$|\/) {
-          try_files $uri/ =404;
-          index index.php;
+      
+      location ~ \.(css|js|svg|gif)$ {
+          try_files $uri $uri/ /index.php$request_uri;
+          expires 6M;         # Cache-Control policy borrowed from `.htaccess`
+          access_log off;     # Optional: Don't log access to assets
       }
-
-      # Adding the cache control header for js, css and map files
-      # Make sure it is BELOW the PHP block
-      location ~ \.(?:css|js|woff2?|svg|gif|map)$ {
-          try_files $uri /index.php$request_uri;
-          add_header Cache-Control "public, max-age=15778463";
-          # Add headers to serve security related headers (It is intended to
-          # have those duplicated to the ones above)
-          # Before enabling Strict-Transport-Security headers please read into
-          # this topic first.
-          #add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload;" always;
-          #
-          # WARNING: Only add the preload option once you read about
-          # the consequences in https://hstspreload.org/. This option
-          # will add the domain to a hardcoded list that is shipped
-          # in all major browsers and getting removed from this list
-          # could take several months.
-          add_header Referrer-Policy "no-referrer" always;
-          add_header X-Content-Type-Options "nosniff" always;
-          add_header X-Download-Options "noopen" always;
-          add_header X-Frame-Options "SAMEORIGIN" always;
-          add_header X-Permitted-Cross-Domain-Policies "none" always;
-          add_header X-Robots-Tag "none" always;
-          add_header X-XSS-Protection "1; mode=block" always;
-
-          # Optional: Don't log access to assets
-          access_log off;
+      
+      location ~ \.woff2?$ {
+          try_files $uri $uri/ /index.php$request_uri;
+          expires 7d;         # Cache-Control policy borrowed from `.htaccess`
+          access_log off;     # Optional: Don't log access to assets
       }
-
-      location ~ \.(?:png|html|ttf|ico|jpg|jpeg|bcmap|mp4|webm|mp3|ogg|wav)$ {
-          try_files $uri /index.php$request_uri;
-          # Optional: Don't log access to other assets
-          access_log off;
+      
+      location / {
+          try_files $uri $uri/ /index.php$request_uri;
       }
   }
 
@@ -191,8 +187,14 @@ Nextcloud in a subdir of the nginx webroot
 
 The following config should be used when Nextcloud is placed within a subdir of
 the webroot of your nginx installation.
-In this example the webroot is located at
-``/var/www`` and the Nextcloud instance is accessed via ``http(s)://cloud.example.com/nextcloud/``
+In this example the NextCloud files are located at
+``/var/www/nextcloud`` and the Nextcloud instance is accessed via ``http(s)://cloud.example.com/nextcloud/``.
+The configuration differs from the "Nextcloud in webroot" configuration above in the following ways:
+
+- All requests for ``/nextcloud`` are encapsulated within a single ``location`` block, namely ``location ~ ^/nextcloud($|/)``.
+- The string ``/nextcloud`` is prepended to all prefix paths.
+- The URI ``/nextcloud`` is *aliased* to ``/var/www/nextcloud``, rather than the domain itself being *rooted* at ``/var/www/nextcloud``.
+- The blocks that handle requests for paths outside of ``/nextcloud``, i.e. ``/robots.txt``, ``/.well-known``, and hidden files, are pulled out of the ``location ~ ^/nextcloud($|/)`` block.
 
 .. code-block:: nginx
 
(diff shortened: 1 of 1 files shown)


MorrisJobke:
cc @josh4trunks

jivanpal:
Force-pushed to overwrite commit and make some minor spelling/formatting fixes

jivanpal:
Force-pushed to overwrite commit and make some minor spelling/formatting fixes

josh4trunks:
Thanks for looking at this @jivanpal! If possibly, please dont force overwrite the commit, I had reviewed the previous version of the file and now I can't see what was changed. Next time, just push another commit on top so we get the whole messy history, lol. No worries we all have typos and need to change things.

josh4trunks:
I agree with you that we should strive to match the **.htaccess** file as well as we can without sacrificing security/performance; that way we can more easily incorporate changes added to Nextcloud's codebase. I'll try adding my comments into the file, so its easier to know what I'm asking and you can respond when you have time.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
+      # `/updater`, `/ocm-provider`, `/ocs-provider`), and thus
+      # `try_files $uri $uri/ /index.php$request_uri`
+      # always provides the desired behaviour.
+      index index.php index.html index.htm /index.php$request_uri;
+      
+      # Default Cache-Control policy
+      expires 1m;
+      
+      # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
+      if ( $http_user_agent ~ DavClnt ) { rewrite ^/$ /remote.php/webdav/ redirect; }
```
josh4trunks, line 114: Not really a fan of "if" but not sure how else we can accomplish this.
The rewrite rule doesn't really rewrite anything, we might as well just `return 302 /remote.php/webdav`

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
           fastcgi_intercept_errors on;
           fastcgi_request_buffering off;
       }
-
-      location ~ ^\/(?:updater|oc[ms]-provider)(?:$|\/) {
-          try_files $uri/ =404;
-          index index.php;
+      
+      location ~ \.(css|js|svg|gif)$ {
+          try_files $uri $uri/ /index.php$request_uri;
```
josh4trunks, line 169: Do we need the $uri/ here? I don't think we ever need or want file.css to be treated like a folder

admin_manual/installation/nginx.rst
jivanpal: Yes, can't do conditional based on user agent without `if`. We use `rewrite` because we only want to redirect MS DAV client if it requests `/`, so we specify `^/$`, and because we want to maintain the query string, if any (`rewrite` automatically handles this). The only other way to implement this that I can see would be an additional `location` block as follows:

```
location = / {
    if ($http_user_agent ~ DavClnt) { return 302 /remote.php/webdav/$is_args$args; }

    try_files $uri $uri/ /index.php$request_uri;
}
```

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
+      location ^~ /.well-known {
+          # The following 6 rules are borrowed from `.htaccess`
+      
+          rewrite ^/\.well-known/host-meta\.json  /public.php?service=host-meta-json  last;
+          rewrite ^/\.well-known/host-meta        /public.php?service=host-meta       last;
+          rewrite ^/\.well-known/webfinger        /public.php?service=webfinger       last;
+          rewrite ^/\.well-known/nodeinfo         /public.php?service=nodeinfo        last;
+          
+          location /.well-known/carddav   { return 301 /remote.php/dav/; }
+          location /.well-known/caldav    { return 301 /remote.php/dav/; }
```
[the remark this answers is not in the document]
jivanpal: The Apache regex rules only seek to match the beginning of the URI, e.g. `^/\.well-known/carddav/`, so they match URIs such as `/.well-known/carddav/whatever`

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
       }
-
-      location ~ ^\/(?:index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy)\.php(?:$|\/) {
```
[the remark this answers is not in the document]
jivanpal: Yes, it would — I believe the Apache setup permits this too, but I haven't tested it.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
-          fastcgi_split_path_info ^(.+?\.php)(\/.*|)$;
+      
+      # Rules borrowed from `.htaccess` to hide certain paths from clients
+      location ~ ^/(build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
+      location ~ ^/(autotest|occ|issue|indie|db_|console)                 { return 404; }
+      
+      # Respond with HTTP 404 to *all* requests for hidden files;
+      # not just hidden files that reside in the root directory.
+      location ~ /\. { return 404; }
```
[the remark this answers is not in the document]
jivanpal: Fair point, I just assumed that clients never have a legitimate reason to request a URI corresponding to a hidden file — I'll test this

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
           fastcgi_request_buffering off;
       }
-
-      location ~ ^\/(?:updater|oc[ms]-provider)(?:$|\/) {
-          try_files $uri/ =404;
-          index index.php;
+      
+      location ~ \.(css|js|svg|gif)$ {
+          try_files $uri $uri/ /index.php$request_uri;
+          expires 6M;         # Cache-Control policy borrowed from `.htaccess`
```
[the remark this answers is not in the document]
jivanpal: Within this `location` block, we use `expires` rather than `add_header Cache-Control`, so to my knowledge this does not override any/all `add_header` directives in the outer `server` scope — I'll check this, though

admin_manual/installation/nginx.rst
jivanpal: Not needed, but IMO it simplifies the config's readability if we use the same `try_files` directive everywhere. It won't hurt to remove `$uri/` here.

admin_manual/installation/nginx.rst
jivanpal: Actually the `try_files` line in that alternative is unnecessary since we just serve `/index.php` in response to a request for `/`, which is already handled by `index` directive.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,109 @@ webroot of your nginx installation. In this example it is
...
+          location /.well-known/carddav   { return 301 /remote.php/dav/; }
+          location /.well-known/caldav    { return 301 /remote.php/dav/; }
       }
-
-      location ~ ^\/(?:index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy)\.php(?:$|\/) {
-          fastcgi_split_path_info ^(.+?\.php)(\/.*|)$;
+      
+      # Rules borrowed from `.htaccess` to hide certain paths from clients
+      location ~ ^/(build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
+      location ~ ^/(autotest|occ|issue|indie|db_|console)                 { return 404; }
```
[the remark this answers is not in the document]
jivanpal: `/.htaccess` uses the regex `^(?:build|tests|config|lib|3rdparty|templates)/.*`, which already includes `config`, so `/config/.htaccess` is already implemented by that regex. I added `data` to the end to implement `/data/.htaccess`.

Re non-capturing group `(?:`, it is my understanding that Nginx doesn't bother capturing a group to e.g. `$1` if `$1` does not appear anywhere within the `location` block, as is the case here — but I could be wrong about this. Even if Nginx does still internally capture the group, surely any performance gain is negligible?

I'm happy to let you decide whatever you prefer with regards to both of these points.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: In fact, perhaps the `.htaccess` rules need to be altered, because they also match URIs such as `/.well-known/carddavfoo`. Of course, if we know that DAV-compliant clients only ever make requests for `/.well-known/[card|cal]dav`, and not any sub-paths of these, then `location =` would be preferred.

Also, on the topic of `.htaccess` itself potentially needing to be altered, it currently reads:

```
  RewriteRule ^\.well-known/host-meta /public.php?service=host-meta [QSA,L]
  RewriteRule ^\.well-known/host-meta\.json /public.php?service=host-meta-json [QSA,L]
```

But in this order, a request for `/.well-known/host-meta.json` matches the first regex, so a redirect to `/public.php?service=host-meta` occurs rather than to `/public.php?service=host-meta-json`. Incidentally, this is why I have written the corresponding `rewrite` rules in my `location ^~ /.well-known` block in the opposite order.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: Tested now, and yes, this does break the ability to work with hidden files, so this should be removed and we should change the `autotest|occ|...` rule to reflect what is in `.htaccess`

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: NOTE: change `location ~ ^/(autotest|...` to `location ~ ^/(\.|autotest|...`

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: Tested this now, and it does behave as I expected; headers are served correctly, e.g. for `/apps/files_pdfviewer/css/style.css` used on the login page, I get:

```
cache-control: max-age=15552000
date: Sun, 12 Jul 2020 19:59:38 GMT
expires: Fri, 08 Jan 2021 19:59:38 GMT
last-modified: Tue, 26 May 2020 08:09:01 GMT
referrer-policy: no-referrer
x-content-type-options: nosniff
x-download-options: noopen
x-frame-options: SAMEORIGIN
x-permitted-cross-domain-policies: none
x-robots-tag: none
x-xss-protection: 1; mode=block
```

Of course, for requests that are handled by the front-end controller, such as `/css/core/xxxx-xxxx-server.css` used after logging in, the default expiration policy of 1 minute specified in the `server` block applies, instead of 6 months.

jivanpal:
@josh4trunks I've made some changes to address your comments, see commit message for details.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: I am fairly certain NGINX does capture text in parenthesis in a regex. If you search for the word "capture" on this page you can see an example. But feel free to test it.
http://nginx.org/en/docs/http/ngx_http_core_module.html

I think we should make it non-capturing because...
* There is no need to capture the blocked directory/file
* It is (negligibly) better performance
* It matches the `.htaccess`

I did not see "config" there, wasn't reading closely enough. I agree then, your order is best.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: In this case I think what we should do is read up on the RFC6764 and just verify what URLs NC should be responding too. If `.htaccess` isn't doing it correctly ideally we should fix it there not try to match their quirks.

Feel free to do some reading here, or I can try looking at sometime soon.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: If your right, then it totally makes sense to match the Apache/normal method of executing all PHP files.
You mind giving me a few days to test on Apache, I just want to be sure I'm understanding everything correctly.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: Sounds good, I wonder why we had to make this so complicated before then, lol. I didn't do it....
Again, if you could give me a few days to test your config, I would appreciate that.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: I would also be interested in knowing what NGINX does different between `deny` and `return 404`. Not critical we figure that out now, just noting it down for future reference.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,110 @@ webroot of your nginx installation. In this example it is
...
+      # access it despite the existence of the regex rule
+      # `location ~ /(\.|autotest|...)` which would otherwise handle requests
+      # for `/.well-known`.
+      location ^~ /.well-known {
+          # The following 6 rules are borrowed from `.htaccess`
+      
+          rewrite ^/\.well-known/host-meta\.json  /public.php?service=host-meta-json  last;
+          rewrite ^/\.well-known/host-meta        /public.php?service=host-meta       last;
+          rewrite ^/\.well-known/webfinger        /public.php?service=webfinger       last;
+          rewrite ^/\.well-known/nodeinfo         /public.php?service=nodeinfo        last;
```
josh4trunks, line 136: Not critical we figure this out now, but I wonder if these could be "=" locations, with return statements?

admin_manual/installation/nginx.rst
josh4trunks: I like your change here, I think could be improved even more if we make the string compare an exact match. This wouldn't match the `.htaccess`, but I think it is the more correct behavior.

admin_manual/installation/nginx.rst
josh4trunks: It might improve the consistency of the config, but if you actually read the config logically it doesn't actually make sense.
Closing this since you already implemented the change I recommended.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: `deny` returns a 403 🙂

admin_manual/installation/nginx.rst
jivanpal: I think we want these to be internal redirects, rather than responding to the client with 301/302l if so, then I think `rewrite` is the only way to go.

Even if we're okay responding with 301/302, we want to preserve the query string and add the `service=` parameter, so using `rewrite ... redirect` is arguably nicer than `return [301]` then, since it effectively implements Apache's `RewriteRule ... [QSA]` behaviour. Else, we'd need something like:

```
location = /.well-known/host-meta.json {
    return /public.php?service=host-meta-json&$args;
}
```

This has the property that if `$args` is empty, there is a trailing `&` in the redirect — I don't know if this conforms to RFCs regarding URLs/URIs; if it does not, we cannot necessarily guarantee that clients will handle this correctly (though I expect that this isn't an issue in practice).

admin_manual/installation/nginx.rst
jivanpal: Regarding `=` rather than `~`, there are MS DAV clients whose user agent string is not precisely `DavClnt`, e.g. `DavClnt$Mozilla/5.0 (Windows NT 10.0; Win64; x64; Trident/7.0; rv:11.0) like Gecko$SynapseNotifyAgent` — [see here](https://user-agents.net/lookup), search for `davclnt`.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: > I am fairly certain NGINX does capture text in parenthesis in a regex.

It does, but I believe that it is intelligent enough to render capturing groups as non-capturing groups if no captures are used in the block, i.e. no instances of `$1, `$2`, etc. appear. That is, when the config is loaded, Nginx will essentially rewrite the group internally to use `(?:` rather than `(`. I'm not completely sure of this (can't find any documentation on it now, though I'm pretty sure I read it somewhere in the past), and no harm in using non-capturing, so I'll change it

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
jivanpal: Had a read over the RFC, and, in conjunction with [RFC5785 §5.1](https://tools.ietf.org/html/rfc5785#section-5.1), it registers the URI suffixes `caldav` and `carddav` ([see §9.1](https://tools.ietf.org/html/rfc6764#section-9.1)). This means that the URIs `/.well-known/caldav` and `/.well-known/carddav`, as well as sub-URIs (e.g. `/.well-known/caldav/foo`, but not `/.well-known/caldavfoo`), are reserved for use by CalDAV/CardDAV. This is also reflected in [IANA's list of well-known URIs](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml).

However, only usage of the exact URIs `/.well-known/caldav` and `/.well-known/carddav — but not any sub-URIs such as `/.well-known/caldav/foo` — appears to be defined ([see §5](https://tools.ietf.org/html/rfc6764#section-5)). As such, it seems that the Apache rules ought to be as follows:

```
  RewriteRule ^\.well-known/carddav$ /remote.php/dav/ [R=301,L]
  RewriteRule ^\.well-known/caldav$ /remote.php/dav/ [R=301,L]
```

Or at the very least, so as not to match e.g. `/.well-known/caldavfoo`:

```
  RewriteRule ^\.well-known/carddav($|/) /remote.php/dav/ [R=301,L]
  RewriteRule ^\.well-known/caldav($|/) /remote.php/dav/ [R=301,L]
```

Thus, I think we should use `location =`.

josh4trunks:
If you don't mind giving me a few more day, I still want to test and propose any changes I think need to be made.
Haven't found much time this week, am moving to a new home so I am busy with that, but hopefully I can find some time by early next week.

jivanpal:
No rush! I can only imagine how cumbersome moving house is in the middle of a pandemic 😅

josh4trunks:
Haven't forgot about this, still on my list, just recovering from moving with a baby and dog, lol. I got my test server plugged in, and should have some time to look at this tomorrow night and will comment.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,110 @@ webroot of your nginx installation. In this example it is
       gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
       gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
 
-      # Uncomment if your server is build with the ngx_pagespeed module
-      # This module is currently not supported.
+      # Pagespeed is not supported by NextCloud, so if your server is built
```
splitt3r, line 79:
```suggestion
      # Pagespeed is not supported by Nextcloud, so if your server is built
```

Nextcloud

jivanpal:
@kesselb, done 😉

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: That optimization makes sense to me based on how NGINX renders configs, but as you said since it's not documented it makes some sense to use the non-capturing version, that matches apache. I think this still isn't changed (back) in your current (as of now) version, do you mind changing it so I can 'resolve' this conversation? Thanks!

So sorry about the delay reviewing. I didn't have my test server setup at my new home. Finally just spun up a VM, and hope to get through the remaining notes ASAP.

admin_manual/installation/nginx.rst
josh4trunks: Makes sense. Since they both start with DavClnt can we do this instead?
```if ($http_user_agent ~ ^DavClnt)```
If you agree please change and I can resolve. Thanks

File: admin_manual/installation/nginx.rst [hunk 10]
```diff
@@ -205,71 +209,52 @@ In this example the webroot is located at
...
-      location /.well-known/acme-challenge { }
-
-      location ^~ /nextcloud {
+          location /.well-known/carddav   { return 301 /nextcloud/remote.php/dav/; }
+          location /.well-known/caldav    { return 301 /nextcloud/remote.php/dav/; }
 
+          try_files $uri $uri/ /nextcloud/index.php$request_uri;
+      }
+      
+      location ~ ^/nextcloud($|/) {
```
josh4trunks, line 257: Sorry for not commenting on this previously, I didn't notice it. Why did you change this to a regex match instead of using the `^~` modifier?

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: closing since I made the changes already

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,110 @@ webroot of your nginx installation. In this example it is
...
+      location ~ \.woff2?$ {
           try_files $uri /index.php$request_uri;
-          # Optional: Don't log access to other assets
-          access_log off;
+          expires 7d;         # Cache-Control policy borrowed from `.htaccess`
+          access_log off;     # Optional: Don't log access to assets
+      }
+      
+      location / {
+          try_files $uri $uri/ /index.php$request_uri;
```
josh4trunks, line 182: Why does this need its own location, can it just live above or below the `index` parameter?

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +76,110 @@ webroot of your nginx installation. In this example it is
...
+          
+          location /.well-known/carddav   { return 301 /remote.php/dav/; }
+          location /.well-known/caldav    { return 301 /remote.php/dav/; }
       }
-
-      location ~ ^\/(?:index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy)\.php(?:$|\/) {
-          fastcgi_split_path_info ^(.+?\.php)(\/.*|)$;
+      
+      # Rules borrowed from `.htaccess` to hide certain paths from clients
+      location ~ ^/(?:build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
```
josh4trunks, line 143: Again, sorry for not noticing another change. Why is the second group needed here?
In what case does the original version not work?
`location ~ ^/(?:forbidden_folder1|forbidden_folder2)/`

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: Thanks for taking a read through! If you want to make an PR for the **.htaccess** for this/ or any other changed you feel are needed feel free to CC me and I can back you up.

I'll go ahead and implement the `location =` and resolve this comment.

admin_manual/installation/nginx.rst
```diff
@@ -205,71 +209,52 @@ In this example the webroot is located at
...
-        return 301 $scheme://$host:$server_port/nextcloud/remote.php/dav;
-      }
-
-      location /.well-known/acme-challenge { }
-
-      location ^~ /nextcloud {
+          location = /.well-known/carddav   { return 301 /nextcloud/remote.php/dav/; }
+          location = /.well-known/caldav    { return 301 /nextcloud/remote.php/dav/; }
 
+          try_files $uri $uri/ /nextcloud/index.php$request_uri;
```
josh4trunks, line 254: What is this for? If a client is requesting a URL path not under /nextcloud, or another Nextcloud related path like .well-known/caldav, this might not even be for Nextcloud.
For example, in my server config I have different /subdirs for different PHP application. Maybe we should just let NGINX serve a 404 instead of leading any request to Nextcloud.

admin_manual/installation/nginx.rst
josh4trunks: Thanks for the thoroughly explanation! I agree internal redirect is best.

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: Confirmed what you found. `expires` works as expected, unlike `add header`. I'll go ahead and remove the note that is now irrelevant.

josh4trunks:
@jivanpal, I plan on testing an Apache configuration in the next few days and after that can resolve this comment https://github.com/nextcloud/documentation/pull/2197#discussion_r453256288.

Do you mind responding to my 4 additional comments posted today? I think 3 of them can be settled very easily. [This comment](https://github.com/nextcloud/documentation/pull/2197#discussion_r466234520) may take a little more back and forth to reach agreement since it affects more lines, but I'll do my best to respond quickly.

Thank you

admin_manual/installation/nginx.rst
jivanpal: Woops, my recollection of the location nesting rules must've been wrong when I wrote this, will change this to `location ^~ /nextcloud`.

admin_manual/installation/nginx.rst
jivanpal: I was under the impression that NextCloud plugins/apps might want to use other well-known URIs, and the front-end controller would handle those. If you know this not to be the case, or still don't want to fallback to NextCloud controller, we can remove this. It is probably the right thing to do.

admin_manual/installation/nginx.rst
jivanpal: e.g. `/build/` would return 404, but if a request is made for `/build` (no trailing slash), we wouldn't return 404, but instead pass the request to the front-end controller. Using `($|/)` rather than `/` handles that.

admin_manual/installation/nginx.rst
jivanpal: After some googling, I found that it doesn't, but the behaviour of `try_files` in this regard doesn't appear to be documented, and in my experience, it seems to be standard practice to not use `try_files` in `server` scope if there are `location` blocks. IMO, it eases readability and avoids confusion. Even if we take it out of the `location` block, I personally would leave it after the last location block, not place it directly after `index`.

admin_manual/installation/nginx.rst
jivanpal: We should also probably use the same line in the "Nextcloud in webroot" config, i.e. `try_files $uri $uri/ =404`.

admin_manual/installation/nginx.rst
```diff
@@ -140,15 +140,15 @@ webroot of your nginx installation. In this example it is
...
+      location ~ ^/(?:build|tests|config|lib|3rdparty|templates|data)($|/)  { return 404; }
+      location ~ ^/(?:\.|autotest|occ|issue|indie|db_|console)              { return 404; }
       
       # Ensure this block, which passes PHP files to the PHP process, is above the blocks
       # which handle static assets (as seen below). If this block is not declared first,
       # then Nginx will encounter an infinite rewriting loop when it prepends `/index.php`
       # to the URI, resulting in a HTTP 500 error response.
       location ~ \.php($|/) {
-          fastcgi_split_path_info ^(.+?\.php)(/.*|)$;
+          fastcgi_split_path_info ^(.+?\.php)(/.*)$;
```
jivanpal, line 146: Are we sure that `(/.*)` rather than `(/.*|)` correctly sets `$fastcgi_path_info` for a URI such as `index.php` (no trailing slash), wherein `(/.*)` doesn't match anything?

jivanpal: This appears on lines 146 and 316.

josh4trunks: I am fairly certain (but feel free to test). This is how the configuration used to be, until it was changed without a proper discussion [here](https://github.com/nextcloud/documentation/pull/1152).

In the case of `index.php`, `fastcgi_split_path_info` will not match and `$fastcgi_path_info` will not be set, which seems like the appropriate behavior to me.

admin_manual/installation/nginx.rst
josh4trunks: Ok, the main reason I  think it would be good to not nest this fallback try_files in a location is because the `location = /` we made earlier wont inherit this. But we could easily fix this by nesting that `location =` in this fallback location.

admin_manual/installation/nginx.rst
josh4trunks: Ok makes sense, sorry for making the config more ugly but gonna make these non-capturing =P

admin_manual/installation/nginx.rst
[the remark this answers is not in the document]
josh4trunks: Yup Apache lets any valid PHP run. Glad we could simplify this!
I swear I didn't want to make this complicated when I first wrote it, I just was copying the `.htaccess` at the time.

admin_manual/installation/nginx.rst
jivanpal: A `try_files` directive in `server` scope is not inherited by any child `location` blocks, anyway. Instead, it is used iff the URI does not match any `location` blocks. We don't need a `try_files` within `location = /`, because the `index` directive results in `/index.php` being served (if the user agent doesn't match `^DavClnt`).

The following are all behaviourally equivalent:

```nginx
server {
    index index.php index.html /index.php$request_uri;

    location = / {
        if ( $http_user_agent ~ ^DavClnt ) {
            return 302 /remote.php/webdav/$is_args$args;
        }
    }

    location / {
        try_files $uri $uri/ /index.php$request_uri;
    }
}
```

```nginx
server {
    index index.php index.html /index.php$request_uri;

    location = / {
        if ( $http_user_agent ~ ^DavClnt ) {
            return 302 /remote.php/webdav/$is_args$args;
        }
    }

    try_files $uri $uri/ /index.php$request_uri;
}
```

```nginx
server {
    index index.php index.html /index.php$request_uri;

    try_files $uri $uri/ /index.php$request_uri;
    
    location = / {
        if ( $http_user_agent ~ ^DavClnt ) {
            return 302 /remote.php/webdav/$is_args$args;
        }
    }
}
```

josh4trunks: Good point! I did not think to realize `try_files` is irrelevant with `location =`

admin_manual/installation/nginx.rst
josh4trunks: If you're ok with my change/explanation here then it looks like everything is resolved and I am all good giving a "thumbs up"

jivanpal: I will be doing final tests now on a fresh install just to check this and whether there's any obvious issues.

kesselb: https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/

kesselb: > https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/

Nevermind. I mixed it up.

josh4trunks: > I will be doing final tests now on a fresh install just to check this and whether there's any obvious issues.

Ok great. I'm modified my production configuration and am basically using what we have here, so there's that datapoint.

admin_manual/installation/nginx.rst
```diff
@@ -282,75 +262,89 @@ In this example the webroot is located at
...
+          add_header X-Permitted-Cross-Domain-Policies    "none"          always;
+          add_header X-Robots-Tag                         "none"          always;
+          add_header X-XSS-Protection                     "1; mode=block" always;
+          
+          # Remove X-Powered-By, which is an information leak
+          fastcgi_hide_header X-Powered-By;
+          
+          # Path to the root of your installation
+          # Notice that this is an `alias` directive, not a `root` directive
+          alias /var/www/nextcloud;
```
jivanpal, line 283: I've tested the subdir setup, and it turns out that using `alias` is problematic. We can either:

1. Switch back to using `root`, e.g. `root /var/www` like in the currently published config, but this means the name of the Nextcloud root directory must match that of the URI, e.g. `nextcloud`, which may be undesireable; or

2. Keep `alias`, and instead:

  a. change `fastcgi_param SCRIPT_FILENAME` from `$document_root$fastcgi_script_name` to `$request_filename` in the `~\.php$` block; and

  b. change all instances of `/nextcloud/index.php$request_uri` to just `/nextcloud/index.php, and add `fastcgi_param REQUEST_URI $request_uri` instead. Strangely, I faced issues with Nextcloud not handling requests properly when using the former method, but the latter works fine.

I think we should go with (1) for simplicity, but the behaviour I've seen regarding `$request_uri` suggests that perhaps we should be using `fastcgi_param REQUEST_URI $request_uri` in both configs, rather than appending `$request_uri` to `/index.php`.

Thoughts?

josh4trunks: I think ideally we should go to a model that uses `alias`, since some users will want to host their application using a different webdir (like **nc** or **mycloud**). But for now, if you just want to switch back to `root` that should be fine since that is how it was. That will allow us to approve these change and not hold it up for something somewhat unrelated.

Then if you or I or someone else wants to tackle using `alias` we can just create a new PR.

jivanpal: Changes made, feel free to review now

josh4trunks: Everything looks fine to me, my production system is basically using this (subdir using `root`).
Since you started these 2 comments I'll let you resolve them, https://github.com/nextcloud/documentation/pull/2197#pullrequestreview-462693331 https://github.com/nextcloud/documentation/pull/2197#pullrequestreview-463714625.

But otherwise I think these changed are all good.

josh4trunks:
@MorrisJobke this PR looks good to me. @jivanpal put a ton of work into it! :+1:
We should probably remove my name from the docs eventually, lol.

MorrisJobke:
Thanks a lot @jivanpal and @josh4trunks for your help with this :)

juliushaertl:
The new config is actually having issues with the legacy ajax endpoints of the user_ldap app.

The URL https://cloud.example.net/apps/user_ldap/ajax/getConfiguration.php is not going through index.php but being called directly. @josh4trunks @jivanpal Any idea how we could resolve that?

juliushaertl:
I'd probably just revert this line: https://github.com/nextcloud/documentation/pull/2197/files#diff-490ed055ec68f0382edb4b4d680b761bL132

josh4trunks:
@juliushaertl can you confirm if this is also a problem with an Apache server? Because we are only copying what the `.htaccess` would do, which if presented with that URL would directly execute the PHP file, not route it through `index.php`.

Do you know of any other PHP files that need routing through index.php? If this is not common, and most files can/should be executed directly then this should be fixed in `user_ldap`'s code.

juliushaertl:
> @juliushaertl can you confirm if this is also a problem with an Apache server? Because we are only copying what the .htaccess would do, which if presented with that URL would directly execute the PHP file, not route it through index.php.

No, this works fine with the current `.htaccess`, it would basically pass all requests to index if they are not in the exclude list of:

```
  RewriteRule ^core/js/oc.js$ index.php [PT,E=PATH_INFO:$1]
  RewriteRule ^core/preview.png$ index.php [PT,E=PATH_INFO:$1]
  RewriteCond %{REQUEST_FILENAME} !\.(css|js|svg|gif|png|html|ttf|woff2?|ico|jpg|jpeg|map|webm|mp4|mp3|ogg|wav)$
  RewriteCond %{REQUEST_FILENAME} !core/img/favicon.ico$
  RewriteCond %{REQUEST_FILENAME} !core/img/manifest.json$
  RewriteCond %{REQUEST_FILENAME} !/remote.php
  RewriteCond %{REQUEST_FILENAME} !/public.php
  RewriteCond %{REQUEST_FILENAME} !/cron.php
  RewriteCond %{REQUEST_FILENAME} !/core/ajax/update.php
  RewriteCond %{REQUEST_FILENAME} !/status.php
  RewriteCond %{REQUEST_FILENAME} !/ocs/v1.php
  RewriteCond %{REQUEST_FILENAME} !/ocs/v2.php
  RewriteCond %{REQUEST_FILENAME} !/robots.txt
  RewriteCond %{REQUEST_FILENAME} !/updater/
  RewriteCond %{REQUEST_FILENAME} !/ocs-provider/
  RewriteCond %{REQUEST_FILENAME} !/ocm-provider/
  RewriteCond %{REQUEST_URI} !^/\.well-known/(acme-challenge|pki-validation)/.*
  RewriteCond %{REQUEST_FILENAME} !/richdocumentscode(_arm64)?/proxy.php$
  RewriteRule . index.php [PT,E=PATH_INFO:$1]
  RewriteBase /
```

I currently don't have a nginx setup otherwise I'd see to come up with a fix myself, it was just something i was when looking at instance yesterday.

> Do you know of any other PHP files that need routing through index.php? If this is not common, and most files can/should be executed directly then this should be fixed in user_ldap's code.

Ideally yes, though this is nothing we can do for 20 anymore, so we should for now get the webserver config to some working state.

josh4trunks:
Wow, I'm surprised it worked for @jivanpal and I in production. So the need for routing through `index.php` is fairly limited.

But yes, I agree that we should revert the parts of the config that break normal operation for now! Until the day where all PHP files can be executed directly (in which case the `.htaccess` can be simplified as well).

We for sure need to revert the PHP location line you mentioned, possibly others too. But maybe we can just do a PR for that for now, and test the functions that are being effected.

josh4trunks:
How can I verify if `getConfiguration.php` is working? Do I need to activate the user_ldap app?

juliushaertl:
I just found  `/apps/files/ajax/list.php` and it doesn't seem to be used anywhere anymore though it is still there and used the same logic, so you could probably just test with that.

josh4trunks:
I think what we need to do is the following steps for both the regular and `/nextcloud` example (I'm probably missing stuff like the "/updater" folder so feel free to address). If someones wants to make a PR maybe @jivanpal and I can review/test.
Hopefully in the future we can simplify this configuration and Setup.php (which generates a modified .htaccess) by addressing these problem PHP files.

1) Change the `location ~ \.php(?:$|/) {` line to...
```
location ~ ^/(?:index|remote|public|cron|core/ajax/update|status|ocs/v[12]|updater/.+|oc[ms]-provider/.+|richdocumentscode/proxy)\.php(?:$|/) {
```
2) Add...
```
location ~ \.(png|html|ttf|ico|jpg|jpeg|map|webm|mp4|mp3|ogg|wav)$ {
    try_files $uri /owncloud/index.php$request_uri;
}
location = /core/img/manifest.json { }
```
3) Replace `try_files $uri $uri/ /index.php$request_uri;` with...
```
rewrite ^ /owncloud/index.php;
```

jivanpal:
I will probably take a proper look at this and test things today and tomorrow whilst I have the time. @juliushaertl, is that set of `RewriteCond`s new? I don't recall seeing it when my PR was pending. @josh4trunks's suggested changes seem like they'd do the job.

josh4trunks:
They are in `lib/private/Setup.php` and I believe are amended to .htaccess assuming the directory is writable.
I knew about them when I originally wrote the nginx config, but figured they were no longer relevant/gone since things seemed to be working for me. At some point I think ownCloud (at the time) wouldn't even work with nginx unless you routed everything through index.php (except the files listed).

Agree on the arm64 addition, just want to match the functionality of .htaccess.

jlehtoranta:
To me it looks like the only issue with the new Nginx config in Nextcloud 20 are the php files, which need routing via index.php. So, I would just add a one liner rewrite rule for those directly into the php location block.
E.g.
```Nginx
...
location ~ \.php(?:$|/) {
    # Required for legacy support
    rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;

    fastcgi_split_path_info ^(.+?\.php)(/.*)$;
    set $path_info $fastcgi_path_info;

    try_files $fastcgi_script_name =404;
...
```
This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.

josh4trunks:
@jlehtoranta I like that solution since it keeps the progress we made while getting everything to work!
Busy at the moment, my second son is due this month and trying to sell my old home, but hopefully we can get someone to confirm this fixes any problem files.

jlehtoranta:
Thanks! I can confirm that this fixes the user_ldap app (nextcloud/server#16194), since it was why I stumbled upon this in the first place. Anyway, it would definitely be good to have some other people test this with their installations too.

corsac-s:
> To me it looks like the only issue with the new Nginx config in Nextcloud 20 are the php files, which need routing via index.php. So, I would just add a one liner rewrite rule for those directly into the php location block.
> E.g.
> 
> ```nginx
> ...
> location ~ \.php(?:$|/) {
>     # Required for legacy support
>     rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;
> 
>     fastcgi_split_path_info ^(.+?\.php)(/.*)$;
>     set $path_info $fastcgi_path_info;
> 
>     try_files $fastcgi_script_name =404;
> ...
> ```
> 
> This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.

Hi,
I have the `user_ldap` greyed out issue, but it seems that using the rewrite line directly in my nginx setup breaks everything (I get a 404 error on every page). Maybe I'd have to adjust the paths (the location ~ \.php line is inside a location ^~ /nextcloud), but I'm unsure how.

admin_manual/installation/nginx.rst
```diff
@@ -114,75 +69,112 @@ webroot of your nginx installation. In this example it is
...
+      # that file is correctly served; if it doesn't, then the request is passed to
+      # the front-end controller. This consistent behaviour means that we don't need
+      # to specify custom rules for certain paths (e.g. images and other assets,
+      # `/updater`, `/ocm-provider`, `/ocs-provider`), and thus
+      # `try_files $uri $uri/ /index.php$request_uri`
+      # always provides the desired behaviour.
+      index index.php index.html /index.php$request_uri;
+      
+      # Default Cache-Control policy
+      expires 1m;
```
anoymouserver, line 104: This seems to cause update delays for dynamic elements requested via APIs.
For example when using the news plugin, it causes feed items which I've marked as read still being displayed in the list for 60 seconds. Probably it also affects other elements and plugins as well.

jivanpal:
@josh4trunks, you said:

> They are in lib/private/Setup.php and I believe are amended to .htaccess assuming the directory is writable. [...] At some point I think ownCloud (at the time) wouldn't even work with nginx unless you routed everything through index.php (except the files listed).

Regarding #5472, if the behaviour in Apache is to not cache, then I'm guessing that it is due to this app using its own `.htaccess`. Is this a common pattern in Nextcloud? If so, I agree with you saying that Nginx support will require routing everything through `/index.php`, unless we can standardise an Nginx include file, e.g. at `/nginx-include.txt`, that one can include in the main config, and that Nexcloud/Nextcloud apps can modify in the way that they currently use `.htaccess` files. Of course, this would require reloading Nginx after every config change. Proper Nginx support without this sort of hacky workaround would either require apps to not rely on modifying `.htaccess` rules, or rely on routing everything through `/index.php`.

Regarding the specific issue with `Cache-Control` headers there, do you know if Nginx overrides HTTP headers set by PHP? If so, we probably do need to get rid of the `expires` mentioned by @anoymouserver and instead rely on PHP to set the header, or do some sort of `if` on `$http_cache_control`.

schleyk:
> 
> 
> To me it looks like the only issue with the new Nginx config in Nextcloud 20 are the php files, which need routing via index.php. So, I would just add a one liner rewrite rule for those directly into the php location block.
> E.g.
> 
> ```nginx
> ...
> location ~ \.php(?:$|/) {
>     # Required for legacy support
>     rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;
> 
>     fastcgi_split_path_info ^(.+?\.php)(/.*)$;
>     set $path_info $fastcgi_path_info;
> 
>     try_files $fastcgi_script_name =404;
> ...
> ```
> 
> This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.

your rewrite rule fix my issue with the ldap app.

jolly-jump:
> > To me it looks like the only issue with the new Nginx config in Nextcloud 20 are the php files, which need routing via index.php. So, I would just add a one liner rewrite rule for those directly into the php location block.
> > E.g.
> > ```nginx
> > ...
> > location ~ \.php(?:$|/) {
> >     # Required for legacy support
> >     rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;
> > 
> >     fastcgi_split_path_info ^(.+?\.php)(/.*)$;
> >     set $path_info $fastcgi_path_info;
> > 
> >     try_files $fastcgi_script_name =404;
> > ...
> > ```
> > 
> > 
> > This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.
> 
> your rewrite rule fix my issue with the ldap app.

I used this in my NC 20 which was upgraded from way below and then migrated to a new server where I initially put in the nginx-config that was given in the docs (two days ago), which did not work for LDAP

AND did not work for the rainloop addon!

Thanks for the fix. I'll try and tell the rainloop people as well.

jolly-jump:
cross-referencing rainloop which suffers from the same issue as LDAP-config here
https://github.com/pierre-alain-b/rainloop-nextcloud/issues/215

etique57:
> Also my nginx setup is itself behind a reverse proxy so I'm not sure if rewrite will work.

Hello @corsac-s  I'm in the same configuration as you (/nextcloud/ path, reverse-proxy in front) and observes the same effect of the fix as you (general 404).

Did you find a solution?

corsac-s:
> > Also my nginx setup is itself behind a reverse proxy so I'm not sure if rewrite will work.
> 
> Hello @corsac-s I'm in the same configuration as you (/nextcloud/ path, reverse-proxy in front) and observes the same effect of the fix as you (general 404).
> 
> Did you find a solution?

I kind of lost track of what I did and didn't do.

In the end, on my Apache reverse proxy, I use (using Unix sockets but I guess it should be the same with network):

````
ProxyPass / unix:/srv/cloud/cloud.sock|http://localhost/nextcloud/
ProxyPassReverse / unix:/srv/cloud/cloud.sock|http://localhost/nextcloud/
````
Then on the nginx part, I use:

````
 location ^~ /nextcloud {
                # a whole lot of stuff there, then
                
                # Rules borrowed from `.htaccess` to hide certain paths from clients
                location ~ ^/nextcloud/(?:build|tests|config|lib|3rdparty|templates|data)(?:$|/)    { return 404; }
                location ~ ^/nextcloud/(?:\.|autotest|occ|issue|indie|db_|console)                { return 404; }

                # Ensure this block, which passes PHP files to the PHP process, is above the blocks
                # which handle static assets (as seen below). If this block is not declared first,
                # then Nginx will encounter an infinite rewriting loop when it prepends
                # `/nextcloud/index.php` to the URI, resulting in a HTTP 500 error response.
                location ~ \.php(?:$|/) {
                        fastcgi_split_path_info ^(.+?\.php)(/.*)$;
                        set $path_info $fastcgi_path_info;

                        try_files $fastcgi_script_name =404;

                        include fastcgi_params;
                        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
                        fastcgi_param PATH_INFO $path_info;
                        fastcgi_param HTTPS on;

                        fastcgi_param modHeadersAvailable true;         # Avoid sending the security headers twice
                        fastcgi_param front_controller_active true;     # Enable pretty urls
                        fastcgi_pass php-handler;

                        fastcgi_intercept_errors on;
                        fastcgi_request_buffering off;
                }
}
````

Hope that helps.

**Comment by jivanpal** [comment]
@etique57 @corsac-s — The solution should be to change the rewrite rule to account for the subpath/subdir you're using, e.g. if your site is at `www.example.com/nextcloud`, then change `rewrite ^/(?!index...` to `rewrite ^/nextcloud/(?!index...`.

I am not sure whether this warrants a PR to modify the default Nginx config, given that the behaviour of the LDAP app relies on modifying Apache `.htaccess` files in situ. IMO, designing apps so that they don't rely on that would be better. Alternatively, allowing apps to specify Nginx rules in a file to be included via the main config might be acceptable, but that would then rely on app maintainers to provide the rules they need for both Apache *and* Nginx.

**Comment by etique57** [comment]
Thanks to both of you. I managed to get it working!

**Comment by ChristophWurst** [comment]
> This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.

Thanks a lot @jlehtoranta. I could verify that the current config doesn't work with the LDAP wizard and that your config revives the legacy routes. Could you open this as pull request? Otherwise I will do so if nobody sees any other issues with this config. I suppose those who used this trick still use the config in production, right? :grin:

**Comment by ChristophWurst** [comment]
PR submit at https://github.com/nextcloud/documentation/pull/7141.

**Comment by djalfy** [comment]
Hello.

This is probably a stupid question but where exactly should i place this in my nginx configuration?
```
`...
location ~ \.php(?:$|/) {
    # Required for legacy support
    rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;

    fastcgi_split_path_info ^(.+?\.php)(/.*)$;
    set $path_info $fastcgi_path_info;

    try_files $fastcgi_script_name =404;
...
`
```
I am running nextcloud in an unraid docker container and i could not find the right configuration file to place that fix. I either broke the webui entirely or if the webui worked, LDAP still didn't allow me to do anything in the setup wizard page.

Thanks!

KoMa1012:
> 



> > > To me it looks like the only issue with the new Nginx config in Nextcloud 20 are the php files, which need routing via index.php. So, I would just add a one liner rewrite rule for those directly into the php location block.
> > > E.g.
> > > ```nginx
> > > ...
> > > location ~ \.php(?:$|/) {
> > >     # Required for legacy support
> > >     rewrite ^/(?!index|remote|public|cron|core\/ajax\/update|status|ocs\/v[12]|updater\/.+|oc[ms]-provider\/.+|.+\/richdocumentscode\/proxy) /index.php$request_uri;
> > > 
> > >     fastcgi_split_path_info ^(.+?\.php)(/.*)$;
> > >     set $path_info $fastcgi_path_info;
> > > 
> > >     try_files $fastcgi_script_name =404;
> > > ...
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This approach seems to work at least in my use case, and it can also be very simply reverted, when the legacy support isn't required anymore.
> > 
> > 
> > your rewrite rule fix my issue with the ldap app.
> 
> I used this in my NC 20 which was upgraded from way below and then migrated to a new server where I initially put in the nginx-config that was given in the docs (two days ago), which did not work for LDAP
> 
> AND did not work for the rainloop addon!
> 
> Thanks for the fix. I'll try and tell the rainloop people as well.

this does not work for me, nextcloud 23 I get:
```nginx: [emerg] duplicate upstream "php-handler" in /config/nginx/site-confs/default.orig:1```

PR #2200, opened 2020-07-14
fixed tar.bz2 extraction command

Rizwan-Hasan:


skjnldsv:
```bash
$ tar --help | grep j   
  -j, --bzip2                filter the archive through bzip2
```

I think this needs to stay

Rizwan-Hasan:
> ```shell
> $ tar --help | grep bzip   
>   -j, --bzip2                filter the archive through bzip2
> ```
> 
> I think this needs to stay
> cc @MorrisJobke ?

### Yeah. You're right. 
### Again I've checked it and the result is "-xjvf" or "-xfv" works. But don't know why "-xjfv" doesn't. I've attached a screenshot from my Ubuntu 20.04.

Screenshot from 2020-07-14 17-34-57

skjnldsv:
Maybe report it to the tar repository? :)
Thanks for your interest in Nextcloud! :rocket:

PR #2204, opened 2020-07-17
Mentions `app:remove`

aleksandar-todorovic:
Just stumbled upon #2054, which mentions removing and installing the app again as a solution. Checked this page, which didn't mention anything about app removal.

I basically winged it with `app:remove` and it worked:

```bash
$ php occ app:remove calendar
calendar disabled
calendar removed
$ php occ app:install calendar
calendar installed
calendar enabled
```

This PR just adds app:remove under app commands list.


aleksandar-todorovic:
Note that the line above appears as changed because there was a space character at the end, which I've removed.

PR #1824, opened 2020-02-10
Makes the transifex build automatic.

pierreozoux:
The build is failing because we need to add transfiex credentials to the bot.

Can somebody help with that? I have no access to travis.

Closes #216

PR #2549, opened 2020-07-30
Build all the langs automatically.

### LINKED ISSUE #216: Documentation update and translation with transifex

pierreozoux:
You can see the result here:
https://doc-nextcloud-fr.indie.host/en/

I don't know why the build is failing on the CI, it works on my machine :/.

There is currently no default lang on the root ( "/" ), I hesitate to put en there. No strong opinion here.

We'll still have to take care of versioning, but let's do it in a another PR.

Makefile is a pain, that's why I created another script in the build folder, if you manage to put this line in Makefile, be my guest :) 
Even better, if you manage to tell transifex integration to send po files instead of pot, didn't find clues in their doc.

There is no more release or com, but just html as it is the same doc.

Closes #216
Closes #1824


MorrisJobke:
> Makefile is a pain, that's why I created another script in the build folder, if you manage to put this line in Makefile, be my guest :)
> Even better, if you manage to tell transifex integration to send po files instead of pot, didn't find clues in their doc.

In the server and apps code we have this line, that takes care of putting it in a `.po` file: https://github.com/nextcloud/server/blob/master/.tx/config#L6

MorrisJobke:
> There is no more release or com, but just html as it is the same doc.

\o/

MorrisJobke:
> There is no more release or com, but just html as it is the same doc.

I noticed that as well - before we had a special theme for the release (that's what you can find in the top right menu > help in your Nextcloud installation). Currently we just put in there the read the docs theme 🙈 That was most likely lost during the transition from our theme to read the docs theme.

pierreozoux:
Ok, managed to fix the build.

pierreozoux:
> > Makefile is a pain, that's why I created another script in the build folder, if you manage to put this line in Makefile, be my guest :)
> > Even better, if you manage to tell transifex integration to send po files instead of pot, didn't find clues in their doc.
> 
> In the server and apps code we have this line, that takes care of putting it in a `.po` file: https://github.com/nextcloud/server/blob/master/.tx/config#L6

MorrisJobke:
> @MorrisJobke You use the tx push and the tx pull in the other repo, and it is different than the transifex integration (I think). Do you have a strong opinion about it, or can we leave it this way.

Ah - I assumed that they use the same tool for their integration. Thus I thought it will just work, when you put the config in there.

pierreozoux:
Yeah, they must use the same tool, but, I'm a bit lazy to be honest, for 3 reasons:
 - we'll have to revisit for handling the versions
 - we'll have to retrigger the integration, and it will be again be a lot of PRs
 - there are 34 rst files, and so, I'll have to write this for 34 files, if I have to, I'll do it, but not before I'm sure I have to, and last time I did tx pull by hand, I got po, soI have good hope I don't need to write this file :)

skjnldsv:
> That was most likely lost during the transition from our theme to read the docs theme.

No, this was deliberately removed.
To simplify our process. This was a pain and failing all the time. Le'ts  only build one version and call it a day. No one is maintaining the old theme

user_manual/Makefile
```diff
@@ -39,22 +40,36 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 
-clean:
+local_clean:
```
skjnldsv, line 43: clean is still used in other places

user_manual/_templates/versions.html
```diff
@@ -0,0 +1,40 @@
+{% if READTHEDOCS %}
```
skjnldsv, line 1: What is this file about exactly?

pierreozoux: This is the template to have this:

582-456-max

As recommended [here](https://github.com/readthedocs/sphinx_rtd_theme/issues/857).

You can see it live on the [test website](https://doc-nextcloud-fr.indie.host/en/) I pushed.

skjnldsv: Thanks! :)
I was wondering why we needed a custom template for this.
Maybe we should add a comment to explain this is not part of th official rtd theme yet so we have to use our own?

PR #2714, opened 2020-07-31
Update general_troubleshooting.rst with correct CalDAV/CardDAV redirection

stefan123t:
I had setup CalDAV and CardDAV rewrite rules as suggested with trailing slash, which was always mentioned as missing under  Security & setup warnings. I removed the trailing slashes and voila it still works and the warning under Security & Setup is finally gone for good.


stefan123t:
Please review and commit the doc change.

PR #2715, opened 2020-07-31
Update example_centos.rst

hanserasmus:
Not sure why the needed to be built manually? But a fresh install with remi 7.4 gives `php-pecl-imagick-3.4.4-10.el8.remi.7.4` and `php-pecl-redis5-5.3.1-1.el8.remi.7.4` respectively.


skjnldsv:
I have no idea :shrug:

juliushaertl:
The idea was basically to also provide a way to run this without 3rdparty repos if needed. However I think remirepos are very common so fine by me :+1:

PR #2772, opened 2020-08-07
Example centos walkthrough, add working secure apache config

AndrewwHummer:
It would be really helpful if the example of a CentOS Configuration Walkthrough would directly contain a working and secure apache configuration.
Just the Reference to the detailed apache config is not adequate and too easy to overlook and users will end up with an unsecure and broken config. This could lead to the usual htaccess not working questions.

So here is my proposal :)

### DIFF
--- a/admin_manual/installation/example_centos.rst
+++ b/admin_manual/installation/example_centos.rst
@@ -21,8 +21,26 @@ Apache
 ::
 
     dnf install -y httpd
+    
+Create a virtualhost in ``/etc/httpd/conf.d/nextcloud.conf`` and add the following content to it:
 
-See :ref:`apache_configuration_label` for details.
+::
+
+  <VirtualHost *:80>
+    DocumentRoot /var/www/html/nextcloud/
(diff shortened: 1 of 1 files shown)


admin_manual/installation/example_centos.rst
```diff
@@ -21,8 +21,26 @@ Apache
...
+    <Directory /var/www/html/nextcloud/>
+      Require all granted
+      AllowOverride All
+      Options FollowSymLinks MultiViews
+
+      <IfModule mod_dav.c>
+        Dav off
+      </IfModule>
 
-See :ref:`apache_configuration_label` for details.
```
MorrisJobke, line 25: I would keep this to provide further documentation as well.

AndrewwHummer: good point

AndrewwHummer:
Is something still missing, or can me merge this ? :)

MorrisJobke (approved):

PR #2980, opened 2020-08-12
Add info about LDAP info caching

MorrisJobke:
From a chat with @blizzz

### DIFF
--- a/admin_manual/configuration_user/user_auth_ldap.rst
+++ b/admin_manual/configuration_user/user_auth_ldap.rst
@@ -703,12 +703,19 @@ Those mappings are done in the database table ``ldap_user_mapping`` and
 if something else is specified in *User Home Folder Naming Rule*), which
 contains files and meta data.
 
-As of Nextcloud 5 the internal user name and a visible display name are separated.
+The internal user name and a visible display name are separated.
 This is not the case for group names, yet, i.e. a group name cannot be altered.
 
 That means that your LDAP configuration should be good and ready before putting
 it into production. The mapping tables are filled early, but as long as you are
 testing, you can empty the tables any time. Do not do this in production.
+
+The attributes of users are fetched on demand (i.e. for sharing autocompletion
+or in the user management) and then stored inside the Nextcloud database to
+allow a better performance on our side. They are typically checked twice a day
+in batches from all users again. Beside that they are also refreshed during a
+login for this user or can be fetched manually via the occ command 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_user/user_auth_ldap.rst
```diff
@@ -703,13 +703,20 @@ Those mappings are done in the database table ``ldap_user_mapping`` and
...
-As of Nextcloud 5 the internal user name and a visible display name are separated.
+The internal user name and a visible display name are separated.
 This is not the case for group names, yet, i.e. a group name cannot be altered.
 
 That means that your LDAP configuration should be good and ready before putting
 it into production. The mapping tables are filled early, but as long as you are
 testing, you can empty the tables any time. Do not do this in production.
 
+The attributes of all users are fetched on demand and then stored inside the
+Nextcloud database to not overload the LDAP server with requests. They are
```
blizzz, line 714: It is not about request against LDAP server (searches and existence checks are done anyway unless caches), but performance on our side. Checking and storing data on sequentially on several places for several records has an impact on UX, therefore it is brought to the background. For login's it is less of a problem, as it is just one user being updated.

MorrisJobke: Refined the wording.

admin_manual/configuration_user/user_auth_ldap.rst
```diff
@@ -703,13 +703,20 @@ Those mappings are done in the database table ``ldap_user_mapping`` and
...
 it into production. The mapping tables are filled early, but as long as you are
 testing, you can empty the tables any time. Do not do this in production.
 
+The attributes of all users are fetched on demand (i.e. for sharing
+autocompletion or in the user management) and then stored inside the Nextcloud
+database to allow a better performance on our side. They are typically checked
+twice a day from all users again. Beside that they are also refreshed during a
+login for this user or can be fetched manually via the occ command 
+``occ ldap:check-user --update USERID`` where ``USERID`` is Nextcloud's user id.
+
```
blizzz, line 719:
```suggestion
The attributes of users are fetched on demand (i.e. for sharing
autocompletion or in the user management) and then stored inside the Nextcloud
database to allow a better performance on our side. They are typically checked
twice a day in batches from all users again. Beside that they are also refreshed during a
login for this user or can be fetched manually via the occ command 
``occ ldap:check-user --update USERID`` where ``USERID`` is Nextcloud's user id.

```

PR #3020, opened 2020-08-13
add documentation for the new "notification"-option of the rentention app

schiessle:
add documentation for the new "notification"-option of the rentention app


schiessle:
@MorrisJobke another one... maybe you can have here a quick look as well... Thanks!

PR #3191, opened 2020-08-21
Update external_storage_configuration_gui.rst

jospoortvliet:
Note that server-side encryption doesn't work with Nextcloud servers used as external storage.


jospoortvliet:
I dunno how backporting works but we probably should bring this back to older releases, at least it's relevant that you can't do this...

PR #3809, opened 2020-09-04
Fixed local redis socket configuration

jens7841:
Fixed a fail in the configuration for local redis socket.
The Configuration above worked for me (Debian 10, Nextcloud 19, php 7.3)


MorrisJobke:
This is just an example. For APCu the PHP module needs to be installed as well. Redis works there too, but let's keep it like it is now.

PR #3811, opened 2020-09-06

Bill-Thorsteinson:
Fix paths for apache2 mod_rewrite - add missing / at start of paths


MorrisJobke:
Works here without the leading `/`.

rakekniven:
Same here. Closing it?

donfranio:
Did not work here, I also even created a PR with the same content.

PR #4437, opened 2020-09-08
Update search.rst

daita:


developer_manual/digging_deeper/search.rst
```diff
@@ -183,7 +183,7 @@ Next, you'll see a dummy provider that returns a static set of results using the
         }
 
         public function search(IUser $user, ISearchQuery $query): SearchResult {
-            return MySearchResultEntry::complete(
```
ChristophWurst, line 186: in that case the definition of `MySearchResultEntry` is obsolete

ChristophWurst (changes requested):

developer_manual/digging_deeper/search.rst
nickvergessen: yeah that class should be wiped completly

daita: the thing is that in your example `MySearchResultEntry` extends `SearchResultEntry` that do not contains the method `complete` (from `SearchResult`)

ChristophWurst: Right, because that is documented incorrectly. Of course the search method should use `\OCP\Search\SearchResult::complete` to build the result

nickvergessen:
Should review and merge this sooner :)

MorrisJobke (approved):
Looks good 👍

PR #4459, opened 2020-09-10
Fix name of deprecated oc_requesttoken

ecsecta:
oc_requesttoken is marked as deprecated.

### DIFF
--- a/developer_manual/basics/front-end/js.rst
+++ b/developer_manual/basics/front-end/js.rst
@@ -34,14 +34,14 @@ listen to the ``OCA\Files::loadAdditionalScripts`` event:
 Sending the CSRF token
 ----------------------
 
-If any other JavaScript request library than jQuery is being used, the requests need to send the CSRF token as an HTTP header named **requesttoken**. The token is available in the global variable **oc_requesttoken**.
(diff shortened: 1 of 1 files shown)


MorrisJobke:
cc @juliushaertl @skjnldsv @ChristophWurst

developer_manual/basics/front-end/js.rst
```diff
@@ -34,14 +34,14 @@ listen to the ``OCA\Files::loadAdditionalScripts`` event:
 Sending the CSRF token
 ----------------------
 
-If any other JavaScript request library than jQuery is being used, the requests need to send the CSRF token as an HTTP header named **requesttoken**. The token is available in the global variable **oc_requesttoken**.
+If any other JavaScript request library than jQuery is being used, the requests need to send the CSRF token as an HTTP header named **requesttoken**. The token is available in the global variable **OC.requestToken**.
```
ChristophWurst, line 37: we should recommend https://nextcloud.github.io/nextcloud-auth/modules/_requesttoken_.html#getrequesttoken instead, but that requires a bundler. Not sure if that is too much for the tutorial.

juliushaertl: Yeah i think we could at least mention the package here, but for now I'd say having the OC.requestToken is still fine as a default example.

ChristophWurst: yeah, sure

ChristophWurst (approved):

PR #5038, opened 2020-09-23
Clarify the virtual host and directory installation methods.

srbaker:
The existing documentation makes it look as though Debian-likes only
supports directory-based configuration, and RH-likes only support
VirtualHost configuration.

I think this update is more clear about the options.

Signed-off-by: Steven R. Baker <steven@stevenrbaker.com>


MorrisJobke (approved):
Yes - a lot better indeed. 👍

PR #5039, opened 2020-09-23
Create example_openbsd.rst

gonzalo-:
Initial document for OpenBSD users, probably a lot of typos and missing parts, but it will be a great help so far.

### DIFF
--- a/admin_manual/installation/example_openbsd.rst
+++ b/admin_manual/installation/example_openbsd.rst
@@ -97,7 +97,6 @@ Create a virtualhost in ``/etc/httpd.conf`` and add the following content to it:
 	  }
   }
 
-::
 
 Make sure that httpd(8) is enabled and started::
 
@@ -122,15 +121,12 @@ It is recommended to add opcache to it::
   opcache.revalidate_freq=1
   opcache.save_comments=1
   
-::
 
 And increase some limits::
 
   post_max_size = 513M
   upload_max_filesize = 513M
   
-::
-
    
 We can enable the PHP modules with::
 
@@ -190,7 +186,6 @@ We installed redis before, we need to enable it and start it and also add it to 
     ),
     ...
     
-::
 
 Cron job
 --------
--- a/admin_manual/installation/example_openbsd.rst
+++ b/admin_manual/installation/example_openbsd.rst
@@ -1,10 +1,7 @@
 .. _openbsd_installation_label:
 
-==================================================================
-=								 =
-= Nextcloud does not have official OpenBSD or other *BSD support =
-=								 =
-==================================================================
+Nextcloud does not have official OpenBSD or other *BSD support
+==============================================================
 
 
 Example installation on OpenBSD
@@ -124,12 +121,14 @@ It is recommended to add opcache to it::
   opcache.max_accelerated_files=10000
   opcache.revalidate_freq=1
   opcache.save_comments=1
+  
 ::
 
 And increase some limits::
 
   post_max_size = 513M
   upload_max_filesize = 513M
+  
 ::
 
    
@@ -190,6 +189,7 @@ We installed redis before, we need to enable it and start it and also add it to 
       'timeout' => 0.0,
     ),
     ...
+    
 ::
 
 Cron job
--- a/admin_manual/installation/example_openbsd.rst
+++ b/admin_manual/installation/example_openbsd.rst
@@ -1,6 +1,6 @@
 .. _openbsd_installation_label:
 
-Nextcloud does not have official OpenBSD or other *BSD support
+Nextcloud does not have official OpenBSD or other BSDs support
 ==============================================================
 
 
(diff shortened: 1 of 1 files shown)


kesselb:
Thanks for your work 👍

Nextcloud does not support FreeBSD or OpenBSD. If we add a guide to the documentation people could assume that OpenBSD is supported.

gonzalo-:
I am the port maintainer of Nextcloud on OpenBSD what kind of support are you talking about?

kesselb:
Nextcloud is developed for Linux based operating systems. Most features should work on *BSD but it's not tested.

gonzalo-:
Well that is a pity, anyway, if there is not other place on the official documents for people outside of Linux, you can close the ticket, thanks.

kesselb:
I'm leaving it open. It's not my decision to merge or not. Probably good to merge with a disclaimer/warning.

MorrisJobke:
> Probably good to merge with a disclaimer/warning.

I would be fine with a warning box at the top of the page. 👍 

@rullzer @karlitschek @jancborchardt @nickvergessen Opinions on this?

karlitschek:
Yes. I think documentation about this is cool. The only thing we should avoid is to create the impression that it's working on BSD but in reality it's not. So if we are sure that is't working then let's add this documentation. If parts are not working we should add warnings. It is not working at all then no documentation please. I hope this makes sense :-)

landryb:
There are lots of *BSD users, for which nextcloud *works*. Sure there are some sorry linuxisms here and there, but *BSD users generally dont expect 'support' from upstream, rather from their packager. Giving them the middle finger in the documentation is not a nice solution either, so maybe some middle ground ? ;)

(yeah, using NC 19 on OpenBSD 6.7, it generally works.)

gonzalo-:
> The only thing we should avoid is to create the impression that it's working on BSD but in reality it's not.

It works without any big issue, maybe you want to say "We have no official *BSD support*?

Sounds more precise to me.

MorrisJobke:
> It works without any big issue, maybe you want to say "We have no official _BSD support_?

Yep - that sounds good IMO 👍

gonzalo-:
So should I do something or you will commit this with the warning we talked about?

kesselb:
>  missing parts

What parts are missing?

admin_manual/installation/example_openbsd.rst
```diff
@@ -0,0 +1,240 @@
...
+		  pass
+	  }
+
+	  location "/nextcloud/core/*" {
+		  root "/nextcloud"
+		  request strip 1
+		  pass
+	  }
+
+	  location "/nextcloud/settings/*" {
```
kesselb, line 70: This section is not longer necessary. Settings is nowadays an app in /nextcloud/apps.

admin_manual/installation/example_openbsd.rst
```diff
@@ -0,0 +1,240 @@
...
+    ),
+    ...
+::
+
+Cron job
+--------
+
+We need to add the Nextcloud cron job to get some tasks done by adding this entry on your cronjob::
+
+  */15 * * * * /usr/bin/ftp -Vo - https://domain.tld/cron.php >/dev/null
```
kesselb, line 199:
```suggestion
  */5 * * * * /usr/bin/ftp -Vo - https://domain.tld/cron.php >/dev/null
```

We also suggest to run this from CLI.

admin_manual/installation/example_openbsd.rst
gonzalo-: Cool, good to know.

admin_manual/installation/example_openbsd.rst
gonzalo-: Yes, but usually httpd(8) runs under a chroot() so for some users is more useful to just put a cronjob.

gonzalo-:
> What parts are missing?

Nothing really important, just cosmetic things, but you can have NC working on OpenBSD with this tutorial plus our own README file.

gonzalo-:
@kesselb @MorrisJobke  Can we have this in?

kesselb:
Can you fix the linter warnings?

gonzalo-:
@kesselb do you mean the cron and the "location "/nextcloud/settings/*" {" things? I changed it already, maybe I did it wrong :/

kesselb:
image
image
image
image
image

gonzalo-:
@kesselb better now? :D

kesselb:
@gonzalo- thanks :+1: 

image

To fix the above warning we need to reference the guide: https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/index.rst.

image

I don't know how to fix this warning but it's not a blocker.

image

Please sign the commits: https://github.com/nextcloud/documentation/pull/5039/checks?check_run_id=1925653519

gonzalo-:
@kesselb cool! 

thanks for the hint! Already open another PR with that, and fixed the other one (I think).

PR #5157, opened 2020-10-04
add migration to tutorial

dassio:
Signed-off-by: dassio <dassio@icloud.com>

during development, migration can be changed multiple times, using the updating version tag method to trigger migration has the following drawbacks:

1. version number has to be increased multi times when only one feature is added 
2. [wrong migration has to be deleted from the oc_migrations table manually](https://help.nextcloud.com/t/clean-migration-for-unpublished-app/89911), or the server will complain same migration already be used 

the migration command does not have that constraint, just drop the table and run `sudo -u www-data php ./occ migrations:execute photos 000000Date20201002183800` will recreate the table for you.

### DIFF
--- a/developer_manual/basics/events.rst
+++ b/developer_manual/basics/events.rst
@@ -24,7 +24,7 @@ This mechanism is a versatile and typed approach to events in Nextcloud's php co
 Naming scheme
 `````````````
 
-The name should reflect the subject and the actions. Prefixing event classes with `Event` makes it easier to recognize their purpose.
+The name should reflect the subject and the actions. Suffixing event classes with `Event` makes it easier to recognize their purpose.
 
 For example, if a user is created, a `UserCreatedEvent` will be emitted.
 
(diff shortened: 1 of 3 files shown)


developer_manual/basics/events.rst
```diff
@@ -24,7 +24,7 @@ This mechanism is a versatile and typed approach to events in Nextcloud's php co
 Naming scheme
 `````````````
 
-The name should reflect the subject and the actions. Prefixing event classes with `Event` makes it easier to recognize their purpose.
+The name should reflect the subject and the actions. Suffixing event classes with `Event` makes it easier to recognize their purpose.
```
ChristophWurst, line 27: Unrelated to your actual PR's topic but :+1: on this one

ChristophWurst:
Mind submitting 097b700 on a separate branch/PR to keep things a bit separate? :v:

MorrisJobke (approved):
Fine with me with the two separate commits.

Issue #610, opened 2017-11-19
Use GET instead of POST User Provisioning API > Get members of a group example block


dachinat:
If you take a look at https://docs.nextcloud.com/server/12/admin_manual/configuration_user/user_provisioning_api.html page, there is *Get members of a group* section, where you use GET method to retrieve members of group, example block by mistake states POST.
<img width="667" alt="screen shot 2017-11-19 at 3 41 22 pm" src="https://user-images.githubusercontent.com/22872749/32990276-285ed890-cd40-11e7-97c8-d732954ca11e.png">

MorrisJobke:
cc @rullzer @schiessle @nickvergessen

PR #5228, opened 2020-10-08
Use GET instead of POST User Provisioning API

Nienzu:
Solve #610


kesselb:
LGTM 👍

Please have a look at https://github.com/nextcloud/documentation/runs/1224800043 to fix the build.

kesselb (approved):

PR #5230, opened 2020-10-08
Document CI and static analysis for apps

ChristophWurst:
https://github.com/nextcloud/mail/pull/3757, but a bit more generic so we can spread the awesomeness of psalm.

Rendered: 
Screenshot_2020-10-08 Continuous Integration — Nextcloud latest Developer Manual latest documentation


skjnldsv (approved):
:elephant:

PR #5245, opened 2020-10-12
sync_android.rst: With Nextcloud app: Specify the contact-grouping method that the user needs to choose

jivanpal:
User must specify "Groups are per-contact categories" even when using the Nextcloud app flow. This behaviour is seen with the following platform:

- Android 9.0 (Samsung One UI 1.0)
- Nextcloud for Android, version 3.13.1
- DAVx⁵ for Android, version 3.3.2-ose, installed via F-Droid

If this is not the case on other platforms, perhaps mention that the new step (4) may happen automatically?


MorrisJobke:
cc @tobiasKaminsky

tobiasKaminsky:
No idea, to be honest :-)

@skjldnv

skjnldsv (approved):
Yes, so far we still don't support groups as unique vcards https://github.com/nextcloud/contacts/issues/213

PR #5249, opened 2020-10-14
Add admin occ app installation docs

ChristophWurst:


admin_manual/configuration_server/occ_command.rst
```diff
@@ -201,6 +201,14 @@ The ``app`` commands list, enable, and disable apps::
   app:update       update an app or all apps
   app:remove       disable and remove an app
 
+Download and install an app::
+
+ sudo -u www-data php occ app:install twofactor_totp
+
+Install but don't enable (e.g. in a clustered environment)::
```
MorrisJobke, line 208: This does only make sense if this is executed on every node. In those cases it's often just deployed via another tool (aka configuration management).

```suggestion
Install but don't enable (e.g. in a clustered environment where this needs to be executed on every node)::
```

ChristophWurst: is that so? you can also rsync the app over :thinking:

ChristophWurst: I'd rather drop the clustered environment note instead of giving an unrelated instruction here

MorrisJobke: > is that so? you can also rsync the app over 🤔

This would be one way, but then you can also download the archive directly, if other tools are used anyways. Also it could be that the Nextcloud directory is on a shared NFS directory and then it doesn't matter anymore.

To sum it up: "It depends" and we should maybe drop that comment about clustered environment.

ChristophWurst: > I'd rather drop the clustered environment note instead of giving an unrelated instruction here

:roll_eyes:

MorrisJobke (approved):

PR #5284, opened 2020-10-24
nginx: fix autoloading for some builtins apps

Mic92:
In https://github.com/nextcloud/documentation/commit/6b40a23ab928a9b84d777377f9bc7dc2f75b15f1#diff-e6fc0cd0bc2157559d74ab9196ea4209f9205a94865839d0f7b32b9810365c78 
the way nginx access files was changed. This broke some apps like user_ldap, where ajax function are called from the directory directly. They need to be executed through the top-level index.php, or otherwise they will find their dependencies. Maybe this should be reverted back to `rewrite ^ /index.php;`? cc @jivanpal

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -174,7 +174,7 @@ webroot of your nginx installation. In this example it is
       }
       
       location / {
-          try_files $uri $uri/ /index.php$request_uri;
+          try_files  $uri/ /index.php$request_uri;
       }
   }
 


MorrisJobke:
cc @josh4trunks

admin_manual/installation/nginx.rst
```diff
@@ -174,7 +174,7 @@ webroot of your nginx installation. In this example it is
       }
       
       location / {
-          try_files $uri $uri/ /index.php$request_uri;
+          try_files  $uri/ /index.php$request_uri;
```
MorrisJobke, line 177:
```suggestion
          try_files $uri/ /index.php$request_uri;
```

josh4trunks:
I think the full fix is explained here.
https://github.com/nextcloud/documentation/pull/2197#issuecomment-703085112

Does anyone mind making this a PR so we can review/test?

Mic92:
OK. Maybe original PR should be revoked completely.

PR #5410, opened 2020-11-12
Update transfer_ownership.rst

cdehe7519:
Traduction en Français du chapitre Transférer la Propriété - Translation in French of the section "Transférer la propriété"


MorrisJobke:
For translations we use transifex. So the repo has the English version and the French version is in transifex: https://www.transifex.com/nextcloud/nextcloud-user-documentation/language/fr/

Therefore I will close this here.

PR #5431, opened 2020-11-20

donfranio:
this was necessary for me to have those rules work and also makes perfect sense as a `/` is imho always at the beginning of the path.


MorrisJobke:
Same as #3811 

cc @rakekniven

nickvergessen:
Rebased in https://github.com/nextcloud/documentation/pull/8093

PR #5438, opened 2020-11-22
Improve suggested command line

mat-m:
Since most of the times,the www-data user has no shell, you cannot easily cd as current user, because directory is protected, and  then cannot sudo locally.


mat-m:
force push after amending my sign-off to pass DCO

MorrisJobke (approved):

PR #5484, opened 2020-11-26
Add app code checker for CI

ChristophWurst:
Adding this for the sake of completeness. Though we hopefully drop this at some point.

Rendered:

Bildschirmfoto von 2020-11-26 11-36-17

### DIFF
--- a/developer_manual/digging_deeper/continuous_integration.rst
+++ b/developer_manual/digging_deeper/continuous_integration.rst
@@ -65,6 +65,45 @@ A lint of all php source files can find syntax errors that could crash the appli
         run: composer run lint
 
 
+App code checker
+----------------
+
+Nextcloud comes with an *app code checker* that analyzes the code of an app and warns if deprecated APIs are used and similar. Most of the checks can also be done with :ref:`static analysis and 3rd party tools<app-static-analysis>` but a few checks are currently only possible with this tool.
+
+The following Github action checks an app against the latest development version of Nextcloud, the `stable20` branch and the v20.0.0 release. Adjust this to your needs.
+
+.. code-block:: yaml
+
+  name: App code check
+  on: pull_request
+    app-code-check:
+      runs-on: ubuntu-latest
+      strategy:
+        matrix:
+          nextcloud-versions: ['master', 'stable20', 'v20.0.0']
+      name: Nextcloud ${{ matrix.nextcloud-versions }} app code check
+      steps:
+      - name: Set up php7.4
+        uses: shivammathur/setup-php@master
+        with:
+          php-version: 7.4
+          extensions: ctype,curl,dom,gd,iconv,intl,json,mbstring,openssl,posix,sqlite,xml,zip
+          coverage: xdebug
+      - name: Checkout Nextcloud
+        run: git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b ${{ matrix.nextcloud-versions }} nextcloud
+      - name: Run tests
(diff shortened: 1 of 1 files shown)


developer_manual/digging_deeper/continuous_integration.rst
```diff
@@ -65,6 +65,45 @@ A lint of all php source files can find syntax errors that could crash the appli
...
+      name: Nextcloud ${{ matrix.nextcloud-versions }} app code check
+      steps:
+      - name: Set up php7.4
+        uses: shivammathur/setup-php@master
+        with:
+          php-version: 7.4
+          extensions: ctype,curl,dom,gd,iconv,intl,json,mbstring,openssl,posix,sqlite,xml,zip
+          coverage: xdebug
+      - name: Checkout Nextcloud
+        run: git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b ${{ matrix.nextcloud-versions }} nextcloud
```
skjnldsv, line 93: Why ot use the `actions/checkout@master` action?

developer_manual/digging_deeper/continuous_integration.rst
```diff
@@ -65,6 +65,45 @@ A lint of all php source files can find syntax errors that could crash the appli
...
+      - name: Run tests
+        run: php -f nextcloud/occ maintenance:install --database-name oc_autotest --database-user oc_autotest --admin-user admin --admin-pass admin --database sqlite --database-pass=''
+      - name: Checkout
+        uses: actions/checkout@master
```
skjnldsv, line 97: We should use an official tag, not master 
`actions/checkout@v2`

developer_manual/digging_deeper/continuous_integration.rst
skjnldsv, line 93:
```suggestion
      - name: Checkout Nextcloud
        uses: actions/checkout@v2
        with:
          repository: nextcloud/server
          ref:  ${{ matrix.nextcloud-versions }}
```

nickvergessen:
app:check-code should be deprecated and not be used anymore.
Psalm has you covered and is much more accurate. The lists for the check-code command have not been updated in quite a while (oldest deprecation is 14: https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/lib/private/App/CodeChecker/DeprecationCheck.php )

developer_manual/digging_deeper/continuous_integration.rst
ChristophWurst: :shrug: I copied it from Mail. IIRC there was a time when the checkout action didn't work for us

ChristophWurst:
> app:check-code should be deprecated and not be used anymore.
> Psalm has you covered and is much more accurate. The lists for the check-code command have not been updated in quite a while (oldest deprecation is 14: https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/lib/private/App/CodeChecker/DeprecationCheck.php )

Well then there is really no point in continuing this.

MorrisJobke:
> The lists for the check-code command have not been updated in quite a while (oldest deprecation is 14:

But we can add it them again?

Can Psalm also check for deprecations? The private namespace usage is most likely covered by the usage of @ChristophWurst library that provides the OCP interfaces only (instead of using the server code itself).

ChristophWurst:
> Can Psalm also check for deprecations?

Yes: https://psalm.dev/docs/running_psalm/issues/DeprecatedMethod/

> The private namespace usage is most likely covered by the usage of @ChristophWurst library that provides the OCP interfaces only

Also yes :wink:

PR #5485, opened 2020-11-26
Remove dead preupdate.php note

ChristophWurst:
This file is not included anymore

Ref https://github.com/nextcloud/server/pull/24393


juliushaertl (approved):
:broom:

PR #5491, opened 2020-11-29
Add HSTS Header

bradleyburgess:
Added HSTS Header as prompted by Security Setup Warnings in admin panel.


MorrisJobke:
This is already in there. Line 83 and 233 in that file. They are only commented out because they have implications on other subdomains as well and therefore need to be explicitly added on purpose by the admin.

https://github.com/nextcloud/documentation/pull/5491/files#diff-e6fc0cd0bc2157559d74ab9196ea4209f9205a94865839d0f7b32b9810365c78R233

Therefore I close this. Thanks for the contribution anyways :)

bradleyburgess:
@MorrisJobke Ah, thanks!

PR #5571, opened 2020-12-06
Fix punctuation and "Share"

Valdnet:


user_manual/files/sharing.rst
```diff
@@ -27,16 +27,16 @@ A number of options are available for public *folder* shares:
 .. figure:: ../images/sharing_public_folder.png
 
 * **Read only** to allow viewing and downloading
-* ** Allow upload and editing**
-* With ** File drop**, the sharee can only upload files to a folder without seeing the files that are already in that folder.
```
tflidd, line 31: that is a person to which the folder is shared (https://en.wiktionary.org/wiki/sharee)
If that term is perhaps less know there might be a better word or some explanation when hovering over the term...

user_manual/files/sharing.rst
```diff
@@ -48,12 +48,12 @@ When sharing with users, groups, circles or members of a Talk conversation, righ
 
 .. figure:: ../images/sharing_internal.png
 
-As a sharee, you can configure if you automatically want to accept all incoming shares and have them added to your root folder, or if you
```
tflidd, line 51: same, that's a person...

tflidd (changes requested):
Apart from the `sharee`, the rest looks ok for me

user_manual/files/sharing.rst
Valdnet: I understand that word now. Well, we leave them unchanged.

user_manual/files/sharing.rst
Valdnet: @tflidd Also corrected :wink:.

tflidd (approved):

PR #5682, opened 2020-12-11
Lots of fixes in "access_webgui"

Valdnet:


wiswedel (approved):
Thanks for the fixes!

PR #5784, opened 2020-12-12
fix a typo

voicevon:


ChristophWurst:
this actually works. both variants to :wink:

PR #5785, opened 2020-12-13
Update "Development Environment" guide

nina-py:
While setting up a local dev checkout of nextcloud/server, I noticed that the part about creating the 'config' directory is outdated: this directory already exists in the repository and doesn't need to be created during setup.

I have fixed that part and also clarified some wordings elsewhere in this guide.


tflidd (approved):

nina-py:
Thank you @tflidd!

PR #5831, opened 2020-12-19
Updated the docs for ease

anirudhdggl:
Modified the docs so that an admin does not have to go through the troubleshooting and figuring out and spend excessive efforts there. The following have been included in this PR

- Added a note to warn the user of issues he can experience while setting up cron, thereby preventing him the trouble of finding his way out
 - Added a small curl command to enable the admin to find out if HTTP2 is already enabled in the application or not

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -64,8 +64,6 @@ Using the operating system cron feature is the preferred method for executing
 regular tasks.  This method enables the execution of scheduled jobs without the
 inherent limitations the Web server might have.
 
-.. note:: Please make sure that your cron.php file has executable permissions and is owned by www-data
-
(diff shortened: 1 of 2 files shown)


admin_manual/configuration_server/background_jobs_configuration.rst
```diff
@@ -64,6 +64,8 @@ Using the operating system cron feature is the preferred method for executing
 regular tasks.  This method enables the execution of scheduled jobs without the
 inherent limitations the Web server might have.
 
+.. note:: Please make sure that your cron.php file has executable permissions and is owned by www-data
```
tflidd, line 67: If you use the command in the documentation, you execute `php`, so `cron.php` does not have to be executable itself, no?

anirudhdggl: Yes, you're right. Made the required changes 😃

tflidd (approved):

tflidd (approved):

PR #5855, opened 2020-12-30
Update example_ubuntu.rst

musicaljoeker:
The flags for the tar command must be in a different order otherwise tar will complain that it can't find the file. See this post for further explanation: https://superuser.com/a/727659


Omemanti:
Duplicate of #5853

musicaljoeker:
Yes it appears that it is @Omemanti. I'll close this issue and defer the fix to yours.

PR #5856, opened 2021-01-03
style(server config): alphabetise preview providers

SimJoSt:


MorrisJobke:
@SimJoSt Thanks for this - this specific part is fetched from the config sample in the server repo: https://github.com/nextcloud/server/blob/master/config/config.sample.php

So that file needs to be updated instead.

Could you send a PR against server?

SimJoSt:
I will, as soon as I have the time.

SimJoSt:
@MorrisJobke I finally did it in https://github.com/nextcloud/server/pull/38876

PR #5857, opened 2021-01-03
remove global browser caching

### LINKED ISSUE #5472: Nginx caching delays for dynamic elements
https://github.com/nextcloud/documentation/blob/122ed88e536d45afa6b671c8f90a8d43f9510b21/admin_manual/installation/nginx.rst#L103-L104

This seems to cause update delays for dynamic elements requested via APIs.

For example when using the news plugin, it causes feed items which I've marked as read still being displayed in the list for 60 seconds (`/apps/news/items?limit=40&oldestFirst=false&search=&showAll=false&type=6`)

The live update in the Administration > Logging section has the same problem `/apps/logreader/poll?lastReqId=UwWXOsccR9Rw6DoVbATf` is only executed every 1 minute, every other call is loaded from the cache.

@jivanpal, @josh4trunks any idea how to fix this other than removing the default caching policy?

_Originally posted in https://github.com/nextcloud/documentation/pull/2197#discussion_r530570586_

anoymouserver:
API calls for the frontend get stuck in the browser cache and the Apache `.htaccess` also doesn't contain anything similar.
Fixes #5472

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -100,9 +100,6 @@ webroot of your nginx installation. In this example it is
       # always provides the desired behaviour.
       index index.php index.html /index.php$request_uri;
       
-      # Default Cache-Control policy
-      expires 1m;
-      
       # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
       location = / {
           if ( $http_user_agent ~ ^DavClnt ) {
@@ -287,9 +284,6 @@ The configuration differs from the "Nextcloud in webroot" configuration above in
           # `try_files $uri $uri/ /nextcloud/index.php$request_uri`
           # always provides the desired behaviour.
           index index.php index.html /nextcloud/index.php$request_uri;
-          
-          # Default Cache-Control policy
-          expires 1m;
 
           # Rule borrowed from `.htaccess` to handle Microsoft DAV clients
           location = /nextcloud {


MorrisJobke:
See https://github.com/nextcloud/documentation/pull/2197#discussion_r530570586

cc @jivanpal @josh4trunks

josh4trunks:
@jivanpal it looks like this was added back in summer 2020. I also do not see any default `max-age` in the `.htaccess`. Please confirm if there is any specific reason you want to keep it, otherwise I agree with @anoymouserver that we should remove it.

jivanpal:
@josh4trunks, the Nginx config before I made PR #2197 had `add_header Cache-Control "public, max-age=15778463";` I used `expires 1m` as a default for the current config just because it's generally good practice to cache things, but if this is affecting functionality, then of course please change it; I am not familiar enough with the inner workings of Nextcloud to really be able to provide any more info than I already have in the comments on the config, so I am happy for you to make any changes to it.

Also, from #2197, I said:

> Regarding the specific issue with `Cache-Control` headers there, do you know if Nginx overrides HTTP headers set by PHP? If so, we probably do need to get rid of the `expires` mentioned by @anoymouserver and instead rely on PHP to set the header, or do some sort of `if` on `$http_cache_control`.

If `expires 1m` is *never* of any real benefit, then I would suggest just getting rid of it.

josh4trunks:
:+1:

MorrisJobke (approved):

PR #5894, opened 2021-01-06
Caddyfile reverse_proxy example

serafdev:
Added a Caddy reverse_proxy example


MorrisJobke (approved):

l3miage-pelletl:
I think this is better:
```
subdomain.example.com {
        rewrite /.well-known/carddav /remote.php/dav
        rewrite /.well-known/caldav /remote.php/dav
        reverse_proxy localhost
}
```

serafdev:
I agree @l3miage-pelletl that looks much better, thank you!

PR #6001, opened 2021-01-16
Add documentation for release automation via GitHub actions

Grotax:
As discussed in #4545 it's very easily possible to automate your app releases, which makes your life easier and prevents mistakes like "locally I didn't checkout the git tag".

We at https://github.com/nextcloud/news use this for a while now and are very happy with the result.
For me as a maintainer and the one who published the app, things got so much easier and reduced mistakes to basically zero.

Feedback welcome.

I saw there is a rule to wrap after 80 characters, https://github.com/nextcloud/documentation/blob/master/README.rst#manuals as that wasn't applied to other files and I think such a rule doesn't make sense, I didn't apply it.

### DIFF
--- a/developer_manual/app_publishing_maintenance/release_automation.rst
+++ b/developer_manual/app_publishing_maintenance/release_automation.rst
@@ -1,0 +1,203 @@
+==================
+Release Automation
+==================
+
+.. sectionauthor:: Benjamin Brahmer <info@b-brahmer.de>
+
+Automation is a great thing, it prevents mistakes and makes your life easier, which gives you more time to work on all those features you wanted to implement.
+
+GitHub Actions
+--------------
+If your application's repository lives on GitHub, as many Nextcloud applications do, GitHub Actions is a great way to automate the release of your app from the git repository into the Nextcloud App Store.
+
+One easy way to get you started is to use https://github.com/R0Wi/nextcloud-appstore-push-action in your repository together with a few other actions. You can automatically build your app and publish it to the App Store. It supports pre-releases and code signing.
+To get started you create a new yaml file in the ``.github/workflows`` directory.
+
+.. code-block:: yaml
+
+  name: Build and publish app release
+
+  on:
+    release:
+        types: [published]
+
+  env:
+    APP_NAME: news
+
+  jobs:
+    build_and_publish:
+      environment: release
+      runs-on: ubuntu-latest
+      name: "Release: build, sign and upload the app"
+      strategy:
+        matrix:
+          php-versions: ['7.4']
+          nextcloud: ['stable21']
+          database: ['sqlite']
+      steps:
+        - name: Checkout
+          uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+
+        - name: Setup PHP
+          uses: shivammathur/setup-php@afefcaf556d98dc7896cca380e181decb609ca44
+          with:
+            php-version: ${{ matrix.php-versions }}
+            extensions: pdo_sqlite,pdo_mysql,pdo_pgsql,gd,zip
+            coverage: none
+
+        - name: Set up server non MySQL
+          uses: SMillerDev/nextcloud-actions/setup-nextcloud@fae87e29aa7cdf1ea0b8033c67f60e75b10be2cd
+          with:
+            cron: false
+            version: ${{ matrix.nextcloud }}
+            database-type: ${{ matrix.database }}
+
+        - name: Prime app build
+          run: make
+
+        - name: Configure server with app
+          uses: SMillerDev/nextcloud-actions/setup-nextcloud-app@fae87e29aa7cdf1ea0b8033c67f60e75b10be2cd
+          with:
+            app: ${{ env.APP_NAME }}
+            check-code: false
+
+        - name: Create signed release archive
+          run: |
+            cd ../server/apps/${{ env.APP_NAME }} && make appstore
+          env:
+            app_private_key: ${{ secrets.APP_PRIVATE_KEY }}
+            app_public_crt: ${{ secrets.APP_PUBLIC_CRT }}
+
+        - name: Upload app tarball to release
+          uses: svenstaro/upload-release-action@483c1e56f95e88835747b1c7c60581215016cbf2
+          id: attach_to_release
+          with:
+            repo_token: ${{ secrets.GITHUB_TOKEN }}
+            file: ../server/apps/${{ env.APP_NAME }}/build/artifacts/appstore/${{ env.APP_NAME }}.tar.gz
+            asset_name: ${{ env.APP_NAME }}.tar.gz
+            tag: ${{ github.ref }}
+            overwrite: true
+
+        - name: Upload app to Nextcloud appstore
+          uses: R0Wi/nextcloud-appstore-push-action@a011fe619bcf6e77ddebc96f9908e1af4071b9c1
+          with:
+            app_name: ${{ env.APP_NAME }}
+            appstore_token: ${{ secrets.APPSTORE_TOKEN }}
+            download_url: ${{ steps.attach_to_release.outputs.browser_download_url }}
+            app_private_key: ${{ secrets.APP_PRIVATE_KEY }}
+            nightly: ${{ github.event.release.prerelease }}
+
+        - name: Delete crt and key from local storage
+          run: rm -f ~/.nextcloud/certificates/*
+
+Make sure to check the used actions for useful updates, as they are pinned to a specific sha1 to prevent unnoticed harmful changes.
+
+For this workflow to work we need to provide a few variables.
+
+* ``APP_NAME`` set the name of your app, directly in the yaml
+
+Then we have a few secrets make sure to handle them with care.
+If your repository lives within the nextcloud organization you need to use an environment.
+
+.. code-block:: yaml
+
+      jobs:
+        build_and_publish:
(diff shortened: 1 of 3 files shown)


ChristophWurst:
One thing we should definitely warn about is using this process in our github.com/nextcloud organization because I think anyone with write permissions could push a tag and therefore trigger a release.

Grotax:
Is there any option on the organization level to change that? Only allow members of the team that belongs to the repository, to create releases? Tags are a different topic, they are not used as trigger here.

But I agree a warning should be included.

ChristophWurst:
> EDIT: well I guess tag and release are tightly connected to each other. But still it would be maybe a good thing if I wasn't able to create a new release/push a tag to the server repository.

I agree, but we can't change the Github authorization model

Bildschirmfoto von 2021-01-18 09-54-02

https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization

Grotax:
Yea I saw that, but you could drop the general write access in the organization.

To gain write access someone would need to be in a specific team, that would prevent the 575 members also from creating new random branches in our repository, I know that would require a bit of work but I think it would be worth the effort.
The management of the teams could be delegated to the maintainer(s) of a repository.

ChristophWurst:
> To gain write access someone would need to be in a specific team, that would prevent the 575 members also from creating new random branches in our repository, I know that would require a bit of work but I think it would be worth the effort.
> The management of the teams could be delegated to the maintainer(s) of a repository.

Full ack. I'll discuss with our admins.

ChristophWurst:
It was decided that the Github default permissions will remain for the organization.

Therefore my personal suggestions is that we don't document this potentially exploitable setup.

Grotax:
Well that's a shame, so that means no automation in the nextcloud organization. Should we migrate news out of the organization?

tobiasKaminsky:
@Grotax have you seen environments?
It seems that you can then associate workflows to be run only within this specific environement.
This can be limited to people/teams.
I I understand it correctly you could then create a "release" environment with e.g. only 2-3 devs in it.

Grotax:
@tobiasKaminsky no I didn't see that. Thank you very much, we will try that and if it works for us I will update the documentation accordingly.

tobiasKaminsky:
> @tobiasKaminsky no I didn't see that. Thank you very much, we will try that and if it works for us I will update the documentation accordingly.

We are also trying it right now.

tobiasKaminsky:
@Grotax 
we tested it and it seems to work exactly as it is needed (at least from my understanding):
There is a release workflow, which only runs within environment "test".
https://github.com/tobiasKaminsky/android-nextcloud/blob/master/.github/workflows/release.yml

Environment test is controlled only by admins:
image

Every user with write access can tag, but this specific release workflow needs approval by the specified user/teams.
Only then the app is published to e.g. app store.

tobiasKaminsky:
Once you confirm that this also work, we can rewrite the documentation.
If you need any help, or have further questions, just ping me here :+1:

MorrisJobke:
> Once you confirm that this also work, we can rewrite the documentation.
> If you need any help, or have further questions, just ping me here 👍

Keep in mind that this is something for the nextcloud org mostly - so both ways should be fine IMO.

Grotax:
I will document both and make the implications clear, after I could test the new setup a bit.

Grotax:
After a couple of successful releases and improvements I updated the documentation.

developer_manual/app_publishing_maintenance/release_automation.rst
```diff
@@ -0,0 +1,197 @@
...
+
+    env:
+    app_private_key: ${{ secrets.APP_PRIVATE_KEY }}
+    app_public_cert: ${{ secrets.APP_PUBLIC_CERT }}
+
+Also make sure to remove ``environment: release`` .
+
+Makefile changes for code signing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+As your certificate and your private key now are stored in environment variables you need somehow covert them to a file.
```
tobiasKaminsky, line 124:
```suggestion
As your certificate and your private key now are stored in environment variables you need somehow convert them to a file.
```

tobiasKaminsky (approved):

developer_manual/app_publishing_maintenance/release_automation.rst
```diff
@@ -0,0 +1,197 @@
...
+            id: attach_to_release
+            with:
+              repo_token: ${{ secrets.GITHUB_TOKEN }}
+              file: ../server/apps/${{ env.APP_NAME }}/build/artifacts/appstore/${{ env.APP_NAME }}.tar.gz
+              asset_name: ${{ env.APP_NAME }}.tar.gz
+              tag: ${{ github.ref }}
+              overwrite: true
+
+          - name: Upload app to Nextcloud appstore
+            uses: R0Wi/nextcloud-appstore-push-action@v1
```
LukasReschke, line 81: How does version pinning work? What happens if a malicious `v1.0.4` is released or such? Similar for the other versions above and the references to `main`.

Grotax: What do you mean by version pinning?

We have two areas were an app can be "released", we have GitHub and the App Store.
Anyone that is part of the nextcloud organization or got dedicated write access to a repository within nextcloud can create a new release on GitHub, a new release automatically triggers the creation of a tag and to my understanding there is nothing we can do about that.

So let's say one approval is required for app X to merge into main, User A commits the malicious code, User B reviews and approves it, the code gets merged. User A continues to release a new version via GitHub, a new tag is created at the same time, we end up with a release on GitHub without the `app.tar.gz`

For the run of this workflow as you find it in the documentation an approval is required. Anyone that is an admin of the nextcloud organization or the repository can change the approval requirements. There is no protection against a rouge admin in GitHub apart from audit logs, which I can't access. The approval is handled via GitHub currently either directly in the action or via the beta notification system, things might change though.

If someone prefers to host their app in a private account the organization part of course can be ignored and it only matters who gains write access to the repository.

Grotax: https://github.com/nextcloud/documentation/blob/dc3fe5d805244f04d013e414b96a3ce5b5912a50/developer_manual/app_publishing_maintenance/release_automation.rst#the-process

Grotax:
I will test, the workflow with a sha1 pinned version of the appstore workflow, this will take some time, depending on when news will release a new update.

developer_manual/app_publishing_maintenance/release_automation.rst
```diff
@@ -0,0 +1,201 @@
...
+
+One easy way to get you started is to use https://github.com/R0Wi/nextcloud-appstore-push-action in your repository together with a few other actions. You can automatically build your app and publish it to the App Store. It supports pre-releases and code signing.
+To get started you create a new yaml file in the ``.github/workflows`` directory.
+
+.. code-block:: yaml
+
+    name: Build and publish app release
+
+    on:
+    release:
```
MorrisJobke, line 21: The indentation looks off here and as it is YAML it most likely will cause issues when copied as it is. 🙈 Minde to adjust this?

developer_manual/app_publishing_maintenance/release_automation.rst
```diff
@@ -0,0 +1,201 @@
...
+If your repository lives within the nextcloud organization you need to use an environment.
+
+.. code-block:: yaml
+
+      jobs:
+        build_and_publish:
+          environment: release
+          runs-on: ubuntu-latest
+
+In this example we use the "release" environment, open the settings of your repository and open the "Envionments" tab, add a new environment with the name release, make sure to activate "Required reviewers" only add the people you trust, they will be able to approve a release.
```
MorrisJobke, line 107:
```suggestion
In this example we use the "release" environment, open the settings of your repository and open the "Envionments" tab, add a new environment with the name "release", make sure to activate "Required reviewers" only add the people you trust, they will be able to approve a release.
```

developer_manual/app_publishing_maintenance/release_automation.rst
```diff
@@ -0,0 +1,201 @@
...
+Save your rules and at the bottom add the following envionment secrets.
+
+* ``APP_PRIVATE_KEY`` your apps private key
+* ``APP_PUBLIC_CRT`` your apps certificate, this one could be public but for easy usage we add it as a secret
+* ``APPSTORE_TOKEN`` you get this from the App Store as a registered developer https://apps.nextcloud.com/account/token
+
+If your app does not live in the Nextcloud organization you may also add the secrets above to the "Secrets" section but be careful everyone with write access to your repository will be able to create releases. Make also sure do delete the environment statement.
```
MorrisJobke, line 114:
```suggestion
If your app does not live in the Nextcloud organization you may also add the secrets above to the "Secrets" section but be careful everyone with write access to your repository will be able to create releases. Make also sure to delete the environment statement.
```

MorrisJobke (approved):
Beside my nitpicks this looks great 👍 Thanks :)

developer_manual/app_publishing_maintenance/release_automation.rst
Grotax: yea fixed and updated

Grotax:
I fixed the last few things that @MorrisJobke pointed out, squashed my commits and rebased it on master.
Any other comments or can this be merged?

MorrisJobke (approved):

PR #1469, opened 2019-06-01
Update access_webdav.rst

iPhysicX:
added a simple reminder for using an app-password for login.
i had to try and error a lot until i realize this. :)

PR #6065, opened 2021-01-20
Add note about app passwords for webdav

MorrisJobke:
Same as #1469 but with fixed typo.


MorrisJobke:
Already approved over there.

Issue #4758, opened 2020-09-16
The suggested configuration for PHP MySQL driver is wrong due to MySQL (Original) beeing removed since PHP 7.0


kroko:
## Issue

The docs about PHP MySQL driver configuration are so outdated they are wrong

https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/linux_database_configuration.html#configuring-a-mysql-or-mariadb-database

https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_database/linux_database_configuration.rst#L101-L120

> The PHP configuration in `/etc/php7/conf.d/mysql.ini` could look like this:  

    ```ini
    # configuration for PHP MySQL module
    extension=pdo_mysql.so
    [mysql]
    mysql.allow_local_infile=On
    mysql.allow_persistent=On
    mysql.cache_size=2000
    mysql.max_persistent=-1
    mysql.max_links=-1
    mysql.default_port=
    mysql.default_socket=/var/lib/mysql/mysql.sock  # Debian squeeze: /var/run/mysqld/mysqld.sock
    mysql.default_host=
    mysql.default_user=
    mysql.default_password=
    mysql.connect_timeout=60
    mysql.trace_mode=Off
    ```

The example shows [runtime configuration](https://www.php.net/manual/en/mysql.configuration.php) for [MySQL (Original)](https://www.php.net/manual/en/intro.mysql.php) which is removed since PHP 7.0

> This extension is deprecated as of PHP 5.5.0, and has been removed as of PHP 7.0.0. Instead, either the mysqli or PDO_MySQL extension should be used. See also the MySQL API Overview for further help while choosing a MySQL API.

Given that the required PHP runtime for Nextcloud is [PHP 7.2+]((https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html#server)) the docs should not reflect any possible configuration applicable to PHP version lower than 7.

### 1

At first the docs could just refer to applicable PHP.net entries, namely

* [Mysqlnd runtime configuration](https://www.php.net/manual/en/mysqlnd.config.php)

```ini
; configuration for php mysql module
; priority=10
extension=mysqlnd.so

; https://www.php.net/manual/en/mysqlnd.config.php
; mysqlnd.collect_statistics=1
; mysqlnd.collect_memory_statistics=0
; mysqlnd.debug=
; mysqlnd.log_mask=0
; mysqlnd.mempool_default_size=16000
; mysqlnd.net_read_timeout=86400
[...]

PR #6081, opened 2021-01-23
Update PHP setting recommendations for php7

Steve8291:
The recommended config for php.ini seems to be very outdated. This was mentioned in #4758
The user is told to install PDO_MySQL but the config given appears to be for an old version of the ext/mysql. It also contains comments `#` that are no longer allowed since php7.0
Also, at least in Ubuntu when you install the pdo_mysql it creates a working config with good defaults so it's confusing to a new user when reading the current document since they really wouldn't have to change anything on their setup.

### DIFF
--- a/admin_manual/configuration_database/linux_database_configuration.rst
+++ b/admin_manual/configuration_database/linux_database_configuration.rst
@@ -98,26 +98,24 @@ Please refer to the `page in the MySQL manual <https://mariadb.com/kb/en/library
 
 .. note:: MariaDB is backwards compatible with MySQL.  All instructions work for both. You will not need to replace mysql with anything.
 
-The PHP configuration in :file:`/etc/php7/conf.d/mysql.ini` could look like this:
+Enabling the pdo_mysql extension could be done in various ``php.ini`` files depending on your installation of PHP.
+This could be done within :file:`/etc/php/php7.2/fpm/php.ini` file if you are using php-fpm.
+On Ubuntu you might find :file:`/etc/php/php7.2/mods-available/pdo_mysql.ini` with the following:
 
 ::
 
-  # configuration for PHP MySQL module
+  ; configuration for php mysql module
+  ; priority=20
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_database/linux_database_configuration.rst
```diff
@@ -98,26 +98,24 @@ Please refer to the `page in the MySQL manual <https://mariadb.com/kb/en/library
 
 .. note:: MariaDB is backwards compatible with MySQL.  All instructions work for both. You will not need to replace mysql with anything.
 
-The PHP configuration in :file:`/etc/php7/conf.d/mysql.ini` could look like this:
+Enabling the pdo_mysql extension could be done in various ``php.ini`` files depending on your installation of PHP.
+This could be done within :file:`/etc/php/php7.2/fpm/php.ini` file if you are using php-fpm.
```
tflidd, line 102: it's not so much if it is ubuntu, rather than php-fpm or apache module.

tflidd (commented):
The parts you remove, they are also removed on modern linux default-php.ini files? I mostly run with default configs and just adjusted a few parameters.
If we want to do things differently, we have to be careful. If users are running other php applications, we then might have a specific php.ini adjusted for Nextcloud (if this is really required).

admin_manual/configuration_database/linux_database_configuration.rst
Steve8291: Yes the main difference would be between php-fpm and the apache module. Since most install with a working config it seems confusing to make suggestions to modify the configs. What would you suggest doing?

tflidd: If we can use default configs in general, we should just focus on things that must or should be changed. Either they must be changed, or they allow a better performance.

kesselb:

I am closing the pull request because there has been no activity for a while now.

PR #6094, opened 2021-01-31
Adds warning about user deletion and object store

pierreozoux:
Documents bug https://github.com/nextcloud/server/issues/9690


nickvergessen (approved):

pierreozoux:
@nickvergessen I thought it was now fixed:
https://github.com/nextcloud/server/pull/26792

If it is fixed, we should revert this commit.

PR #6311, opened 2021-03-23
Added info about webfinger and nodeinfo

Turbulentarius:
I am not sure what exactly webfinger and nodeinfo is, but at least these redirects seem to fix the:

Your web server is not properly set up to resolve “/.well-known/webfinger”...
Your web server is not properly set up to resolve “/.well-known/nodeinfo”...

Errors that many users seem to get when running Nextcloud in a subfolder.


rolandu:
Thanks, this helped me get rid of the warning, but I think some more info would be needed about what webfinger and nodeinfo actually are.

susnux:
Thank you for your pull request! Meanwhile this is included in the documentation.
But thanks again and sorry for the long delay!

PR #6326, opened 2021-03-27
Fix password policy documentation

### LINKED ISSUE #5858: Minimum password length should be consistent
in the manual text and the corresponding screenshot
https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_user/user_password_policy.rst

The current text just says 10 characters, while the app itself says 8 characters.

Also concerns https://github.com/nextcloud/password_policy/blob/master/README.md
and https://github.com/nextcloud/password_policy/blob/19a4714a95830f3b151872925962da9195c3a3cc/lib/PasswordPolicyConfig.php#L56 in the password policy app.

Is there a consensus in the team about which value should be suggested as a placeholder? 

And should the manual then be adjusted just for Nextcloud 21 or for all Nextcloud versions that are currently still maintained?

Chartman123:
Fixes #5858 

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de>


jotoeri:
👍 It's actually this since NC19 (and partially even longer).
@MorrisJobke I don't know the workflow here - do you backport everything that is right on this repo or just important stuff?

(Concerning the image: The small perfectionist inside me rebels a bit, as havibeenpwned is disabled by default. So i'd include an image of the default-settings. But that should be minor thing. 😉 )

jotoeri (approved):

Chartman123:
Oh... I had checked the defaults in the code before I took the screenshot but must have overlooked that 😆

jotoeri:
Yeah it's a bit hidden below all the setters and written a bit differently. 😉 
If it's no big work for you, then yes, update would be nice imo. 👍

Chartman123:
Could you please check again if it's correct now? 🙈

jotoeri:
Looks good to me 👍 

Can you squash the two commits into one? Just rebase interactively `git rebase -i master` and replace the `pick` on the second commit by `f` or `fixup`. 😉

Chartman123:
I don't know if this worked the way you wanted?

jotoeri:
Ahh, nearly... 😉 
Looks like you did it right, but then did a `git pull` as you could not push?

As a squashed commit is not straight onto the old history (but basically a new squashed commit) you need to do a `git push -f`. This `-f` (force-push) overwrites an old history, so one needs to be a bit careful with that, but it is necessary to update an edited history.

So now to fix that:
- We have now four commits in the history, of which we only want to use the third one, as it contains the desired code
- Do another interactive rebase `git rebase -i`
- Remove all lines except the one with the third commit (Hash should be `6901b65`)
- Ideally shortly check the log `git log`, you should now have only the single commit based onto the master-branch.
- Then force-push `git push -f`

Chartman123:
I think that it worked now. I also managed to GPG sign the new commit by adding an extra line to the rebase file 👍🏻

jotoeri:
Yeah, nice! 🚀 
Want to push the green button yourself? 😉 
Morris can then still decide afterwards how far to backport 👍

jotoeri (approved):

Chartman123:
I can't merge the PR as I don't have write access to this repository...

jotoeri:
Ahh, that's a pity. 🙁

Chartman123:
No problem 🙂 Thank you for your help!

PR #6333, opened 2021-04-01
Document primary object storage breaking SSE

mashedkeyboard:
Server-side encryption breaking when using object storage as a primary storage backend is significant, and people should be aware of it when reading documentation rather than running into the issue after having already set up Nextcloud.

All this does is document the existence of issue https://github.com/nextcloud/server/issues/22077.

### DIFF
--- a/admin_manual/configuration_files/primary_storage.rst
+++ b/admin_manual/configuration_files/primary_storage.rst
@@ -27,6 +27,9 @@ Because of this primary object stores usually perform better than when using the
 object store as external storage but it restricts being able to access the files from
 outside of Nextcloud.
 
+.. warning:: Using an object store as primary storage will mean you are `unable to use
+	Nextcloud's server-side encryption functionality <https://github.com/nextcloud/server/issues/22077>`_.
+
 -------------
 Configuration
 -------------


tflidd (approved):

tflidd:
I would prefer the fix of the bug. However, it's been out there for at least 2-3 major versions and the server-side encryption is exactly for such a use case, so you don't store the keyfiles on the same storage as the encrypted data.

mashedkeyboard:
I'd also prefer the bug to be fixed, but unfortunately as you say this has been around for quite some time, and as noted in https://github.com/nextcloud/server/issues/22077#issuecomment-747411624, even my attempts to investigate options to pay Nextcloud GmbH for a fix have failed. Nobody seems to quite know how the server-side encryption code works, or have the time to reverse engineer it without the support of the Nextcloud GmbH engineers, unfortunately.

elsiehupp:
Thank you for taking the initiative with this!

I definitely support adding a warning, but I think I’d make it a little more explicit. Like, rather than:

> **Warning:** Using an object store as primary storage will mean you are unable to use Nextcloud's server-side encryption functionality.

I would make it something like:

> **Warning:** Using an object store as primary storage is incompatible with the default server-side encryption module. *Using both at the same time may cause files to become irretrievably corrupted.*

I would also add a corresponding warning to [`encryption_configuration.rst`](https://github.com/nextcloud/documentation/blob/5ba93468795510b7bb0fc70122e2b570646bcaf3/admin_manual/configuration_files/encryption_configuration.rst) with the two parts of sentence swapped, like so:

> **Warning:** The default server-side encryption module is incompatible with using an object store as primary storage. *Using both at the same time may cause files to become irretrievably corrupted.*

It may also be worth mentioning the possibility of using `s3backer` as a workaround, though it seems a bit beyond the scope to write an entire tutorial for how to do so. So maybe something like:

> As a workaround, the FUSE-based object-storage frontend [`s3backer`](https://github.com/archiecobbs/s3backer) incorporates encryption and is known to work with Nextcloud, but its use is not officially supported.

Obviously you can word any of these things slightly differently if you think it would be better.

I’m able to merge this, but I’ve been hesitant to use that power because I don’t actually work for Nextcloud, and I don’t want to step on people’s toes.

elsiehupp:
Oh, and, yes, it’s a good idea to link to the open issue, though you can decide what text to have as the hyperlink.

CarlSchwan (approved):

skjnldsv (approved):

PR #6408, opened 2021-04-18
fix HSTS header in nginx subdir configuration

fdiwald:
Headers defined in a specific section are overwritten if there are additional headers defined in a subsection. This is true for the header for HSTS (`Strict-Transport-Security`) in the `server` section which is overwritten by the headers in the `location ^~ /nextcloud` section. This is documented in the [nginx documentation](http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header).
This solution only sends the header for the nextcloud-subdir. I guess it would be more correct to set the header for all locations but I did not find an elegant solution to that problem. Maybe an nginx pro could come up with a better solution.
Signed-off-by: Florian Diwald <fdiwald@hotmail.com>


Tachi107:
> This solution only sends the header for the nextcloud-subdir. I guess it would be more correct to set the header for all locations but I did not find an elegant solution to that problem.

What about just adding `add_header` in both sections? May not be elegant, but an imperfect solution is better than no solution at all :)

fdiwald:
Correct. This is the consequence of my approach. If you add a directory (for instance `/collabora`) you should add the headers there too.

ChristophWurst:
Conflicts, unfortunately

fdiwald:
@ChristophWurst is there still something I have to fix or is this pull request ready for review? Honestly, I am a bit overwhelmed by the checks here. Not sure what they mean.

ChristophWurst:
It should be fine

ChristophWurst (approved):

PR #6470, opened 2021-04-28
Update tutorial with dev-setup alternative

schrotie:
Added a much simpler and less intrusive alternative development setup using podman or docker.


MorrisJobke (approved):
Fine with me 👍 

Podman is indeed nice 🚀

PR #6626, opened 2021-05-10
Fix documentation stable/latest intro

skjnldsv:
The intro was confusing compared to the urls we're using:
https://docs.nextcloud.com/server/latest/developer_manual/
https://docs.nextcloud.com/server/stable/developer_manual/

I edited that to properly reflect the diffs between latest/stable and last versions.

I also used the aliases `stable` and `latest` for the url as this is easier to change when a new release is pushed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -219,21 +219,33 @@
 								page for more information.</p>
 							</div>
 
-							<div class="section" id="nextcloud-21">
-								<h2>Nextcloud 21<a class="headerlink" href="#nextcloud-21" title="Permalink to this headline">¶</a></h2>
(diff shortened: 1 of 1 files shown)


artonge (approved):
Looks good to me.

To go even further, maybe it would be better to duplicate some reference to avoid erroneous reference after a release. This would look like this:

```
Nextcloud Latest --> latest
Nextcloud Stable --> stable
Nextcloud 22 --> 22
Nextcloud 21 --> 21
...
```

For example, after the release of 22, we only have to add a line for 23. But in the mean time, people can still access 23 with `latest` and 22 would not link to `latest`.

nickvergessen (approved):

PR #6711, opened 2021-05-20
corrects a sentence in search.rst

StCyr:


ChristophWurst (approved):

tflidd:
Best way would be if you could sign your commit, so we can merge directly....

PR #6720, opened 2021-05-24
Add a warning about limited support for NGINX

acsfer:
Warn user about limited support to webservers other than Apache.

### DIFF
--- a/admin_manual/installation/nginx.rst
+++ b/admin_manual/installation/nginx.rst
@@ -1,10 +1,14 @@
 ===================
-Nginx configuration
+NGINX configuration
 ===================
 
-This page covers example Nginx configurations to use with running a Nextcloud
-server. These configurations examples were originally provided by
-`@josh4trunks <https://github.com/josh4trunks>`_ and are community-maintained. (Thank you contributors!)
+.. warning::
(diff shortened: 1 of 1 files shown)


admin_manual/installation/nginx.rst
```diff
@@ -1,10 +1,14 @@
 ===================
-Nginx configuration
+NGINX configuration
 ===================
 
-This page covers example Nginx configurations to use with running a Nextcloud
-server. These configurations examples were originally provided by
-`@josh4trunks <https://github.com/josh4trunks>`_ and are community-maintained. (Thank you contributors!)
+.. warning::
```
josh4trunks, line 5: I'm just wondering why we have `.. subtitle::`?
I'm not familiar enough with RST format to know what that is trying to achieve?

josh4trunks (approved):
I just have 1 question, but it isn't really substantive. Change makes sense to me.

MorrisJobke (approved):
Thanks 👍

admin_manual/installation/nginx.rst
MorrisJobke: I don't see the subtitle in here. Or are you referring to another part of that document?

PR #6743, opened 2021-05-31
Document calendar embed parameters

jooola:
This describe additional parameters in order to embed a calendar using  iframes.

Closes https://github.com/nextcloud/calendar/issues/2938

### DIFF
--- a/user_manual/pim/calendar.rst
+++ b/user_manual/pim/calendar.rst
@@ -90,7 +90,14 @@ Calendars can be published through a public link to make them viewable (read-onl
 There's also an « embedding code » that provides an HTML iframe to embed your calendar into public pages.
 
 Multiple calendars can be shared together by adding their unique tokens to the end of an embed link. Individual tokens can be found at the end of each calendar's public link. The full address will look like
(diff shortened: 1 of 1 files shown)


tcitworld (approved):

jospoortvliet (approved):
Well, that's how we did it in our blog, so LGTM

jospoortvliet:
I guess we also would backport a bit but I'm not sure about the magic commands here so I leave that to somebody who knows more than I do 😄

jospoortvliet:
Hi Alex,

I don't read German very well, but in general if you have questions - you ask support or user questions on help.nextcloud.com. Github is for developers to work together in. As we also say in various places.

As an open source project, features get implemented if either somebody spends their free time doing them, or when somebody gets paid to do it (usually because a Nextcloud customer needs it). I'm afraid that beyond that, there's not much you can do.

But I understand you're looking for resource booking, which is now available in Calendar with Nextcloud Hub 22. I don't know the specifics, sadly - I have not seen it in action - but I presume that more UI work is coming. Perhaps you can help test.

PR #6802, opened 2021-06-15
Update example_ubuntu.rst

bruno666:
Php package shouldn't be installed with their version number.
The root user of mysql / mariadb is authenticated by unix account, not password.

### DIFF
--- a/admin_manual/installation/example_ubuntu.rst
+++ b/admin_manual/installation/example_ubuntu.rst
@@ -8,9 +8,9 @@ installation, using Apache and MariaDB, by issuing the following commands in a
 terminal::
 
     sudo apt update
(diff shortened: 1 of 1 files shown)


jospoortvliet (approved):
The rst is fine. I can't judge if this is true on Ubuntu as I don't use it! Anyone else could comment on that?

tflidd:
We have a follow up to make the documentation for Ubuntu 22.04. Perhaps you can review the topic and bring up your concerns by explicitely stating the php version number.
https://github.com/nextcloud/documentation/pull/8186

tflidd:
Thanks for your contribution to https://github.com/nextcloud/documentation/pull/8186 where everything was fixed and your remarks were considered.

PR #6872, opened 2021-06-17
Remove traces of migrations:generate-from-schema

PVince81:
Remove traces of occ migrations:generate-from-schema which was removed
in NC 22.

Also adjust text that specifies that now ALL migrations:* commands are only
available in debug mode.

### DIFF
--- a/developer_manual/app_publishing_maintenance/upgrade-guide.rst
+++ b/developer_manual/app_publishing_maintenance/upgrade-guide.rst
@@ -79,11 +79,6 @@ The original `PSR-0` standard was deprecated in 2014 and therefore the support f
 .. _`PSR-0`: https://www.php-fig.org/psr/psr-0/
 .. _`PSR-4`: https://www.php-fig.org/psr/psr-4/
 
-Last version with database.xml support and migration
-****************************************************
-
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):

developer_manual/app_publishing_maintenance/upgrade-guide.rst
```diff
@@ -79,11 +79,6 @@ The original `PSR-0` standard was deprecated in 2014 and therefore the support f
 .. _`PSR-0`: https://www.php-fig.org/psr/psr-0/
 .. _`PSR-4`: https://www.php-fig.org/psr/psr-4/
 
-Last version with database.xml support and migration
```
ChristophWurst, line 82: could you please revert this change? This is the upgrade section for people who update their app for Nextcloud 21. This information is still true. Only in 22 this was dropped.

ChristophWurst: instead open a new section for 22 and document that this is now gone.

PVince81: can do, but I'll have to remove the link because the section about the occ command has been removed.
not sure if we should create a link between different versions of the documentation ?!

ChristophWurst: Just drop the link. Developers should be able to figure out the rest :)

PVince81: PR here: https://github.com/nextcloud/documentation/pull/6874

PR #6875, opened 2021-06-20
Update nginx configuration

nnhien:
Add `location` directive for generic `/.well-known` handling

Remove `rewrite` directives because `redirect` option returns a 302 status code, which is inconsistent.


tflidd:
It's going the other way around to use specific redirects:
https://github.com/nextcloud/documentation/pull/8007

PR #6876, opened 2021-06-20
Add dev documentation for workflow operations

cwilby:
I wrote this while learning about the workflow engine so I could refactor a media converter I was building to use that rather than timed background jobs and a bunch of config JSON nonsense.

I saw that documentation was needed, so I watched/read through blizzz's talk and read the code for the workflow engine, and wrote this up to help build my own understanding.

### DIFF
--- a/developer_manual/digging_deeper/flow.rst
+++ b/developer_manual/digging_deeper/flow.rst
@@ -2,8 +2,154 @@
 Nextcloud Flow
 ==============
 
-Nextcloud Flow is the workflow engine of Nextcloud. It offers flexible, user-defined event-based task automation. Apps can register events or triggers which can start a flow, as well as actions which get executed once a trigger is hit and the constraints are satisfied.
+Nextcloud Flow is a flexible, user-defined event-based workflow engine. 
+
+The workflow engine broadcasts events when certain trigger conditions are met.
+
+Applications register event listeners with the workflow engine to receive events broadcast by the workflow engine when one or more trigger conditions are met.
+
+Event listeners then register operations to handle those events.
+
+An example flow might be converting Word documents into PDF when they are added to one folder, then move that Word document to another folder once it's finished.
+
+The following sections define the individual components your application will need to integrate with the workflow engine.
 
 At 36c3 blizzz gave a talk explaining Flow and `how to write actions and triggers. <https://mirror.eu.oneandone.net/projects/media.ccc.de/congress/2019/h264-sd/36c3-oio-174-eng-Building_Nextcloud_Flow_sd.mp4>`_ You can `find the slides of their talk here. <https://nextcloud.com/wp-content/themes/next/assets/files/Building_nextcloud_flow.pdf>`_
 
-Contributions to this documentation are, as always, very welcome!
+An example workflow can be found at https://github.com/nextcloud/workflow_pdf_converter.
+
+Events
+======
+
+First, you should consider the events you want users to configure in your flow.  Nextcloud currently dispatches the following events to the workflow engine:
+
+Calendar
+--------
+
+- ``\OCA\DAV\CalDAV\CalDavBackend::createCalendar``
+- ``\OCA\DAV\CalDAV\CalDavBackend::updateCalendar``
+- ``\OCA\DAV\CalDAV\CalDavBackend::deleteCalendar``
+- ``\OCA\DAV\CalDAV\CalDavBackend::createCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::createCachedCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::updateCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::updateCachedCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::deleteCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::deleteCachedCalendarObject``
+- ``\OCA\DAV\CalDAV\CalDavBackend::createSubscription``
+- ``\OCA\DAV\CalDAV\CalDavBackend::updateSubscription``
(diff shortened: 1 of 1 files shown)


cwilby:
I do not know what to do about any of the above 👆

cwilby:
I'll likely come back and edit this once I finish with the app.

developer_manual/digging_deeper/flow.rst
```diff
@@ -2,8 +2,154 @@
...
+
+Events
+======
+
+First, you should consider the events you want users to configure in your flow.  Nextcloud currently dispatches the following events to the workflow engine:
+
+Calendar
+--------
+
+- ``\OCA\DAV\CalDAV\CalDavBackend::createCalendar``
```
ChristophWurst, line 29: \OCA\* is an app namespace. It's not public api and therefore shouldn't be documented/recommended as such

marcelklehr: I tend to disagree. Since the workflow engine exposes these events (if it really does, no idea if that's correct), then the assumption is that apps may want to be able to handle those events in an operation.

ChristophWurst: If WFE exposes the events for other apps they should be in OCP

marcelklehr: And what about third party apps, like Bookmarks? They expose events via WFE as well, should they all be in OCP?

marcelklehr: I would disagree, but it's not my decision :D :shrug:

ChristophWurst: looping in other lead developers @juliushaertl @nickvergessen. wdyt?^

nickvergessen: General remark:

> If WFE exposes the events for other apps they should be in OCP

WFE might be a different thing here, as it's shipped.
But I'm definetly not going to send 80+ Events we have in Talk as OCP to the server. That's way to slow development cycle and I'm not going to add a temporary internal one until a public OCP one was merged.
So no, from my POV all events that are documented intentionally by the Devs are okay to be considered public API. For Talk that is: https://nextcloud-talk.readthedocs.io/en/latest/events/

---

Regarding the case here:

> Nextcloud currently dispatches the following events to the workflow engine:

I'm not sure where this though it coming from. I'm not aware where this list would origin from. There are public events available to register entities, operations and checks: https://github.com/nextcloud/server/blob/aa039c986a3e8aaa97569d65ff1afd47a9b94a53/apps/workflowengine/lib/Manager.php#L685-L710

But I'm not aware dav would register calendars, subscriptions, etc. for those. Neither shares nor users.
The only known entity is files with those events:
https://github.com/nextcloud/server/blob/7d7ef1d2e68310cc8e0988a926406cddd241c257/apps/workflowengine/lib/Entity/File.php#L111-L117

juliushaertl: I'd also agree that apps should still be able to use their own Events when considering them as public APIs.

Regarding the list I can only imagine that they are examples since the workflow engine would add listeners dynamically depending on the events that are configured, but by this depends on the actual Entity classes that are registered from apps https://github.com/nextcloud/server/blob/efb7448cd271f7f0cd21fd284fd1d3f9d729f20f/apps/workflowengine/lib/AppInfo/Application.php#L73-L78

juliushaertl: Still, for things like Calendar (or anything shipped) we should have those rather in OCP.

cwilby:
New version ready for review, I limited the events to only File events after reviewing the WFE manager class, and made some readability / grammar changes.

developer_manual/digging_deeper/flow.rst
```diff
@@ -2,8 +2,87 @@
...
+            RegisterFlowOperationsListener::class
+        );
+    }
+
+Operations
+==========
+
+Operations are classes registered with one or more event listeners.
+
+This class must implement one of the following `IOperation <https://github.com/nextcloud/server/blob/master/lib/public/WorkflowEngine/IOperation.php#L33>`_ interfaces:
```
nickvergessen, line 51: They don't have to. Talk is implementing `IOperation` directly. Works as well. Only when you want to have the other things you need to implement those.

developer_manual/digging_deeper/flow.rst
```diff
@@ -2,8 +2,87 @@
...
+    module.exports = webpackConfig
+
+Then, create ``src/main.js`` with the following contents::
+
+    import ConvertToPdf from './ConvertToPdf'
+
+    OCA.WorkflowEngine.registerOperator({
+        id: 'OCA\\WorkflowPDFConverter\\Operation',
+        operation: 'keep;preserve',
+        options: ConvertToPdf,
```
nickvergessen, line 84: This is "legacy". going forward element should be used and allows apps to provide their content isolated, so it does not cause conflicts when the app and server use different Vue versions, etc. See the following PR and from there linked PRs and issues:
https://github.com/nextcloud/spreed/pull/14367

nickvergessen (changes requested):
Otherwise looks good. I'm just not a fan of linking to files (and lines) from master branch, because those links might shift. Instead you could link against the lines with the commit hash, so instead of
https://github.com/nextcloud/server/blob/master/lib/public/WorkflowEngine/IOperation.php#L33
use:
https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/public/WorkflowEngine/IOperation.php

developer_manual/digging_deeper/flow.rst
nickvergessen: Also the line looks unrelated here?

skjnldsv:
As there is no feedback since a while I will close this ticket. If you still want this, please make sure to rebase to the latest master. After that, feel free to reopen and ping me :)

PR #6962, opened 2021-07-14
Complete Windows Requirements

github-tomster:
@thosetryingtocode
if you follow the microsoft SAC way with your code it has to be clear in the docs.


github-tomster:
follow here
https://github.com/nextcloud/desktop/pull/3546
it is double-documentation somehow... 🤷‍♂️

tflidd:
Yes, it's better to take it from the desktop's requirements:
https://docs.nextcloud.com/desktop/3.2/installing.html#system-requirements

Not sure if it is better to copy or to link directly.

FlexW:
I think linking would be better. I just see how we forget to update this.

PR #6967, opened 2021-07-15
Desktop client minimum supported Windows version is 8.1+.

allexzander:
Signed-off-by: allexzander <blackslayer4@gmail.com>


allexzander:
@jospoortvliet We need this changed everywhere on the website too

FlexW:
Maybe we can instead just link to the client manual. Would be easier than to update the system requirements in the future.

tobiasKaminsky (approved):

PR #7035, opened 2021-08-13
Add docs for flow

marcelklehr:
This has bugged me for so long that I finally did it myself.

This is a first pass, suggestions for improvement welcome.

### DIFF
--- a/developer_manual/digging_deeper/flow.rst
+++ b/developer_manual/digging_deeper/flow.rst
@@ -4,7 +4,7 @@ Nextcloud Flow
 
 Nextcloud Flow is a user-defined event-based workflow engine.
 
-Applications can expose entities with specific events on the one hand, and operations that act on these events on the other hand.
+Applications can expose entities with specific events on the one hand, and operations that act on these events on the other hand. A user defines a workflow rule that is triggered by a certain event which will then execute the operation based on user defined checks.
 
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Looks good but I'm not familiar with this subsystem

developer_manual/digging_deeper/flow.rst
```diff
@@ -2,8 +2,53 @@
 Nextcloud Flow
 ==============
 
-Nextcloud Flow is the workflow engine of Nextcloud. It offers flexible, user-defined event-based task automation. Apps can register events or triggers which can start a flow, as well as actions which get executed once a trigger is hit and the constraints are satisfied.
+Nextcloud Flow is a user-defined event-based workflow engine.
+
+Applications can expose entities with specific events on the one hand, and operations that act on these events on the other hand.
```
juliushaertl, line 7:
```suggestion
Applications can expose entities with specific events on the one hand, and operations that act on these events on the other hand. A user defines a workflow rule that is triggered by a certain event which will then execute the operation based on user defined checks.
```

juliushaertl (commented):
Small addition but seems like a good first start for this. Thanks a lot for that :tada:

juliushaertl (approved):

PR #7037, opened 2021-08-13
Document correct db transaction handling

ChristophWurst:
Bildschirmfoto von 2021-08-13 11-27-21

### DIFF
--- a/developer_manual/basics/storage/database.rst
+++ b/developer_manual/basics/storage/database.rst
@@ -45,6 +45,33 @@ Inside your database layer class you can now start running queries like:
     }
 
 
+Transactions
+------------
+
+Database operations can be run in a transaction to commit or roll back a group of changes in an atomic fashion.
+
+.. code-block:: php
+
+    <?php
+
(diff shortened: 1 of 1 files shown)


developer_manual/basics/storage/database.rst
```diff
@@ -45,6 +45,33 @@ Inside your database layer class you can now start running queries like:
     }
 
 
+Transactions
+------------
+
+Database operation can be run in a nested transaction to commit or roll back a group of changes in an atomic fashion.
```
marcelklehr, line 51: operation**s**

What is the significance of *nested*? Can I put a transaction into another transaction?

ChristophWurst: > Can I put a transaction into another transaction?

Yes. You can have multiple levels.

ChristophWurst: But I will drop this. It's unnecessarily confusing :)

marcelklehr (approved):
:+1:

PR #7045, opened 2021-08-15
Update sync_ios.rst

Schievel1:
just providing the name of the domain wont let you in on an iphone. It will give errors like "SSL not possible, would you like to try without SSL?" (It always gives this error for some reason, even if you provide it with an nonsense URL it can't even resolve.)

### DIFF
--- a/user_manual/pim/sync_ios.rst
+++ b/user_manual/pim/sync_ios.rst
@@ -11,7 +11,8 @@ Calendar
 #. Select Add Account.
 #. Select Other as account type.
 #. Select Add CalDAV account.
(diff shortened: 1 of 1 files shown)


tflidd:
If you have set up the `.well-known/` redirects correctly, ios should find the proper URL itself.... or doesn't this work any more. Configuration is much easier like that.

BernieO:
> If you have set up the .well-known/ redirects correctly, ios should find the proper URL itself.... or doesn't this work any more. Configuration is much easier like that.

I can confirm that service discovery is still working as it should.

PR #7113, opened 2021-08-26
fix warnings on webfinger/nodeinfo on OpenBSD

nabbisen:
As to the latest package of NextCloud on OpenBSD, this commit mitigates warnings on webfinger and nodeinfo.

Note:
In order to follow the current official Nginx configuration strictly, it may be necessary to include consideration on `/.well-known/acme-challenge` and `/.well-known/pki-validation` , although it had nothing to do with my usage:
https://docs.nextcloud.com/server/stable/admin_manual/installation/nginx.html?highlight=nginx

Screenshot_2021-08-26_12-48-04

### DIFF
--- a/admin_manual/installation/example_openbsd.rst
+++ b/admin_manual/installation/example_openbsd.rst
@@ -88,7 +88,11 @@ Create a virtualhost in ``/etc/httpd.conf`` and add the following content to it:
 	  }
 
 	  location "/.well-known/webfinger" {
-		  block return 301 "https://$SERVER_NAME/nextcloud/public.php?service=webfinger"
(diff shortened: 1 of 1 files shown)


tflidd:
There are some open pull requests regarding this topic and nginx:
https://github.com/nextcloud/documentation/pull/8007

nabbisen:
@tflidd 
Thank you for your information.
I read #8007 and agree that:

> Webfinger and Nodeinfo should be forwarded to index.php.

The OpenBSD example, however, seems to still use `public.php` due to:

https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/example_openbsd.rst

PR #7122, opened 2021-08-31
Fix file picker html templates not served [nginx]

vfauth:
Based on [this issue](https://github.com/nextcloud/server/issues/28200).


susnux:
Thank you very much.
The Filepicker was refactored in Nextcloud 27 to use vue, so no longer needed.

PR #7124, opened 2021-09-02
typo: update formatting and correct capitalization

SethFalco:
I was just going through the documentation, so I could try to set up Nextcloud development through Docker.
This just makes some minor edits to improve the documentation.

### Alternate Setups
* Splits the heading from the list of setups.
* Formats them in code blocks, before they displayed invalid commands. (ie `--` became `—`)
* Removes `--replace` argument on Docker, this is not a valid argument for the `docker run` command.

### Links to nextcloud.org
There are links to nextcloud.org scattered around the documentation, I only updated one of them to nextcloud.com because the page name and content is literally the same.

There are however some dead links remaining that I haven't addressed. Someone familiar with the documentation and where resources should be linked may want to `CTRL` + `SHIFT` + `F` and check for all instances of `nextcloud.org`.

### Noun Capitalization
Github → GitHub

### .gitignore
Just ignores Visual Studio Code files from being committed to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,8 @@ developer_manual/design/icons.txt
 venv
 
 # VScode
-.vscode
+.vscode/
+*.code-workspace
+
+# Local History for Visual Studio Code
+.history/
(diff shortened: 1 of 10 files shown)


kesselb (approved):
Thanks :+1:

Issue #6157, opened 2021-02-09
[NC 21 RC1] No info about webfinger and nodeinfo settings


tflidd:
On the settings, I get these warnings:
![image](https://user-images.githubusercontent.com/8111200/107366380-61248780-6ade-11eb-8664-1c8b69766fe0.png)

The given link redirects me to https://docs.nextcloud.com/server/21/admin_manual/issues/general_troubleshooting.html#service-discovery, however there is only information about `.well-known/carddav` and `.well-known/caldav`, nothing about `webfinger` and `nodeinfo`.

Since I am using nginx, I was also checking the example config: https://docs.nextcloud.com/server/21/admin_manual/installation/nginx.html?highlight=nginx But nothing about `webfinger` and `nodeinfo` on a first sight. However, there were some changes: https://github.com/nextcloud/documentation/pull/5825/files

I suggest to change the landing page by mentioning the webfinger/nodeinfo being generally handled by an updated .htaccess and that nginx users might check their configuration.

NC version: NC21 RC1
webserver: nginx

tiiiecherle:
I stumbled accross the same problem after upgrading to NC21 using nginx.
I solved it by adding this to the nginx config:

```
        location /.well-known {       
            location = /.well-known/carddav   { return 301 /nextcloud/remote.php/dav/; }
            location = /.well-known/caldav    { return 301 /nextcloud/remote.php/dav/; }
            
            # according to the documentation these two lines are not necessary, but version 21.0.0 will produce warnings in the overview setup check
            location = /.well-known/webfinger   { return 301 /nextcloud/index.php$uri; }
            location = /.well-known/nodeinfo   { return 301 /nextcloud/index.php$uri; }
            
            # anything else is dynamically handled by Nextcloud
            location ^~ /.well-known          { return 301 /nextcloud/index.php$uri; }
    
            try_files $uri $uri/ =404;
        }
```

Lawkss:
I had the same problem on apache. I had to add these lines to htaccess make the warnings go away:

RewriteRule ^\.well-known/webfinger /YOURSUBFOLDER/index.php/.well-known/webfinger [R=301,L]
RewriteRule ^\.well-known/nodeinfo /YOURSUBFOLDER/index.php/.well-known/nodeinfo [R=301,L]

The documentation only mentions carddav and caldav but not webfinger and nodeinfo.

DerVerruckteFuchs:
[...]

PR #7130, opened 2021-09-13
Add nginx documentation about setting for fix warning of webfinger and nodeinfo.

ynott:
This fixes nextcloud#6157.

Add a description of the settings to be added to nginx's nextcloud.conf to turn off the following warnings in nginx.

Your web server is not properly set up to resolve “/.well-known/webfinger”
Your web server is not properly set up to resolve “/.well-known/nodeinfo”


mddvul22:
How does this fix #6157 if it only addresses documentation for Nginx?  This bug happens in Apache, also.  Also, I don't understand how the lack of webfinger and nodeinfo well-known urls should be generating a warning, if they aren't used by any applications in a Nextcloud configuration.

danger89:
Hello Nextcloud people? Why not solve this **Nginx** documentation page regarding this issue now??

image

PR #7142, opened 2021-09-25
selinux: All directory labels should be deleted on uninstall

pedrxd:
The command for remove the directory label of the log of aws-sdk has been added.


kesselb (approved):
Thanks :+1:

PR #7159, opened 2021-09-30
Add documentation for admin right privilege setting

### LINKED ISSUE #7158: Admin settings delegation
We need docs for https://github.com/nextcloud/server/pull/28189

CarlSchwan:


ChristophWurst:
> Fix #7158

For that ticket I was actually more thinking about documenting how developes can adapt to this. But having admin dosc is also great :) Thanks :)

ChristophWurst (approved):

CarlSchwan:
> For that ticket I was actually more thinking about documenting how developes can adapt to this. But having admin dosc is also great :) Thanks :)

Oh will do that then next :)

PR #7170, opened 2021-10-05
Update the 'Migrating' page to include optional 'maintenance:data-fingerprint' command step.

allexzander:
Signed-off-by: alex-z <blackslayer4@gmail.com>


jospoortvliet (approved):
Looks good to me!

jospoortvliet:
> does this not apply to older releases? I think we should backport

pretty sure, yes. But I don't know how to do backports, well, something with /backport or so, but 🙈

@allexzander this is relevant for all current releases, right?

ChristophWurst:
This has to be done manually

PR #7355, opened 2021-10-17
Added steps to enable mod_env and mod_rewrite

2br-2b:
Added steps to enable mod_env and mod_rewrite (for people like me who didn't realize how to do it without searching it).


2br-2b:
Actually, this only applies to Apache. Should I leave this open or close it?

PR #7383, opened 2021-10-25
Lightning add-on no longer needed

always-tinkering:
Removed the first step of the recommended method which was to install the Lightning add-on.


ChristophWurst:
see https://github.com/nextcloud/documentation/pull/7574 and https://github.com/nextcloud/documentation/pull/7575

nextcloud has stable branches for each major release. this change is relevant to all because someone who uses nextcloud 21 will likely still use a recent version of thunderbird. hence I let the bot copy the change to older release docs as well.

PR #7534, opened 2021-11-06
Added example for WebDAV search by last updated time

cogk:


developer_manual/client_apis/WebDAV/search.rst
```diff
@@ -212,4 +212,35 @@ Get all png and jpg files over 10MB.
         </d:basicsearch>
     </d:searchrequest>
 
+Get all files last modified after a given date.
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="UTF-8"?>
+    <d:searchrequest xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
```
nickvergessen, line 220:
```suggestion
    <d:searchrequest xmlns:d="DAV:">
```

Should work too?

nickvergessen:
Doesn't seem to work for me on Nextcloud 22, 23 and 24:

```xml
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotImplemented</s:exception>
  <s:message>There was no plugin in the system that was willing to handle this SEARCH method.</s:message>
</d:error>
```

I just copied the curl from above the xml and replaced the body. Is that wrong?

nickvergessen (approved):

PR #7576, opened 2021-11-15
Document how to override the log level with occ

ChristophWurst:
This works thanks to https://github.com/nextcloud/server/pull/3966


PVince81 (approved):
:+1: brilliant

PR #7604, opened 2021-11-22
Fix dev manual links to the app controllers page

ChristophWurst:
Regression of https://github.com/nextcloud/documentation/pull/2165

Found in https://help.nextcloud.com/t/possible-annotations-for-controller-methods/127597

### DIFF
--- a/developer_manual/app_development/tutorial.rst
+++ b/developer_manual/app_development/tutorial.rst
@@ -83,7 +83,7 @@ On the client side we can call these URLs with the following jQuery code:
         // handle failure
     });
 
-On the server side we need to register a callback that is executed once the request comes in. The callback itself will be a method on a :doc:`controller <requests/controllers>` and the controller will be connected to the URL with a :doc:`route <requests/routes>`. The controller and route for the page are already set up in **notestutorial/appinfo/routes.php**:
+On the server side we need to register a callback that is executed once the request comes in. The callback itself will be a method on a :doc:`controller <../basics/controllers>` and the controller will be connected to the URL with a :doc:`route <../basics/controllers>`. The controller and route for the page are already set up in **notestutorial/appinfo/routes.php**:
 
 .. code-block:: php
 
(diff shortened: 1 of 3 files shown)


christianlupus:
I saw some more bugs in this. I did not yet finish, please be patient one or two days more. Thanks.

christianlupus:
I removed quite some more broken links in the RST files. Honestly, I am not so much into this file type, so I made it a best-effort approach. I hope, I got the context always correct.

I added @ChristophWurst as a reviewer as I  changed a few more things in the documentation in the branch. That should be somehow confirmed as I feel.

There are still some warnings during the build (see e.g. [here](https://github.com/nextcloud/documentation/runs/4312786962?check_suite_focus=true#step:4:172)). @ChristophWurst does it make sense to file an issue (here or against the sphinx builder in the actions) to look out in the logs of the actions for warnings and trigger a warning her in GH? That way, at least there is a good chance a change will not introduce regressions. What do you think?

I had to replace some links from ``:doc:`<file/name>` `` by `` `Text <file/name.html#anchor>`__``. I hope this will work with readthedocs. Experience is highly anticipated.

christianlupus (approved):
From my side, this can be merged.

I have made some changes to the PR so at least someone should have a look on this as well if I missed the target completely.

ChristophWurst:
> There are still some warnings during the build (see e.g. [here](https://github.com/nextcloud/documentation/runs/4312786962?check_suite_focus=true#step:4:172)). @ChristophWurst does it make sense to file an issue (here or against the sphinx builder in the actions) to look out in the logs of the actions for warnings and trigger a warning her in GH? That way, at least there is a good chance a change will not introduce regressions. What do you think?

That would be super helpful. Please open a ticket.

ChristophWurst:
IMO the links are so fragile because we use paths. Ideally there would be more unique labels that can be used for cross linking because then moving around files won't cause as much damage :)

Issue #7637, opened 2021-11-25
Get rid of existing sphinx erors/warnings


christianlupus:
When building the documentation in the GitHub actions, some errors in the existing code are rosen. These should be removed to avoid cluttering the annotations of any PR involved.

christianlupus:
@rakekniven I think you are one of the admins of the transifex integration, am I right?

There are some strings badly translated causing the documentation to break (partially). Do you have a good idea, how one could tackle this problem in general?

I could create a list of languages and files where there are issues (although this will take some time). I am not able to read all languages involved. Thus, I cannot solve this all on myself. After #7646 is merged there  are only a hand full of problems left.

rakekniven:
Hello @christianlupus I am just a volunteer spending time for NC project.

I am not into this sphinx thing. But I will spend an eye on it the next days.

christianlupus:
> Hello @christianlupus I am just a volunteer spending time for NC project.

Me too. I just saw you a moderator on the transifex platform although I did not find a quick way to contact you there, @rakekniven.

> I am not into this sphinx thing. But I will spend an eye on it the next days.

It is more a thing of the correct syntax of the restructured text files. If the syntax is bad, links are not generated but just an arbitrary (to the reader) text is displayed. I was just wondering how we can get rid of these _bad_ translations in favor of syntactically correct ones. My thought was to extract said list, have a look at which translations are an issue and then try to recover. Most probably, the punctuation is wrong or some technical words have been translated that need to keep in English (link names etc). 

I can look at the translations and tell you what is wrong. I can't change that in transifex. The question is: Could you alter the relevant strings in transifex? Then we would soon be done, I think.

rakekniven:
> The question is: Could you alter the relevant strings in transifex?

Yes, for sure.

> I can't change that in transifex.

You can get the necessary permissions to do that. Help is welcome.

christianlupus:
Hello, @rakekniven. I found issues in the following languages: `br`, `da`, `de`, `el`, `es`, `fr`, `gl`, `it`, `nl`, `nn`, `pl`, `pt_BR`, `tr`, `zh_CN`, and `zh_HK`.

In transifex, I have access to `de`, `es`, `fr`, `pl`, and `nl`. I will provide updates to these languages soon.
How can I obtain access to the other languages? Or would you prefer me to send the changes to you in some way?

christianlupus:
[...]

PR #7646, opened 2021-11-26
Remove some errors in the build process

christianlupus:
This is a first step to reduce the number of build errors/warnings and thereby GitHub annotations (#7637).

It should remove any syntactical error in the existing code. However, the translations of the user documentation are handled in transifex. Thus, the annotations introduced by bad translations are not recovered/tackled.


christianlupus:
Sorry for the many force-pushes. They were due to the fact that I had to tweak the workflow files.

@ChristophWurst I am not sure, who might be a good choice for reviewing this PR. Maybe you have a good suggestion or are able to do this yourself?

christianlupus:
This PR is superseded by the mentioned PRs for simpler review.

PR #7764, opened 2021-12-05
Removed fixed ROW_FORMAT for MariaDB 10.6

obel1x:
As statet in https://mariadb.com/kb/en/innodb-compressed-row-format/, the ROW_FORMAT = compressed is made read-only in MariaDB 10.6 and will mostly be removed in MariaDB 10.7. 
I made a pull request in nextcloud/server#30101 for removing the default fixed ROW_FORMAT in code and pulling this to the docs.


ChristophWurst:
https://github.com/nextcloud/server/pull/30129 is the new pr

PR #7795, opened 2021-12-07
Update section on `config_is_read_only` to reflect its changed meaning

mejo-:
This depends on https://github.com/nextcloud/server/pull/30130.

Signed-off-by: Jonas Meurer <jonas@freesources.org>


nickvergessen:
I think this file is autogenerated. Adjusting it in config/config.sample.php should be enough

mejo-:
> I think this file is autogenerated. Adjusting it in config/config.sample.php should be enough

Ah, didn't know that. Thanks for the clarification. Added the document changes directly to `config/config.sample.php` in https://github.com/nextcloud/server/pull/30130 now.

PR #7827, opened 2021-12-16
Removed obsolete tokens

christianlupus:
Currently, there are some obsolete parameters in the workflow files. This should remove some error messages from the workflows annotations/checks. This is part of #7646.


ChristophWurst (approved):

ChristophWurst:
I guess we should backport this?

christianlupus:
@ChristophWurst I think this might be a good idea as it is an issue with the GitHub action and not with the version of NC.

PR #7828, opened 2021-12-16
Added dedicated build step to build only the English version.

christianlupus:
This should allow us to always see the errors in the original files, even if translations rise more errors than allowed.

Currently, a GitHub workflow can only add 10 annotations per step. When the different languages have individual issues/errors, the basic error in the underlying English original text might get missed. This is critical as all translations will inherit syntax from that files.

By adding a dedicated English build, there are 10 annotations for the English build reserved.

This is part of #7646


ChristophWurst (approved):

ChristophWurst:
conflicts :grimacing:

PR #7829, opened 2021-12-16
Update path of icon images to match new file structure

christianlupus:
The location of the images has changed according to [the codes](https://github.com/nextcloud/documentation/blob/859f105da94bb8e8e5e2d0975de0d0e1e3b65336/developer_manual/html_css_design/icons.rst?plain=1#L15). This should correct the path in general to avoid wrong paths.

This does however **not** solve the problem, that the icons are not correctly generated. I will tackle this part in a separate PR.

This is part of #7646.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,8 @@ _build
 *.snag
 
 # Icons files
-developer_manual/design/img
-developer_manual/design/icons.txt
+developer_manual/html_css_design/img
+developer_manual/html_css_design/icons.txt
 
 # Exclude Eclipse project
 .project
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):
Awesome

ChristophWurst:
do we need a backport?

christianlupus:
Only in `stable23` the path is changed. In `stable20`, `stable21`, and `stable22` the folder structure is still the old one.

@ChristophWurst Is the `master` branch representing the latest stable or the development version of the docs?

PR #7851, opened 2021-12-17
Update configuration.rst, mention HintException

mejo-:
With https://github.com/nextcloud/server/pull/30130, `setConfigValue` throws a HintException when config is read-only.


ChristophWurst:
I'd add a note about this in the beginning of the page, or show an example of *handling* the exception

mejo-:
> I'd add a note about this in the beginning of the page, or show an example of _handling_ the exception

Agreed and changed :)

mejo-:
@ChristophWurst are you ok with this? :blush:

ChristophWurst (approved):

PR #7877, opened 2021-12-27
add description of new OCS endpoint

sualko:
for recommendations of files and folders

see nextcloud/recommendations#458


ChristophWurst:
Should we document what happens when the app is disabled?

sualko:
For me that's beyond the description of this single endpoint and related more general to the way Nextcloud handles ocs. Not sure if that has been documentated somewhere already. If you like I can put it somewhere or also on the new page. A hint hurts no one :-)

ChristophWurst:
Fair enough. Let's just add a note/warning that the app isn't always enabled

sualko:
I added a notice that the recommendations app needs to be enabled.

ChristophWurst (approved):

PR #7890, opened 2022-01-06
Add Rocky Linux installation example

berlin2123:
Install example for Rocky Linux 9 (Not Released) or Centos Stream 9  (Tested).

copy from example_centos.rst

### DIFF
--- a/admin_manual/installation/example_rockylinux.rst
+++ b/admin_manual/installation/example_rockylinux.rst
@@ -215,7 +215,7 @@ If you need more SELinux configs, refer to the above-mentioned URL, return to th
 
 Once done with with SELinux, please head over to ``http://your.server.com/nextcloud`` and follow the steps as found :doc:`../installation/installation_wizard`, where it will explain to you exactly how to proceed with the final part of the install, which is done as admin user through your web browser.
 
-.. note:: If you use this tutorial, and you see warnings in the web browser after installation about ``OPcache`` not being enabled or configured correctly, you need to make the suggested changes in ``/etc/opt/rh/rh-php74/php.d/10-opcache.ini`` for the errors to disappear. These warnings will be on the Admin page, under Basic settings.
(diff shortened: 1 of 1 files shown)


CarlSchwan:
I'm not sure it's a good idea to have a rocky Linux documentation page and I would prefer to have one page for rhel/rocky/almalinux page and maybe another one for centos stream (if needed). This will be less work to maintain long term

berlin2123:
rhel9/centosstream9/rocky9/almalinux9  can use the same document.

tflidd:
Wouldn't it be better to update the CentOS version and keep this updated. You can then mention that it was tested with other distributions as well.

I think it will be too much to ask to keep manuals for too many Linux distributions up to date. If there is some experience that isn't documented in the official documentation, you can also post in the howto section of the forum help.nextcloud.com

berlin2123:
> Why not using the CentOS stream 9 ?
1,The downstream version of RHEL, CentOS, which has more than 10 years of update support, is over now. The upstream version, CentOS stream, only has nearly 5 years of update support. 
2, The software version of CentOS stream is too new, which may has more bugs.
Choosing other downstream distributions of RHEL,  may make the system run more stable, and get longer update support.

> Why not just update 'example_centos.rst' ?
The old 'example_centos.rst' is suitable for RHEL8/RockyLinux8.
When using RHEL9/RockyLinux9/Almalinux9, there is no CentOS anymore. CentOS Stream 9 is not that stable for NextCloud.

> Why choosing Rocky?
It is the only one Not from any commercial company, which may have longer live time, and avoiding goto Null as CentOS did.

> Merging this now?
The stable RHEL9 is Not Released yet, so the same for the downstream RockyLinux9. We may merge this after the release of RHEL9.

skjnldsv:
As this sounds like a nice feature, the requests for this are obviously quite low. Currently there a no plans to officially support such distros. Thus I will close this ticket for now. 

Sorry we took too long to address that PR, we're trying to gain control again on the opened PRs and clean our backlog so we can best assist the community :)
I thank you again for your initial commitment towards our documentation and collaborative work :bow:

PR #7961, opened 2022-01-14
fix: Update links to calendar, contacts & sync_gnome doc (pim -> groupware)

35am:
Three links were redirecting to doc in the _pim_ directory which doesn't exist anymore and has been replaced with _groupware_.
Issues are encountered with the _firstrunwizard_ displaying links to the calendar and contacts doc :
- https://github.com/nextcloud/firstrunwizard/blob/master/templates/page.clients.php#L59
- https://github.com/nextcloud/firstrunwizard/blob/master/templates/personal-settings.php#L66

Also it seems that most translations are referring to the _pim_ directory even in translations for the _groupware_ one.
See : https://github.com/nextcloud/documentation/search?l=Gettext+Catalog&q=pim
Not being familiar with how the translation works here, an issue will be created to discuss it.


miaulalala (approved):

ChristophWurst:
This effectively reverts https://github.com/nextcloud/documentation/pull/7633 and therefore will break the calendar docs URL from https://github.com/nextcloud/calendar/pull/3800/files

PR #7984, opened 2022-01-26
Document syslog option for the admin audit app

nickvergessen:
Ref https://github.com/nextcloud/server/pull/30763


AndyScherzinger:
According to https://github.com/nextcloud/server/pull/30787 it is also backported to v21

ChristophWurst (approved):

Issue #7797, opened 2021-12-07
Developer documentation for profile page APIs


ChristophWurst:
https://github.com/nextcloud/server/pull/28751 needs documentation in the dev manual, including a backport to 23.

@Pytal could you take care of this please?

ChristophWurst:
@Pytal :ping_pong:

PR #7999, opened 2022-01-28
Document profile actions API

Pytal:
Close https://github.com/nextcloud/documentation/issues/7797

### DIFF
--- a/developer_manual/digging_deeper/profile.rst
+++ b/developer_manual/digging_deeper/profile.rst
@@ -5,7 +5,7 @@ Profile
 The user profile presents the information of a user including their full name,
 profile picture, organisation, role, location, biography, headline, as well as
 actionable information that we call profile actions. These actions can include
-starting a `Nextcloud Talk <https://nextcloud.com/talk/>` chat with the user,
+starting a `Nextcloud Talk <https://nextcloud.com/talk/>`_ chat with the user,
 sending an email, visiting their website, opening their Twitter profile,
 calling their phone number, scheduling an appointment, and more.
 
@@ -14,7 +14,7 @@ App developers can integrate into profile and provide their own actions.
 Register a profile action
 -------------------------
 
-A profile action is represented by a class implementing the `OCP\\Profile\\ILinkAction`
+A profile action is represented by a class implementing the ``OCP\\Profile\\ILinkAction``
 interface. This class is instantiated whenever the user profile is loaded.
 
 .. code-block:: php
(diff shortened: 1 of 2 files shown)


developer_manual/digging_deeper/profile.rst
```diff
@@ -0,0 +1,185 @@
...
+
+        /**
+         * Returns the app ID of the action
+         *
+         * e.g. 'spreed'
+         *
+         * @since 23.0.0
+         */
+        public function getAppId(): string {
+            return 'my_app_id';
```
skjnldsv, line 77: `Application::APP_ID` ?

developer_manual/digging_deeper/profile.rst
```diff
@@ -0,0 +1,185 @@
...
+         * e.g. 'mailto:user@domain.com'
+         *
+         * @since 23.0.0
+         */
+        public function getTarget(): ?string {
+            return $this->urlGenerator->linkToRouteAbsolute('my_app_id.Page.index') . '?pingUser=' . $this->targetUser->getUID();
+        }
+    }
+
+The `MyProfileAction` class needs to be registered during the :ref:`app bootstrap<Bootstrapping>`.
```
Valdnet, line 155: Is it displayed well?
obraz

I have noticed that the other documentation is a similar problem.

developer_manual/digging_deeper/profile.rst
```diff
@@ -0,0 +1,185 @@
+=======
+Profile
+=======
+
+The user profile presents the information of a user including their full name,
+profile picture, organisation, role, location, biography, headline, as well as
+actionable information that we call profile actions. These actions can include
+starting a `Nextcloud Talk <https://nextcloud.com/talk/>` chat with the user,
```
Valdnet, line 8: Is it displayed well?
obraz

developer_manual/digging_deeper/profile.rst
nickvergessen: That would assume people have it defined in the application file and in a documentation I think it's better to not make cross assumptions

nickvergessen: On that it seems to be defined in the example code later on, so should be fine

Valdnet:
In terms of grammar, it is fine in my opinion 👍.

Valdnet (approved):

developer_manual/digging_deeper/profile.rst
Pytal: > That would assume people have it defined in the application file and in a documentation I think it's better to not make cross assumptions

Yeah, pretty much was making it more verbose as it would be trivial for app devs to make that tweak themselves if they so choose

For a first time app dev it would be clearer to keep this and they can do it DRY when they get the hang of it

ChristophWurst (approved):
:pray:

Pytal:
@nickvergessen you resolved right when I replied 😂

PR #8006, opened 2022-02-07
Use consistent data size units

carlcsaposs:
Signed-off-by: Carl Csaposs <carl@csaposs.com>

### DIFF
--- a/admin_manual/configuration_files/big_file_upload_configuration.rst
+++ b/admin_manual/configuration_files/big_file_upload_configuration.rst
@@ -1,6 +1,6 @@
-===========================
+============================
 Uploading big files > 512 MB
-===========================
+============================
 
 The default maximum file size for uploads is 512 MB. You can increase this 
 limit up to what your filesystem and operating system allows. There are certain 
--- a/admin_manual/configuration_files/big_file_upload_configuration.rst
+++ b/admin_manual/configuration_files/big_file_upload_configuration.rst
@@ -1,14 +1,14 @@
 ===========================
-Uploading big files > 512MB
+Uploading big files > 512 MB
 ===========================
 
-The default maximum file size for uploads is 512MB. You can increase this 
+The default maximum file size for uploads is 512 MB. You can increase this 
 limit up to what your filesystem and operating system allows. There are certain 
 hard limits that cannot be exceeded:
 
-* < 2GB on 32Bit OS-architecture
-* < 2GB with IE6 - IE8
-* < 4GB with IE9 - IE11
+* < 2 GB on 32Bit OS-architecture
+* < 2 GB with IE6 - IE8
+* < 4 GB with IE9 - IE11
 
 64-bit filesystems have much higher limits; consult the documentation for your 
 filesystem.
@@ -23,7 +23,7 @@ System configuration
 * Make sure that the latest version of PHP is installed
 * Disable user quotas, which makes them unlimited
 * Your temp file or partition has to be big enough to hold multiple 
-  parallel uploads from multiple users; e.g. if the max upload size is 10GB and 
+  parallel uploads from multiple users; e.g. if the max upload size is 10 GB and 
   the average number of users uploading at the same time is 100: temp space has 
   to hold at least 10x100 GB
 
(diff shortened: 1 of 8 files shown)


carlcsaposs:
Should binary byte units be specified according to [ISO/IEC 80000-13](https://en.wikipedia.org/wiki/ISO/IEC_80000#Part_13:_Information_science_and_technology) (e.g. kibibyte, KiB)?

PHP uses "kilobyte" to represent 1024 bytes: https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes
Nextcloud's front-end appears to do the same

developer_manual/client_apis/WebDAV/search.rst
```diff
@@ -166,7 +166,7 @@ Get a file by id.
         </d:basicsearch>
     </d:searchrequest>
 
-Get all png and jpg files over 10MB.
+Get all png and jpg files over 10 MB.
```
carlcsaposs, line 169: This example uses 10 MB = 10000000 bytes (see line 207)

admin_manual/configuration_files/big_file_upload_configuration.rst
```diff
@@ -1,14 +1,14 @@
 ===========================
-Uploading big files > 512MB
+Uploading big files > 512 MB
 ===========================
```
nickvergessen, line 3:
```suggestion
============================
```

nickvergessen, line 3:
```rst
============================
Uploading big files > 512 MB
============================
```

Need to adjust the count of `=` as well

admin_manual/configuration_files/big_file_upload_configuration.rst
```diff
@@ -151,9 +151,9 @@ For upload performance improvements in environments with high upload bandwidth,
 
  sudo -u www-data php occ config:app:set files max_chunk_size --value 20971520
 
-Put in a value in bytes (in this example, 20MB). Set ``--value 0`` for no chunking at all.
+Put in a value in bytes (in this example, 20 MB). Set ``--value 0`` for no chunking at all.
 
-Default is 10485760 (10 MiB).
+Default is 10485760 (10 MB).
```
nickvergessen, line 156: MiB is actually correct here.
1MB = 1.000 KB
1 MiB = 1.024 KiB

admin_manual/configuration_server/config_sample_php_parameters.rst
```diff
@@ -1909,7 +1909,7 @@ Values that undershoot the minimum will be ignored in favor of the minimum.
 
 	'hashingMemoryCost' => PASSWORD_ARGON2_DEFAULT_MEMORY_COST,
 
-The memory in KiB to be used by the algorithm for computing a hash. The value
+The memory in KB to be used by the algorithm for computing a hash. The value
```
nickvergessen, line 1912: also incorrect

admin_manual/installation/nginx.rst
```diff
@@ -91,11 +91,11 @@ or CSS (.css) files via gzip could also cause such issues.
 Another cause of this issue could be not properly including mimetypes in the
 http block, as shown `here. <https://www.nginx.com/resources/wiki/start/topics/examples/full/>`_
 
-Upload of files greater than 10 MiB fails
+Upload of files greater than 10 MB fails
```
nickvergessen, line 94: also wrong

admin_manual/installation/nginx.rst
```diff
@@ -91,11 +91,11 @@ or CSS (.css) files via gzip could also cause such issues.
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If you configure nginx (globally) to block all requests to (hidden) dot files,
-it may be not possible to upload files greater than 10 MiB using the webpage
+it may be not possible to upload files greater than 10 MB using the webpage
```
nickvergessen, line 98: also wrong

nickvergessen (changes requested):

carlcsaposs:
> Should binary byte units be specified according to [ISO/IEC 80000-13](https://en.wikipedia.org/wiki/ISO/IEC_80000#Part_13:_Information_science_and_technology) (e.g. kibibyte, KiB)?
> 
> PHP uses "kilobyte" to represent 1024 bytes: https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes Nextcloud's front-end appears to do the same

@nickvergessen Here's a branch with ISO 80000-13 units: https://github.com/nextcloud/documentation/compare/master...carlcsaposs:units-iso

Should the documentation follow the ISO 80000-13 standard or the [Nextcloud front-end](https://github.com/nextcloud/server/blob/6da8a6d62adde9b1ed011a7a18d43984cf5a27a8/lib/private/legacy/OC_Helper.php#L64) and PHP standard?

Please let me know if you'd like me to push the changes on the `carlcsaposs:units-iso` branch to this PR.

---

Also, it appears that the comments in Nextcloud's source code use inconsistent units (see [this example](https://github.com/nextcloud/server/blob/6da8a6d62adde9b1ed011a7a18d43984cf5a27a8/lib/private/legacy/OC_Files.php#L57) and [this example](https://github.com/nextcloud/server/blob/6da8a6d62adde9b1ed011a7a18d43984cf5a27a8/apps/files_external/3rdparty/icewind/smb/src/Native/NativeReadStream.php#L16)).

I'd be happy to submit PRs to update the units across Nextcloud's repositories--please let me know if that would be useful.

tflidd:
It's better to call things what they are. Advantage of the Si-units is that you have MiB and MB to make a difference.

Question now would be, if php consiers 1M to be 1 MiB, and now you want to set a quota, do we show that you need to select a value in MiB, or do we ask a value in MB and then do the internal conversion to MiB to apply the limit via php?
And we have the clients as well.

There was a long discussion before the fork about the use of SI units: https://github.com/owncloud/core/issues/13386

carlcsaposs:
@tflidd I appreciate you sharing the context of the owncloud issue--I didn't think to search owncloud repositories.  Thank you!

PR #8028, opened 2022-02-15
Add space as allowed user id character

nickvergessen:
Ref https://github.com/nextcloud/server/blob/master/lib/private/User/Manager.php#L426


ChristophWurst (approved):
Oh no

PR #8046, opened 2022-02-17
Fix nginx example (cache immutable)

CarlSchwan:


icewind1991:
After some more testing I realized that nginx doesn't use the query params when matching location blocks so that can't be used to distinguish the requests in this way.

I've submitted a version using `map` here: https://github.com/nextcloud/documentation/pull/8083 which works in my testing (without `?v=1` you get normal `cache-control`, with `?v=1` you get `cache-control: immutable`)

PR #8095, opened 2022-02-24
Add mapfile serving for nginx conf examples

skjnldsv:
It got lost  🙄

https://github.com/nextcloud/documentation/pull/1504

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -138,7 +138,7 @@ server {
         fastcgi_max_temp_file_size 0;
     }
 
-    location ~ \.(?:css|js|svg|gif|png|jpg|ico|wasm|tflite)$ {
+    location ~ \.(?:css|js|svg|gif|png|jpg|ico|wasm|tflite|map)$ {
         try_files $uri /index.php$request_uri;
         expires 6M;         # Cache-Control policy borrowed from `.htaccess`
         access_log off;     # Optional: Don't log access to assets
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):

szaimen:
@skjnldsv do you mind adding this also to the .htaccess file?
https://github.com/nextcloud/server/blob/master/.htaccess

szaimen:
Otherwise I fear this will remain broken for native apache instances...

skjnldsv:
> @skjnldsv do you mind adding this also to the .htaccess file? [https://github.com/nextcloud/server/blob/master/.htaccess](https://github.com/nextcloud/server/blob/master/.htaccess?rgh-link-date=2022-02-25T13%3A34%3A41Z)

I  thought we had that already :face_exhaling:

skjnldsv:
https://github.com/nextcloud/server/blob/22a9c54f89ac863abc6320f5db21436f41c56a19/lib/private/Setup.php#L531 
Ok, why :facepalm:

skjnldsv:
Yeah, it seems done already then.
If you have `htaccess.RewriteBase` properly configured (by default), then it's fine

skjnldsv:
> @skjnldsv do you mind adding this also to the .htaccess file? [https://github.com/nextcloud/server/blob/master/.htaccess](https://github.com/nextcloud/server/blob/master/.htaccess?rgh-link-date=2022-02-25T13%3A34%3A41Z)

Also, those lines onlyadd caching, which we probably don't want for map files :wink:

szaimen:
Right, so I guess with apache it should already work?

PR #8111, opened 2022-03-03
Fix nextcloud mail Message-ID header

kamukvark:
See also: https://github.com/nextcloud/mail/issues/5764

(Redirected here from https://github.com/nextcloud/docker/pull/1711)


tflidd (changes requested):
Question: You need to fix this in
https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/nginx-subdir.conf.sample
for the setups within a subfolder as well.

PR #8121, opened 2022-03-14
Add documentation for preview hpb with Imaginary

CarlSchwan:


artonge (approved):

solracsf:
Provide a working docker command here would be a good thing:
https://github.com/h2non/imaginary#docker (like, is the `-enable-url-source` param necessary?)

sebalis:
Will there be some guidance on how to extend a more or less standard installation using Docker Compose? I am guessing that a new image will have to be added and some configuration in the Nextcloud container changed and would appreciate some help with the details.

CarlSchwan:
Here is an example with docker compose: https://github.com/juliushaertl/nextcloud-docker-dev/pull/59 I'll try to add this to the doc too

Issue #8122, opened 2022-03-14
Document profile configuration in admin manual


Pytal:
New configuration options introduced in https://github.com/nextcloud/server/pull/29372

PR #8141, opened 2022-03-19
Document profile configuration options for admins

Pytal:
Close https://github.com/nextcloud/documentation/issues/8122


admin_manual/configuration_user/profile_configuration.rst
```diff
@@ -0,0 +1,28 @@
+=====================
```
szaimen, line 1: I suppose this needs to be somehow included into the doctree?
(github actions warning)

admin_manual/configuration_user/profile_configuration.rst
```diff
@@ -0,0 +1,28 @@
...
+
+To enable or disable profile by default for new users switch the toggle in
+Basic settings under the Administration settings section.
+
+.. figure:: ../images/profile_default_setting.png
+
+.. note:: If you are upgrading from Nextcloud 22 to 23 and want profile to
+  be disabled by default for all users, you must upgrade then switch the toggle
+  in Basic settings before letting any users log in.
+
```
szaimen, line 20:
```suggestion

---

```

admin_manual/configuration_user/profile_configuration.rst
```diff
@@ -0,0 +1,28 @@
...
+disable profile globally to remove all profile functionality.
+
+To enable or disable profile by default for new users switch the toggle in
+Basic settings under the Administration settings section.
+
+.. figure:: ../images/profile_default_setting.png
+
+.. note:: If you are upgrading from Nextcloud 22 to 23 and want profile to
+  be disabled by default for all users, you must upgrade then switch the toggle
+  in Basic settings before letting any users log in.
```
szaimen, line 19: Do we want to document how to set it before upgrading as well? Would probably be `occ config:app:set settings profile_enabled_by_default --value="0"` or so.

szaimen:
Maybe we should also add to the upgrade documentation if there is any?

szaimen (approved):

artonge (approved):

Valdnet (approved):

admin_manual/configuration_user/profile_configuration.rst
nickvergessen: Yeah in https://github.com/nextcloud/documentation/blob/a6197f77d3c41e63104dd5a589add965f53dab9c/admin_manual/configuration_user/index.rst

admin_manual/configuration_user/profile_configuration.rst
nickvergessen: We never do that.

nickvergessen (dismissed):

nickvergessen:
Bildschirmfoto von 2022-03-21 13-41-10

Pytal:
> Maybe we should also add to the upgrade documentation if there is any?

Marketing has been notified and this will be added to any relevant communications

Pytal:
> In which Version should this be available or work? I am using latest 23.0.3 and i dont have that switch nor the profile.enabled -> false in config.php has any effect

Will be available on 23.0.4 as indicated by the milestone in https://github.com/nextcloud/server/pull/31624

PR #8142, opened 2022-03-19
Clarify apps need to be replicated during restores and migrations

thomasthe1st:
Added the advice to think about copying the apps folders since otherwise 2FA might be broken.

### DIFF
--- a/admin_manual/maintenance/backup.rst
+++ b/admin_manual/maintenance/backup.rst
@@ -9,6 +9,8 @@ To backup a Nextcloud installation there are four main things you need to retain
 #. The theme folder
 #. The database
 
+In case you have activated 2FA you should consider to copy your apps and/or custom_apps folder as well.
+
 Maintenance mode
 ----------------
 


ChristophWurst:
could you elaborate? What's special about 2FA in terms of backing up apps?

thomasthe1st:
Hi Christoph,

after migrating to a new server I was not able to login anymore.
I had copied over the db, and the config/, themes/ and data/ folders.

The codes did not work and the 2FA app (twofactor_totp) was not enabled. Therefore nextcloud did not ask for the codes but directly asked for backup codes. The backup codes however did not work as well.
I then googled on how to solve it. I stumbled upon this reddit entry. https://www.reddit.com/r/NextCloud/comments/rs6ewd/2fa_not_working_after_server_migration/
I thought it is worth trying to copy over the apps/ and custom_apps/ folders.
It was working and everything was ok. 2FA was successfully working again.

Therefore I thought it might be worth adding this to the documentation,

ChristophWurst:
Thanks.

The problem you're describing could be critical for any type of app. You are also in trouble if you forget to back up your authentication apps, file system apps and so on.

ChristophWurst:
Let's just add *apps* to the list of things to back up?

thomasthe1st:
Adding the apps folder to the list of things to backup sounds good to me.

tflidd:
There are apps that are storing information to the data folder!
E.g. the updater app under `updater-randomstring`, some apps just as their app names like gallery, files_external, gpxedit, news, ... and others withint the `appdata_randomstring` folder such as:
appstore, bookmarks, cms_pico, external, preview, text, theming, ...

And in the developer documentation, it says there is a controller to handle this (not said, where data is stored though ...):
https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/appdata.html
Here it was implemented:
https://github.com/nextcloud/server/pull/1306

It was done on purpose to separate code from data. You should ask the developers of these apps to fix their apps!

@ChristophWurst Shouldn't you know that?

ChristophWurst:
@tflidd I'm afraid I don't understand. Those apps write to *app data*, right? App data is part of the data directory. This will be backed up.

@thomasthe1st issue was that if you restore only data to a vanilla installation of Nextcloud, any additional apps will be missing until you install them again. Critical apps will render the installation unusable, however. So the idea is to back up data *and* app code.

ChristophWurst:
If we go by example, what if your installation relies on user_saml for authentication. After disaster you restore *data* from backup but take the code from a Nextcloud base installation. That means the user_saml app isn't there and you can't run your Nextcloud to install the app.

I think that was the point of backing up data and code (including apps).

joshtrichards:
> Let's just add apps to the list of things to back up?

Whether the `apps_paths` need to be backed up will depend on the operator's specific restore approach. The *Admin Manual* at the moment mostly just outlines very basic approach that those doing archive style installations can use as a model. There are numerous variations.

The other option might to mention in both the *Restoring* and *Migrating* chapters that the replacement environment should have the same apps installed (one way or another). I guess that isn't noted strongly anywhere.

skjnldsv:
It seems we're not sure about this and we were unable to find a direction to take.

Thanks for the interest in Nextcloud and the effort put into this! 🙇

PR #8153, opened 2022-03-23
Split up app upgrade pages by major release

ChristophWurst:
The old page got longer and longer and it wasn't possible to link to
specific sections as we had multiple headings with the same title on a
single page. E.g. https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#changed-apis breaks as soon as the next major is added, then the link will be something like https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#id11.

The upgrade guide idea was taken from Laravel, and they have one page per release as well: https://laravel.com/docs/9.x/upgrade


skjnldsv (approved):
Oh wow! :heart:

PR #8165, opened 2022-03-24
Give some pointers for the PHP8.1 support for apps

ChristophWurst:
Those are quite generic but matches what we did for the groupware apps. So this should be helpful for others as well.

Bildschirmfoto von 2022-03-24 11-54-15

Ref https://github.com/nextcloud/groupware/issues/27

### DIFF
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_24.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_24.rst
@@ -7,12 +7,51 @@ Upgrade to Nextcloud 24
 General
 -------
 
-tbd
+info.xml
+^^^^^^^^
+
+Make sure your ``appinfo/info.xml`` allows for Nextcloud 24.
+
+.. code-block:: xml
+
+	<dependencies>
+	  <nextcloud min-version="21" max-version="24" />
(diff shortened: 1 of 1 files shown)


developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_24.rst
```diff
@@ -7,13 +7,50 @@ Upgrade to Nextcloud 24
...
+.. code-block:: xml
+
+	<dependencies>
+	  <nextcloud min-version="21" max-version="24" />
+	</dependencies>
 
 Front-end changes
 -----------------
 
 tbd
```
come-nc, line 25:
```suggestion
```

come-nc (approved):
Maybe provide a link to PHP 8.1 changes?
https://www.php.net/migration81 or https://stitcher.io/blog/new-in-php-81#breaking-changes

developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_24.rst
ChristophWurst: those are for the front-end section, I'd leave that for now

PR #8171, opened 2022-04-01
Update supported web browsers

skjnldsv:
Internet explorer is not supported, Edge have issues so we do not recommend it.

### DIFF
--- a/user_manual/webinterface.rst
+++ b/user_manual/webinterface.rst
@@ -35,8 +35,7 @@ latest and supported version of a browser from this list:
    or Google **Chrome**/Chromium 49+ to have the full experience with video calls and 
    screensharing. Google **Chrome**/Chromium requires a additional plugin for screensharing.
    
-.. note:: While we try our best to support it, we do **NOT** recommend the usage of microsoft Edge.
-   Microsoft **Internet Explorer** is **NOT** supported.
+.. note:: Microsoft **Internet Explorer** is **NOT** supported.
 
 Navigating the main user interface
 ----------------------------------
--- a/user_manual/webinterface.rst
+++ b/user_manual/webinterface.rst
@@ -14,17 +14,6 @@ Web browser requirements
 For the best experience with the Nextcloud web interface, we recommend that you use the 
 latest and supported version of a browser from this list:
 
-.. No need to translate
-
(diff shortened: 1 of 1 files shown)


user_manual/webinterface.rst
```diff
@@ -14,29 +14,29 @@ Web browser requirements
...
 .. No need to translate
 
-* Apple **Safari**
+* Microsoft **Edge**
 
 .. note:: If you want to use Nextcloud Talk you need to run Mozilla **Firefox** 52+
    or Google **Chrome**/Chromium 49+ to have the full experience with video calls and 
    screensharing. Google **Chrome**/Chromium requires a additional plugin for screensharing.
+   
+.. note:: While we try our best to support it, we do **NOT** recommend the usage of microsoft Edge.
```
szaimen, line 38: are you sure? Microsoft switched to chromium as base for microsoft edge a while ago...

szaimen (approved):

user_manual/webinterface.rst
skjnldsv: Yeah, even with that I think we had reports iirc. 
But also, most of us are actually never testing with that browser, could be a safe call to not recommend it.

skjnldsv: cc @juliushaertl @danxuliu @ChristophWurst @PVince81 @nickvergessen

nickvergessen: Edge is supported and should work flawlessly (apart from a minor issue in Talk)
I still don't recommend using it as it means you run windows, but should be fine in general

If you still keep this as least fix the typo:
```suggestion
.. note:: While we try our best to support it, we do **NOT** recommend the usage of Microsoft Edge.
```

szaimen: > I still don't recommend using it as it means you run windows

It also runs on macos, ios android and partially on Linux, if I remeber correctly: https://docs.microsoft.com/en-us/deployedge/microsoft-edge-supported-operating-systems

nickvergessen: Thanks for ruining the joke :P

szaimen: Sorry 😅

PR #8174, opened 2022-04-04
Document the string length issue

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


ChristophWurst (approved):
Should also go into https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_24.html I guess :)

PR #8186, opened 2022-04-11
Update example Ubuntu instructions for Ubuntu 22.04 LTS

nathanhaines:
Signed-off-by: Nathan Haines <nhaines@ubuntu.com>

### DIFF
--- a/admin_manual/installation/example_ubuntu.rst
+++ b/admin_manual/installation/example_ubuntu.rst
@@ -1,31 +1,33 @@
 .. _ubuntu_installation_label:
 
-Example installation on Ubuntu 20.04 LTS
+Example installation on Ubuntu 22.04 LTS
 ========================================
 
-You can use .deb packages to install the required and recommended modules for a typical Nextcloud
-installation, using Apache and MariaDB, by issuing the following commands in a
-terminal::
+You can use .deb packages to install the required and recommended modules for a
+typical Nextcloud installation, using Apache and MariaDB, by issuing the
+following commands in a terminal::
 
     sudo apt update
-    sudo apt install apache2 mariadb-server libapache2-mod-php7.4
-    sudo apt install php7.4-gd php7.4-mysql php7.4-curl php7.4-mbstring php7.4-intl
-    sudo apt install php7.4-gmp php7.4-bcmath php-imagick php7.4-xml php7.4-zip
+    sudo apt install apache2 mariadb-server libapache2-mod-php8.1 php8.1-gd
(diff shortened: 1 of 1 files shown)


nathanhaines:
Hello!  As Nextcloud 24 is the first version that supports Ubuntu 22.04 LTS, I thought I would bring the example Ubuntu installation instructions forward to that version.  I have tested these instructions twice on Ubuntu 22.04 beta virtual machines and am happy to confirm that with copying and pasting, they result in a working Nextcloud system once an Apache2 conf file is created as per the example Apache configuration page that is linked to at the end of the instructions.

tflidd:
Cool. And Ubuntu is about to be released just before NC 24. So we can update this for the NC24 version.

I just went through your changes quickly, and it looks good but I didn't spin up a server to test it.

tflidd (approved):

tflidd:
@bruno666 had concerns regarding the use of specific php version numbers. Perhaps we can this address at the same time here.
ref: https://github.com/nextcloud/documentation/pull/6802

bruno666:
Thanks @tflidd 
PHP packages should be installed without the version number on Debian and Ubuntu. The meta-packages without version number always depends on the latest available version, this guarantees smooth upgrades to new Ubuntu or Debian release.
If packages are installed with their version number, their will be mark as manually installed and won't be removed when upgrading the system.

Concerning Mariadb, the right command on a default installation to access mysql admin console is only : `sudo mysql` since the root user is authenticated by is UNIX Account and not with password.

nathanhaines:
Thank you so much for accepting the update!

> @bruno666 had concerns regarding the use of specific php version numbers. Perhaps we can this address at the same time here. ref: #6802

I did look into this very briefly (although since all previous documentation uses specific php version numbers, I didn't spend much time on it).

If I recall specifying version numbers sets up all the Apache mods as well, but not specifying them required some (trivial) extra setup.  Extra packages or a2enmod instructions or something like that.  I'll have more spare cycles in a couple of weeks, after the Ubuntu 22.04 LTS release, if we want to investigate further.  It's probably better for the instructions to be more general, but I'd want to test upgrades (for instance, from 20.04 LTS to 22.04 LTS) and their effect on the php packages before switching to generic package names (which pull in the default versions of php for that release).

nathanhaines:
@bruno666 I've updated the instructions and tested them on a virtual machine, with success.  Thank you for the recommendations.  I think the package installation is much improved.

@tflidd The changes were minimal.  Would you kindly check if there's anything else I need to do?

tflidd (approved):

tflidd:
I created a manual backport request https://github.com/nextcloud/documentation/pull/8312

PR #8209, opened 2022-04-25
Add warnings about scaling issues with ajax and webcron

nickvergessen:
Green sections have been added:
Bildschirmfoto von 2022-04-25 12-23-56

### DIFF
--- a/admin_manual/configuration_server/background_jobs_configuration.rst
+++ b/admin_manual/configuration_server/background_jobs_configuration.rst
@@ -23,6 +23,8 @@ Parameters
 ``maintenance_window_start``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. note:: This setting is only taken into account in ``cron`` mode.
+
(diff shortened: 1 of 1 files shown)


szaimen:
@nickvergessen would it make sense to also add a warning to the admin overview if more than e.g. 5 users are present and webcron is used?
Same for ajax and one user?

nickvergessen:
The red warning that jobs are not executed should be enough? e.g. screenshot from https://github.com/nextcloud/server/issues/31725

szaimen:
> The red warning that jobs are not executed should be enough? e.g. screenshot from [nextcloud/server#31725](https://github.com/nextcloud/server/issues/31725)

does this also show up in the admin overview?

nickvergessen:
> does this also show up in the admin overview?

I guess not. Feel free to add it then.

szaimen:
> > does this also show up in the admin overview?
> 
> I guess not. Feel free to add it then.

will do, thanks :)

szaimen (approved):

PR #8229, opened 2022-05-03
system_reqs: Update Android version for Files

AlvaroBrey:


tflidd (approved):

tflidd:
I see that you work a lot on the android repository. However, in such cases, a reference to the change would be great.

AlvaroBrey:
> I see that you work a lot on the android repository. However, in such cases, a reference to the change would be great.

This has been outdated  for a long time. Min version has been 5.0 since at least March 2021 (probably much earlier, but that's the first changelog entry for vesion), and 6.0 since November 2021. See https://github.com/nextcloud/android/blob/master/CHANGELOG.md

CarlSchwan (approved):

PR #8235, opened 2022-05-04
docs: fix duplicated DB command and add missing

JanMalte:


tflidd (approved):
Looks good, however only the missing indices are documented to be used with the occ command:
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/occ_command.html

nickvergessen (approved):

nickvergessen:
> however only the missing indices are documented to be used with the occ command:

I guess that is another thing to update

PR #8274, opened 2022-05-05
Add profiler doc

CarlSchwan:


developer_manual/digging_deeper/profiler.rst
```diff
@@ -0,0 +1,150 @@
+Profiler
+========
+
+Nextcloud offers a built-in profiler that can help you identifying performance issues
+with your Nextcloud application. This feature is enabled starting from Nextcloud 24.
```
come-nc, line 5:
```suggestion
with your Nextcloud application. This feature is available starting from Nextcloud 24.
```

developer_manual/digging_deeper/profiler.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Get the app
+-----------
+
+The profiler application is available on `GitHub <https://github.com/nextcloud/profiler>`_
+you need to clone the stableX branch if you are using Nextcloud X or master if you are
+using the latest development version of Nextcloud.
```
come-nc, line 12: Why not release it to the appstore? Are other applications recommending this for installation?

developer_manual/digging_deeper/profiler.rst
```diff
@@ -0,0 +1,150 @@
...
+   occ app:enable profiler
+   occ profiler:enable
+
+Disable the app
+---------------
+
+.. code-block:: bash
+
+   occ profiler:disable
+   occ config:system:set debug --value false --type bool # to also disable the debug mode
```
come-nc, line 36: Does the app enable it on its own?
There is no set debug true in the enabling command above.

developer_manual/digging_deeper/profiler.rst
```diff
@@ -0,0 +1,150 @@
...
+            $this->data = [
+                'mydata' => 42
+           ];
+       }
+   }
+
+
+You also need to register the `DataCollector` in your app `boot` method:
+
+.. code-block:: bash
```
artonge, line 137:
```suggestion
.. code-block:: php
```

artonge (changes requested):
In addition, you can also link to an existing implementation as a reference.

developer_manual/digging_deeper/profiler.rst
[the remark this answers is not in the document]
CarlSchwan: no because I don't want users to use it on production without carefully consideration

developer_manual/digging_deeper/profiler.rst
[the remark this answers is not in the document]
CarlSchwan: yes it will enable the debug mode but disable won't disabling it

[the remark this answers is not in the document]
artonge: Could be nice to add it to the doc then :)

developer_manual/digging_deeper/profiler.rst
CarlSchwan: no because I don't want users to use it on production without careful consideration

come-nc (approved):

artonge (approved):

ChristophWurst (approved):
:rocket:

PR #8303, opened 2022-05-12
correct views to publish a calendar

Cellophile:
change Month by dayGridMonth, etc.


tflidd:
Was this changed in the code? The original version was a bit more intuitive...

Cellophile:
Yes I'm agree with you, but the original version doesn't works, and this one yes :)

pierreozoux (approved):
lgtm

PR #1297, opened 2019-03-10
Add service discovery to reverse proxy doc

kesselb:

PR #8318, opened 2022-05-16
Include link to service discovery trouble shooting in iOS sync guide

networkException:
iOS 15 (perhaps earlier versions as well) clients are not able to
connect to Nextcloud (tested 24.0.0) using just the server's domain name
anymore.

This patch updates the documentation to a working url found on the
forum:

https://help.nextcloud.com/t/caldav-and-carddav-ios-issues/73801

Signed-off-by: networkException <git@nwex.de>

### DIFF
--- a/user_manual/groupware/sync_ios.rst
+++ b/user_manual/groupware/sync_ios.rst
@@ -40,8 +40,8 @@ You should now find your contacts in the address book of your iPhone.
   (For this reason a certificate is required at your domain, https://letsencrypt.org/ will do).
 
 
-If it's still not working, have a look at the `Troubleshooting Contacts & Calendar`_
-guide.
+If it's still not working, have a look at `Troubleshooting Contacts & Calendar`_ or `_Troubleshooting Service Discovery`_.
 
 .. _Troubleshooting Contacts & Calendar: https://docs.nextcloud.com/server/stable/admin_manual/issues/general_troubleshooting.html#troubleshooting-contacts-calendar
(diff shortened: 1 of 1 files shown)


tflidd:
Did you configure your service discovery properly?
https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#service-discovery
Without it, the automatic setup won't work:
https://help.nextcloud.com/t/caldav-and-apple-impossible-to-use-caldav-with-mac-and-ios/91560/4

tflidd:
In the documentation they added that it should be done on the reverse proxy:
https://github.com/nextcloud/docker/issues/577#issuecomment-471344129
https://github.com/nextcloud/documentation/pull/1297/files

tflidd:
Either make a small note there directly. On the bottom there is also a link for further troubleshooting and the service discovery is not explicitly mentioned. Have you checked this referring link, if no, then it's probably better to put the information on the page itself.
you can link to the service discovery troubleshooting here: https://docs.nextcloud.com/server/stable/admin_manual/issues/general_troubleshooting.html#service-discovery

kesselb (approved):

extobi:
With CalDav works fine, with CardDav not works for my iPhone/iOS (17.4). On Mac it works both. I become the SSL Error for CardDav (but it's the same SSL as CardDav).

Squ1dw4rd99:
> With CalDav works fine, with CardDav not works for my iPhone/iOS (17.4). On Mac it works both. I become the SSL Error for CardDav (but it's the same SSL as CardDav).

I have the exact same issue

PR #8320, opened 2022-05-17
Add documentation for the default_property_scope config.php key

tcitworld:
Docs for to https://github.com/nextcloud/server/pull/31623

Rendered:
image


szaimen:
Thank you and looks good to me!

Just a small idea: maybe we should document here that these values only apply to newly created users and are not overwritten after the user creation anymore? (or do I understand this incorrectly?)

tcitworld:
Done!

PR #8323, opened 2022-05-18
Improve Upload speed with nginx

CarlSchwan:
This is not tested, but some users report better upload speed with it. See https://github.com/nextcloud/server/issues/25297 and https://blog.cloudflare.com/delivering-http-2-upload-speed-improvements/


CarlSchwan:
@ayanamidev Great idea, I added a comment :)

PR #8324, opened 2022-05-19
Extend translation docs a bit

nickvergessen:
* Adding info that branch protection can only be used in org repositories
* Adding info about the `.tx/backport` config file that no one is aware of I guess :D https://github.com/nextcloud/docker-ci/pull/378/files

Screenshot 2022-05-19 at 16-22-35 Translation — Nextcloud latest Developer Manual latest documentation


ChristophWurst (approved):
:exploding_head:

PR #8557, opened 2022-06-17
Update ocs-status-api.rst table rows to display properly

aaronsegura:
Fix tables so that each row is displayed separately

Signed-off-by: Aaron Segura <aaron@aaronsegura.com>


kesselb (approved):
Thank you very much :+1: 

That looks much better :partying_face:

PR #8585, opened 2022-06-25
App name is Contacs (plural)

p-bo:


p-bo:
I agree, lets omit app name then :-)

skjnldsv (approved):

PR #8929, opened 2022-07-01
Correct name for mentioned symbol

p-bo:
This symbol (see screenshot below this text in document) is representing link, thus it should be named as chain, not impeller.


tflidd (commented):
or a link symbol???
https://emojipedia.org/link/

p-bo:
For consistency sake, there is another occurrence of this symbol mentioned (as "chain link") in manual:
https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webgui.html?highlight=chain
So what about to make these both the same? "Chain link", "Link symbol" - another suggestions?

p-bo:
> To have the same throughout the docs would be the best I think

So what about to (in mentioned other location) use:
Any folder that has been shared is marked with the Shared overlay icon. Public link shares are marked with a --parentheses added-->(chain)<--- link --this also added-->one<--. Unshared folders are not marked:

And the same there, so:
... look for a little (chain) link symbol that looks like this:

This way it would be consistent. In the same time it would retain clear difference between local shares and public link shares in the other occurrence.

kesselb:
Hi @p-bo, can you rebase your branch to have a recent ci run?

p-bo:
Hi @kesselb - done (hopefully) and did change for consistency, as discussed above.

PR #8955, opened 2022-07-02
describe how to show the menu bar

niccokunzmann:
I did not find the menu in the menu bar because it was hidden and because I was looking in the wrong place.
Only as I remembered that there is Edit or Extras usually in the menu bar, I remembered how to access this.

Signed-off-by: Nicco Kunzmann <niccokunzmann@rambler.ru>


tflidd (commented):
if you are on win ... select .... from the menu bar. If the menu bar is hidden, ....
Perhaps a bit better than "These are in the menu bar"?

But very good to clarify this, especially less experienced users are lost easily on such things.

niccokunzmann:
I think, I might not engage with this further. You can either merge it, leave it or close it.

kesselb:

I am closing the pull request because there has been no activity for a while now.

PR #9064, opened 2022-07-07
correct swapped order of log in and out

p-bo:
According this section:
https://docs.nextcloud.com/server/latest/user_manual/en/files/encrypting_files.html#using-encryption

"When encryption has been enabled on your Nextcloud server you will see a yellow banner on your Files page warning you to _log out and then log back in_:"


kesselb (approved):

kesselb:
Thanks @p-bo :+1:

PR #9114, opened 2022-07-24
remove invalid trailing semicolon from hsts header

likeazir:
Just a minor nitpick: Per [RFC6797](https://datatracker.ietf.org/doc/html/rfc6797#section-6.1), the HSTS header must not have a trailing semicolon.


kesselb (approved):
Thanks :+1:

Issue #9113, opened 2022-07-22
Document export size estimation API


Pytal:
Document new API introduced in https://github.com/nextcloud/server/pull/32206

PR #9123, opened 2022-07-26
Document export size estimation API

Pytal:
Close https://github.com/nextcloud/documentation/issues/9113


come-nc (approved):

Valdnet:
@Pytal The unit should be in KB.

Pytal:
As KiB unambiguously means 1024 bytes whereas KB could mean 1000 or 1024 bytes it was chosen in this case

@come-nc can you confirm?

Valdnet:
If we are talking about a binary unit, it is well marked :+1:.

PR #9093, opened 2022-07-15
Improve readability and completedness for Redis

martin-rueegg:
- give more complete examples for Redis config
- use separate sections for the different connection types (TCP, TLS, socket)

Signed-off-by: Martin Rüegg <martin.rueegg@metaworx.ch>

PR #9125, opened 2022-07-27
Update caching_configuration.rst

2fst4u:
Added redis.cluster configuration as I couldn't find this documented anywhere. This information is taken straight from https://github.com/CVO-Technologies/nextcloud-server/blob/master/tests/redis-cluster.config.php#L7 

Signed-off-by: 2fst4u <15257926+2fst4u@users.noreply.github.com>

### DIFF
--- a/admin_manual/configuration_server/caching_configuration.rst
+++ b/admin_manual/configuration_server/caching_configuration.rst
@@ -96,6 +96,25 @@ Redis for the distributed server cache::
        'port' => 6379,
   ],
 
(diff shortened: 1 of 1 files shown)


kesselb:
Hey @2fst4u :wave: 

Thanks for your pull request. I think your changes are covered by https://github.com/nextcloud/documentation/pull/9093 already?

2fst4u:
Ah well that's always the way isn't it! Yea it looks like it, I didn't see those commits so it seems to cover the same information. Happy for this to be declined.

Thanks anyway.

PR #9143, opened 2022-08-23
📝 Include instructions on how to configure the server with Env Variables

LeoColman:
This commit adds to the docs instructions on how to configure a Nextcloud instance to use Environment configuration when values aren't present in config.php.

This should close https://github.com/nextcloud/docker/issues/1544 and https://github.com/nextcloud/docker/issues/1465
It should also render https://github.com/nextcloud/docker/pull/1795 unnecessary.

Signed-off-by: Leonardo Colman <dev@leonardo.colman.com.br>

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -12,6 +12,13 @@ is usually not necessary to edit ``config/config.php``.
    Only manually add configuration parameters to ``config/config.php`` if you need to
    use a special value for a parameter. **Do not copy everything from**
    ``config/config.sample.php`` **. Only enter the parameters you wish to modify!**
+   
+   
+Environment Variables configuration
+------------------------
+
+If a config isn't present in ``config/config.php``, Nextcloud will attempt to get the configuration from the Environment by prepending ``NC_`` before it. For example, if you want to configure the ``default_phone_region`` configuration, you would have to start your application with ``NC_default_phone_region=BR``
+   
 
 Multiple config.php file
 ------------------------


Ornias1993:
Please also document how this works with dicts and lists.

LeoColman:
Hi @Ornias1993 !

Absolutely!
Although I don't really know how it works with dicts and lists.
Could you give me an example?

Ornias1993:
> Hi @Ornias1993 !
> 
> Absolutely! Although I don't really know how it works with dicts and lists. Could you give me an example?

No idea.

LeoColman:
I'll do some exploring :D

kesselb (commented):
Hey :wave:,

thanks for your pull request :+1: 

@PVince81 @CarlSchwan I have mixed feelings about adding this section to the documentation. Afaik we don't recommend this and it's not really done (for example lists are not supported or type conversions via env).

LeoColman:
What would be the correct way to configure these variables using the environment?

kesselb:
> What would be the correct way to configure these variables using the environment?

I'm afraid we don't have something else :see_no_evil: 
`NC_` works for _some_ configuration options in `config.php`. 

Some examples:

It does not work for `trusted_domains` or `objectstore` (because it's an array)

It does not work for `'sharing.enable_share_accept'` (because an boolean is expected but the variable from env is always a string [and `'false' `is `true`](https://3v4l.org/Pm8t1).

It may work for `sharing.mail_link_password_expiration_interval` (but the `.` seems not to work everywhere: https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names)

I'm okay with updating the documentation but then usually people start using a feature and some people might be confused when this way does not work for everything. A library like https://github.com/symfony/dotenv may help us to cover most of the cases. Let's see what Vincent and Carl think about your PR.

LeoColman:
I agree with you:
> A library like [symfony/dotenv](https://github.com/symfony/dotenv) may help us to cover most of the cases.

Having proper implementation is better, even better if we delegate it to somewhere else xD

PR #9149, opened 2022-09-03
Improve migration docs

R0Wi:
* Add section on how migration classes are constructed
* Add a hint for "version" argument on "migrations:execute"

Signed-off-by: Robin Windey <ro.windey@gmail.com>

### DIFF
--- a/developer_manual/basics/storage/migrations.rst
+++ b/developer_manual/basics/storage/migrations.rst
@@ -118,6 +118,33 @@ With this the old column gets removed.
           return $schema;
   }
 
+Construction of migration classes
+---------------------------------
+
+All migration classes are constructed via :ref:`dependency-injection`. So if your migration
+steps need additional dependencies, these can be defined in the constructor of your migration 
+class. 
+
+**Example:** If your migration needs to execute SQL statements, inject a `OCP\\IDBConnection`
+instance into your migration class like this:
+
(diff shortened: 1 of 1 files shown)


kesselb (approved):
Thanks :+1:

kesselb:
I'm a bit unsure about the dependency injection part. Yes you can use dependency injection there. Probably not all services can constructed here.

R0Wi:
> I'm a bit unsure about the dependency injection part. Yes you can use dependency injection there. Probably not all services can constructed here.

Well if i noticed it correctly, the migration classes are constructed via `\OC::$server`. So at least injecting a DB connection works. Not completely sure if everything can be injected but i don't see why not.

Will try to check this 👍

kesselb:
> > I'm a bit unsure about the dependency injection part. Yes you can use dependency injection there. Probably not all services can constructed here.
> 
> Well if i noticed it correctly, the migration classes are constructed via `\OC::$server`. So at least injecting a DB connection works. Not completely sure if everything can be injected but i don't see why not.
> 
> Will try to check this +1

True True. I was probably a bit ambiguous. Usually migrations are executed when some kind of app update is happening. It's possible that old and new code is around. This may lead to weird states. My concern was/is that developers move to much logic to their migrations because they can. But not documenting that DI works there is also wrong.

PR #9154, opened 2022-09-09
Update adding CrowdSec

slimshizn:
Signed-off-by: SS <23506499+slimshizn@users.noreply.github.com>

### DIFF
--- a/admin_manual/installation/harden_server.rst
+++ b/admin_manual/installation/harden_server.rst
@@ -299,3 +299,82 @@ running::
   fail2ban-client status nextcloud
 
 .. _fail2ban download page: https://www.fail2ban.org/wiki/index.php/Downloads
+
+CrowdSec
+^^^^^^^^
+
+Following Ibracorp's and CrowdSec's documentation at https://docs.ibracorp.io/crowdsec/crowdsec/unraid/traefik-bouncer/nextcloud-collection and https://hub.crowdsec.net/author/crowdsecurity/collections/nextcloud.
+
+The following captured from https://docs.crowdsec.net/docs/getting_started/install_crowdsec
+
+CrowdSec can be used with, or replace Fail2ban with the following.
+Install CrowdSec (Linux)
+
+For those that prefer hands-on approach, you can as well manually install crowdsec.
+Install our repositories#
+
+Installing our repositories allows you to access the latest packages of CrowdSec and bouncers.
+info
+
+We are using packagecloud.io service. While curl | sudo bash can be convenient for some, alternative installation methods are available.
+
+    Debian/Ubuntu
+    EL/Centos7
+    EL/Centos Stream 8
+    Amzn Linux 2
+    OpenWRT
+    CloudLinux
+
+curl -s https://packagecloud.io/install/repositories/crowdsec/crowdsec/script.deb.sh | sudo bash
+
+Install CrowdSec#
+
+    Debian/Ubuntu
+    EL/Centos7
+    EL/Centos Stream 8
+    Amzn Linux 2
+    OpenWRT
+    CloudLinux
+
+apt install crowdsec
+
+You now have CrowdSec running ! You can move forward and install a bouncer, or take a tour of the software beforehand !
+
+Directories:
+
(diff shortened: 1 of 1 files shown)


kesselb:
Hey :wave: 

Thanks for your pull request. I'm not a fan of duplicating the installation instructions of another software here. At worst it's outdated in a few weeks. 

I didn't know about CrowdSec until your pull request. However from your pull request I could not figure out what this is about. I would like for our documentation a short introduction about CrowdSec (what is it good for) and a link to their documentation how to configure it with Nextcloud.

slimshizn:
Understood! I can do that for sure. Thanks for answering the PR so it doesn't sit forever. I'll probably get to that this week sometime.

slimshizn (commented):
How's this?

kesselb:
> How's this?

Thank you! Looks good to me. 

I'm still undecided about the step by step guide. How to install the component belongs to their documentation. 

Let's see what Andy thinks about adding CrowdSec.

AndyScherzinger:
I'll check together with @nickvergessen but am currently afk

tflidd:
I think the tipps should be for secure configuration for services that are needed to run Nextcloud (e.g. webserver, database) and that might not be well adjusted in default settings. And it should be rather general, not just for some container solutions that require a special setup....

Beyond that, there are many solutions and it's a lot of work to keep such tutorials updated for different distributions. I'd perhaps mention different possible solutions (intrusion detection, ip blocking, ....) and then link to existing open source solutions and their tutorials.

slimshizn:
Well this is purely for hardening your server, hence "harden_server.rst". I purely respect any input, but I think this is something that is 'aftermarket' and was put in a 'aftermarket' area. Just my thoughts.

admin_manual/installation/harden_server.rst
```diff
@@ -299,3 +299,151 @@ running::
...
+Please keep in mind that raspberry pi OS is designed to work on all raspberry pi versions. Even if the port target is known as armhf, it's not exactly the same target as the debian named armhf port.
+
+The best way to have a CrowdSec version for such an architecture is to do:
+
+    install golang (all versions from 1.16 will do)
+    export GOARCH=arm
+    export CGO=1
+    Update the GOARCH variable in the Makefile to arm
+    install the arm gcc cross compiler (On debian the package is gcc-arm-linux-gnueabihf)
+    Compile CrowdSec using the usual make command
```
nickvergessen, line 385: This is a pure copy of https://docs.crowdsec.net/docs/getting_started/install_crowdsec (which might even violate author rights).

So I would suggest to simply replace it with something like:

> Follow the install instructions presented at https://docs.crowdsec.net/docs/getting_started/install_crowdsec

And the part in our documentation should basically only focus on the configuration, like installing the nextcloud collection from https://hub.crowdsec.net/author/crowdsecurity/collections/nextcloud etc.

admin_manual/installation/harden_server.rst
```diff
@@ -299,3 +299,151 @@ running::
...
+
+``hhttps://docs.ibracorp.io/crowdsec/crowdsec/docker-compose``
```
nickvergessen, line 387: (Type here in the URL)

admin_manual/installation/harden_server.rst
```diff
@@ -299,3 +299,151 @@ running::
...
+Docker CrowdSec Install
```
nickvergessen, line 388: Similar story as above also here. We should not copy the CrowdSec install docs but refer to them instead and focus on the configuration of the nextcloud collection

nickvergessen (changes requested):

admin_manual/installation/harden_server.rst
nickvergessen: But also there don't copy the manual, if it's useful refer to the site, if it's a single line to execute good enough.

skjnldsv:
As there is no feedback since a while I will close this ticket. If this is still something you wish, please make sure to rebase to the latest master commit. After that, feel free to reopen.

PR #9157, opened 2022-09-14
adding token lifetime for loginflow v2

knoxell:
The token lifetime never made it into the developer manual even though relevant.

- As described in the push request for login flow V2 [nextcloud/server#14161](https://github.com/nextcloud/server/pull/14161), the token lifetime is 20 mins.
- In the source code: https://github.com/nextcloud/server/blob/master/core/Db/LoginFlowV2Mapper.php#L34

Signed-off-by: Knoxell <85034894+knoxell@users.noreply.github.com>


kesselb (approved):
Thank you :+1:

PR #9185, opened 2022-09-28
update app list in accordance wiht bundling state of NC 25

blizzz:
contributes to https://github.com/nextcloud/server/issues/34203


szaimen (commented):
Accessibility needs to get removed as well...

blizzz:
> Accessibility needs to get removed as well...

That linked never worked anyway

szaimen (approved):

PR #9188, opened 2022-09-29
Disable xss header and add CSP header

minecrawler:
By now, no modern browser supports the `XSS-Protection` header anymore - and with good reason. It actually has known security vulnerabilities. Instead, it is recommended to disable it and use CSP instead. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

Hence, this patch disables the XSS header and enables CSP instead. The CSP part was copied from an earlier PR (#1714), which was closed because of an incompatible browser matrix. That issue does not exist anymore. Is it possible to merge these changes?

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -69,13 +69,13 @@ server {
     client_body_buffer_size 512k;
 
     # HTTP response headers borrowed from Nextcloud `.htaccess`
-    add_header Referrer-Policy                      "no-referrer"   always;
-    add_header X-Content-Type-Options               "nosniff"       always;
-    add_header X-Download-Options                   "noopen"        always;
-    add_header X-Frame-Options                      "SAMEORIGIN"    always;
-    add_header X-Permitted-Cross-Domain-Policies    "none"          always;
-    add_header X-Robots-Tag                         "none"          always;
-    add_header X-XSS-Protection                     "0"             always;
+    add_header Referrer-Policy                      "no-referrer"             always;
+    add_header X-Content-Type-Options               "nosniff"                 always;
+    add_header X-Download-Options                   "noopen"                  always;
+    add_header Content-Security-Policy              "default-src 'self'"      always;
+    add_header X-Permitted-Cross-Domain-Policies    "none"                    always;
+    add_header X-Robots-Tag                         "none"                    always;
+    add_header X-XSS-Protection                     "0"                       always;
 
     # Remove X-Powered-By, which is an information leak
     fastcgi_hide_header X-Powered-By;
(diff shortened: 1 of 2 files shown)


HLFH:
It would be great to merge.

minecrawler:
I can reverse the second commit and rebase the changes, if that's what it takes.

Also, if this commit still needs other changes (like the check for the presence of the header), please tell me what and where and I can add it :)

iasdeoupxe:
Server side change (remove it completely not just setting it to `0`):
- nextcloud/server#37154
- nextcloud/server#53476

MichaIng (changes requested):
Blocking until https://github.com/nextcloud/server/pull/53476 has been merged, but this is the complementary needed docs part then.

I hope no one has a problem with me adjusting and squashing the commits:
- While disabling XSS explicitly is safe as well, I think it makes more sense to remove and forget this header completely. OWASP recommends to do either of both. For more details see the server PR.
- CSP should not be set at webserver level: It would overwrite or duplicate what Nextcloud sets in PHP already, hence can break things or weaken security. CSP, the way it is designed with very selective permissions all as single header value, needs to be set in the backend as individually. There is no generic safe but functional value that could be applied on webserver level for the whole site.

admin_manual/installation/nginx-root.conf.sample
```diff
@@ -79,7 +79,6 @@ server {
     add_header X-Frame-Options                   "SAMEORIGIN"        always;
     add_header X-Permitted-Cross-Domain-Policies "none"              always;
     add_header X-Robots-Tag                      "noindex, nofollow" always;
-    add_header X-XSS-Protection                  "1; mode=block"     always;
```
nickvergessen, line 82: Should add that to the "critical changes" update manual for admins

MichaIng: IMO it is not problem if admins just leave the header inside. But I will add the note.

admin_manual/release_notes/upgrade_to_32.rst
```diff
@@ -0,0 +1,16 @@
+=======================
+Upgrade to Nextcloud 32
+=======================
+
+System requirements
+-------------------
+
+* PHP 8.1 is now deprecated but still supported.
+* PHP 8.4 is now supported, but 8.3 is recommended.
```
MichaIng, line 9: Is this still accurate?

https://github.com/nextcloud/server/blob/master/lib/versioncheck.php

admin_manual/release_notes/upgrade_to_32.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Web server configuration
+------------------------
+
+* Setup checks do not check for the ``X-XSS-Protection`` response header anymore. It has been removed from Nextcloud's ``.htaccess`` and you may want to adjust your webserver config to not serve it anymore.
+  XSS filtering was supported only until Chromium 78 and similarly old browsers, but had been found to cause more issues, including attack vectors, than it solves.
```
nickvergessen:
```suggestion
  XSS filtering was supported only until Chromium 78 and similarly old browsers, but had been found to cause more issues, including attack vectors, than it solved.
```

MichaIng:
#13445

MichaIng:
#13445

> In addition, the version in the line 1 is outdated.

Or shall we call it the "RPI Ozzy 🤘" version of the Nginx config? 😢🙏

PR #9192, opened 2022-10-01
ocs-api-overview: Talk about user using singular they

jaller94:
The word "it's" should be "its", but I think "their" is nicer because we're talking about a group of humans.

Signed-off-by: Christian Paul <info@jaller.de>


kesselb (approved):

kesselb:
Thank you :+1:

PR #9195, opened 2022-10-07
Update recommend MariaDB versions

J0WI:
- 10.2 is EOL and therefore no longer recommend
- 10.6 is current LTS and supported since https://github.com/nextcloud/server/pull/30129


szaimen:
I would like to get automated testing done before stating 10.6 compatibility officially. See https://github.com/nextcloud/server/pull/33631

szaimen (approved):

PR #9230, opened 2022-10-19
Add JIT to server tuning

pulsejet:


pulsejet:
I know that, which is why I mentioned it won't be bottleneck for the server. "Am I missing something" refers more to the fact that nobody else tried this yet, almost two years since the JIT was released.

Leptopoda:
Just a small nitpick but the last sentence says "To enable ... Enabled:"

Maybe you can rephrase that

PR #9250, opened 2022-10-24
Fix 26 upgrade docs missing from TOC

ChristophWurst:
Leftover from https://github.com/nextcloud/documentation/pull/9246


DaphneMuller (commented):
looks good to me

PR #9259, opened 2022-10-27
Document remote host validation for app devs

ChristophWurst:
For https://github.com/nextcloud/server/pull/34852

## Rendered

image


miaulalala (approved):
If I can understand it, even though I've never heard of it before, so can other devs.

PR #9264, opened 2022-10-28
Shifted NC22 part

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>


nickvergessen:
Yeah, bit the latest and stable etc are symlinked words in the URL, so that's not it, but we also need the 25/26 split.
It's still on the release todo and we just didn't reach it yet.

nickvergessen:
Yeah, sorry

PR #9280, opened 2022-11-02
Document the IConditionalWidget dashboard interface

nickvergessen:
* Server ref https://github.com/nextcloud/server/pull/34832

### DIFF
--- a/developer_manual/digging_deeper/dashboard.rst
+++ b/developer_manual/digging_deeper/dashboard.rst
@@ -147,6 +147,21 @@ however this method is deprecated and will be removed once Nextcloud 19 is EOL.
         }
     }
 
+++++++++++++++++++++++++++++++++
+The IConditionalWidget interface
+++++++++++++++++++++++++++++++++
+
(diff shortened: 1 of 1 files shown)


developer_manual/digging_deeper/dashboard.rst
```diff
@@ -147,6 +147,21 @@ however this method is deprecated and will be removed once Nextcloud 19 is EOL.
         }
     }
 
+++++++++++++++++++++++++++++++++
```
ChristophWurst, line 150: I don't think there should be a line above the heading https://sublime-and-sphinx-guide.readthedocs.io/en/latest/topics.html?highlight=heading#use-headings

nickvergessen: Copied from below in the same document
https://docs.nextcloud.com//server/latest/developer_manual/digging_deeper/dashboard.html#the-ioptionwidget-interface

But yeah let me remove

ChristophWurst (approved):

PR #9281, opened 2022-11-02
Update DI parameters, aliases and services documentation

ChristophWurst:
The info was slightly outdated.

* Updated missing aliases
* Sort parameters alphabetically
* Fix namespaces of services
* Fix heading symbol length

### DIFF
--- a/developer_manual/basics/dependency_injection.rst
+++ b/developer_manual/basics/dependency_injection.rst
@@ -310,45 +310,52 @@ The following parameter names and type hints can be used to inject core services
 Parameters:
 
 * **AppName**: The app id
-* **WebRoot**: The path to the Nextcloud installation
 * **UserId**: The id of the current user
+* **WebRoot**: The path to the Nextcloud installation
+
+Aliases:
+* **appName**: resolves to ``AppName``
+* **Request**: resolves to ``\OCP\\IRequest``
+* **ServerContainer**: resolves to ``\OCP\IServerContainer``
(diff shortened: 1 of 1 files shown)


developer_manual/basics/dependency_injection.rst
```diff
@@ -310,45 +310,52 @@ The following parameter names and type hints can be used to inject core services
...
+* ``\OCP\Contacts\IManager``
+* ``\OCP\IDateTimeZone``
+* ``\OCP\IDBConnection``
+* ``\OCP\Diagnostics\IEventLogger``
+* ``\OCP\Diagnostics\IQueryLogger``
+* ``\OCP\Files\Config\IMountProviderCollection``
+* ``\OCP\Files\IRootFolder``
+* ``\OCP\IGroupManager``
+* ``\OCP\IL10N``
+* ``\OCP\ILogger``
```
juliushaertl, line 342: Should we rather list the LoggerInterface here?

ChristophWurst: Yes, makes sense. And also mark the ILogger as deprecated.

I did not check this list. There is a lot more that can be injected these days and I don't know how to get a reliable list for that info. This has to do with auto-wiring. In the past it was easier to see what is registered on ever app container.

juliushaertl: Right, getting a full list is probably hard

PR #9285, opened 2022-11-03
Document deprecation of pascal case DI container parameters

ChristophWurst:
For https://github.com/nextcloud/server/pull/34929


christianlupus:
I have seen a few more corrections to keep things consistent. Let me commit and push, onemoment.

christianlupus (approved):
This looks good to me in general. Please check my additional commit as there were a few flaws that I fixed (at least I think so).

artonge (approved):

PR #9351, opened 2022-11-14
Use versionadded directive for added OCP events

ChristophWurst:
Why write it out when sphinx can write the text for us?

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-11-14 14-38-20 | Bildschirmfoto vom 2022-11-14 14-38-07 |


ChristophWurst:
I'm not backporting further but only to 25 so we have this in the latest *stable* docs.

kesselb:
Awesome :+1:

juliushaertl (approved):

MorrisJobke:
🥳 Nice one1

PR #9389, opened 2022-11-17
PHP module for Apache: hardcoded version and note

p-bo:
To prevent outdating, used variable instead of hardcoded version of Apache module for PHP. Also added note (and reference) regarding necessity to have only one version enabled at time.

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -179,7 +179,8 @@ these modules:
 * mod_security
 * mod_reqtimeout
 * mod_deflate
-* libapache2-mod-php*filter (use libapache2-mod-php7.4 instead)
+* libapache2-mod-php*filter (use libapache2-mod-php{VERSION} instead. Note: `have only module for one version enabled
+  <https://help.nextcloud.com/t/broken-nextcloud-after-ubuntu-lts-upgrade-to-php8-1/138355/8>`_)
 * mod_spdy together with libapache2-mod-php5 / mod_php (use fcgi or php-fpm
   instead)
 * mod_dav


tflidd (approved):

admin_manual/issues/general_troubleshooting.rst
```diff
@@ -179,7 +179,8 @@ these modules:
 * mod_security
 * mod_reqtimeout
 * mod_deflate
-* libapache2-mod-php*filter (use libapache2-mod-php7.4 instead)
+* libapache2-mod-php*filter (use libapache2-mod-php{VERSION} instead. Note: `have only module for one version enabled
```
kesselb, line 182: > libapache2-mod-php*filter

I'm not sure what this is supposed to mean. libapache2-mod-php7.4-filter or maybe a reference to mod_filter?

> libapache2-mod-php{VERSION}

libapache2-mod-php is the debian/ubuntu name. Can we use mod_php here?

tflidd: I understand to fill in the php version you are using (8.0, 8.1 etc.). It's most generic especially if you think about ubuntu 20.04 users, their default php is 7.4 and soon they need to use a version through external packages, if you then install libapache2-mod-php, is it going to install the right version from 3rd-party or going to install the original php7.4 version stuff removing newer installations again?

kesselb: Thanks for your reply. 

> I understand to fill in the php version you are using (8.0, 8.1 etc.)

I am aware and wonder about "libapache2-mod-php*filter".
If you replace * with 7.4 it's "libapache2-mod-php7.4-filter". What module is that? 

Moreover we use the "apache name" (e.g. mod_security, mod_deflate) for all modules expect the php module. 
I would like to use the "apache name" for the php module as well.

ChristophWurst: Wo which way do we want to go?

joshtrichards: > If you replace * with 7.4 it's "libapache2-mod-php7.4-filter". What module is that?

It's for the ancient apache2filter SAPI that was dropped entirely in PHP 7.0 (in favor of apache2handler).

We can drop this reference entirely.

PR #9392, opened 2022-11-18
lowercase instead of uppercase, colon instead of comma

p-bo:
There was combination of comma, followed by uppercase article which is not correct. So replaced upper case article with lowercase one and tried also other variants instead of comma between sentences. It seems, that colon fits best (but may be only subjective perception).

### DIFF
--- a/admin_manual/maintenance/package_upgrade.rst
+++ b/admin_manual/maintenance/package_upgrade.rst
@@ -6,7 +6,7 @@ Upgrade quickstart
 ------------------
 
 One effective, if unofficial method for keeping Nextcloud current on Linux servers is by configuring 
-your system to use Nextcloud via a self contained "Snap" package, A technology allowing users to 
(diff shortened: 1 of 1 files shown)


tflidd (approved):

admin_manual/maintenance/package_upgrade.rst
```diff
@@ -6,7 +6,7 @@ Upgrade quickstart
 ------------------
 
 One effective, if unofficial method for keeping Nextcloud current on Linux servers is by configuring 
-your system to use Nextcloud via a self contained "Snap" package, A technology allowing users to 
+your system to use Nextcloud via a self contained "Snap" package: a technology allowing users to
```
kesselb, line 9:
```suggestion
your system to use Nextcloud via a self-contained "Snap" package: a technology allowing users to 
```

PR #9402, opened 2022-11-19
Fix formating

p-bo:
- use *italicized text* instead of _italicized text_ , because it was not rendered in used markdown flavour correctly
- remove (forgotten?) long line of dashes


kesselb (approved):
Thank you :+1:

PR #9405, opened 2022-11-19
Remove link to disappeared document

p-bo:
This URL is no longer functional and cannot by found in internet archive :-(


tflidd:
This one?
https://oc.owncloud.com/rs/038-KRL-592/images/Whitepaper_Overview_of_ownCloud_Encryption_Model_ENG.pdf
https://doczz.net/doc/7387460/owncloud-s-data-encryption-model

p-bo:
Unfortunately not, link was pointing to version 2.2 of this document, which is not available any more:

obrazek

That one from doczz.net is from year 2014 (found in pdf metadata) - thus oldest.
obrazek


The one you pointed from owncloud site is from year 2016 (at least from metadata), but is stated as version 2.0.
obrazek


So at least it seems, that version 2.2 disappeared and older version (2.0) is available. Now what - edit link to point to (probably older version) or remove completely?

helix2301 (approved):
Link is not working and needs to be removed

tflidd (approved):

PR #9406, opened 2022-11-21
Add DevContainer for GitHub Codespaces

R0Wi:
This PR adds an appropriate DevContainer configuration to the project. We can now directly open the project inside GitHub Codespaces or a local VSCode DevContainer.

All required dependencies are bundled into the container, so commands like `make html`,  `make pdf` or `make` will work out of the box. Also development via file watching with `make SPHINXBUILD=sphinx-autobuild html` is possible.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM python:3.9 as final
+
+USER root
+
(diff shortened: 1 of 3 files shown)


kesselb (approved):
Thank you :+1:

PR #9507, opened 2022-12-31
Clarify that the "region" field is required for pre-existing AmazonS3 buckets

Christopher-Hayes:
For pre-existing S3 buckets, region MUST be entered since the files_external app fails to find buckets outside of the eu-west-1 region.

Signed-off-by: Chris Hayes <6013871+Christopher-Hayes@users.noreply.github.com>

In response to this GitHub issue for the files_external app: https://github.com/nextcloud/server/issues/35926

### DIFF
--- a/admin_manual/configuration_files/external_storage/amazons3.rst
+++ b/admin_manual/configuration_files/external_storage/amazons3.rst
@@ -7,6 +7,7 @@ To connect your Amazon S3 buckets to Nextcloud, you will need:
 - S3 access key
 - S3 secret key
 - Bucket name
+- Region
 
 In the **Folder name** field enter a local folder name for your S3 mountpoint.
 If this does not exist it will be created.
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

juliushaertl:
Thanks for your pull request. I'm sorry but I've just merged a different similar contribution earlier https://github.com/nextcloud/documentation/commit/b8e05779314cf1d62686499102a30d47575b01a9 so i think your changes are covered by that already.

admin_manual/configuration_files/external_storage/amazons3.rst
```diff
@@ -20,9 +21,12 @@ highly-recommended.
 .. figure:: images/amazons3.png
    :alt:
 
-Optionally, you can override the hostname, port and region of your S3 server,
+Optionally, you can override the hostname, and port of your S3 server,
 which is required for non-Amazon servers such as Ceph Object Gateway.
 
+**Region** should be an [AWS Region Code](https://docs.aws.amazon.com/general/latest/gr/rande.html#region-names-codes).
+If no region is supplied, a new bucket will be created in the `eu-west-1` region.
```
Christopher-Hayes, line 28: @juliushaertl this is an ongoing issue that, should probably get added. With how it's coded, `eu-west-1` is **always used and for all s3-providers**, unless the user specifies the region.

Ideally, "auto" is used and not `eu-west-1`. But in the meantime, the docs need to make it clear that "region" is a required field. Even if you use another S3-compatible provider (like I'm doing with Cloudflare R2"), Nextcloud will still try to use `eu-west-1`.

Christopher-Hayes: Alright, I created #12098 to help clarify this.

PR #9510, opened 2023-01-02
Improve UTF8 manual with older DB versions

nickvergessen:
It's a bit too much to ask people to check "older versions", when they have to go back 5+ versions although 10.2 is still supported. So I added a link to the newest docs that still contain it and turned it into a warning after the painful journey with @TwoDogs2 in https://github.com/nextcloud/spreed/issues/8496

Before | After
---|---
Bildschirmfoto vom 2023-01-02 07-22-30 | Bildschirmfoto vom 2023-01-02 07-22-46


TwoDogs2:
Thank you very much

kesselb (approved):

PR #9528, opened 2023-01-05
Add info to fix encryption key issues

PVince81:
Added information about misplaced encryption keys and how to repair them automatically or manually when needed.

Note: the "encryption:fix-key-location" command in question doesn't repair group folders yet but will eventually, I'm still referencing it as it's worth trying it.


come-nc (approved):

icewind1991 (approved):

PVince81:
merging this as it's already available on master

for the backports we might want to wait for the command's backports to be merged as well

PR #9532, opened 2023-01-07
Update Service Discovery example for Caddy …

rusty-snake:
… in  reverse_proxy_configuration.rst

While the rewrite rules should work I guess, they still trigger a warning on the admin page. Using 301 redirects make the warning go away.


kesselb (approved):
Good catch, thank you :+1: 

https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/reverse_proxy_configuration.rst#apache2

PR #9554, opened 2023-01-13
docs: mention IResponse's getBody(), slash when catching Exception

raphj:
Signed-off-by: raphj <raphael.jakse@xwiki.com>


kesselb (approved):
Thank you :+1:

PR #9558, opened 2023-01-15
Fixed URL

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>

### DIFF
--- a/user_manual/groupware/sync_kde.rst
+++ b/user_manual/groupware/sync_kde.rst
@@ -28,7 +28,7 @@ In Kalendar:
 
 In KOrganizer and Kalendar:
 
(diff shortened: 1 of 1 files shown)


silopolis:
👌

nickvergessen:
I think we should add a redirect instead or additionally.

nickvergessen:
Don't think that's possible, the translations can replace the URL.
But links from older untranslated versions should be mapped to existing pages

tflidd (approved):

tflidd:
I just approved the change to get it corrected. To have links within the language is a more general topic I suppose....

PR #9566, opened 2023-01-18
Add section for troubleshooting quota issues

PVince81:
The workaround about `unencrypted_size` comes from https://github.com/nextcloud/server/issues/25283 and was proven to work for many people already, see https://github.com/nextcloud/server/issues/25283#issuecomment-1386927818 and further below

I had a debugging session with @JuliaKirschenheuter and we did observe that the column `unencrypted_size` is unused when encryption is disabled, so it's safe to set it to zero for non-encrypted file.

I've also verified this on my personal instance where encryption is disabled (`select * from oc_filecache where unencrypted_size > 0;` returns zero results)

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -383,7 +383,7 @@ does not match the actual data stored in the user's ``data/$userId/files`` direc
 .. note::
 
    Metadata, versions, trashbin and encryption keys are not counted in the used space above.
-   Please refer to the quota documentation for details.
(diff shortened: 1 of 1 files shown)


miaulalala (approved):

PVince81:
- [ ] add link to quota.rst in the user docs using the same link format like https://github.com/nextcloud/documentation/blob/41913d1558f9116e61eeeb99d4d9ae0eb53ae2ab/admin_manual/configuration_files/file_sharing_configuration.rst#L124

PVince81:
I believe the unencrypted_size behavior was tweaked in NC 24 only, so not backporting further.

cc @icewind1991 in case I'm wrong

PR #9553, opened 2023-01-13
feat(dev-manual): Warn about the upcoming PSR-0 class loader removal

ChristophWurst:
For https://github.com/nextcloud/server/pull/36114

Bildschirmfoto vom 2023-01-13 13-22-54

PR #9573, opened 2023-01-20
feat(dev-manual): Document performance optimizations with class loaders

ChristophWurst:
This feature was added with https://github.com/nextcloud/server/pull/6853 a while ago for shipped apps in the server repo but has not gained much popularity. I assume it's because nobody knew or profiled the difference.

Thanks to Blackfire profiles I saw warnings that Composer still had to *find* hundreds of files, even if apps had an automated autoloader. My theory: Nextcloud registered its own loader before the app loader is there. So the app loader doesn't even get a chance. Effectively, app loader would only be used for app dependencies.

## Benchmarks

* https://github.com/nextcloud/calendar/pull/4912 showed that we can get from 795 to 101 Composer finds if I optimize the four Groupware apps.
* https://github.com/nextcloud/spreed/pull/8578 allowed me to fully get rid of the Blackfire warning.

As a positive side effect, replacing the server class loader not only disables the PSR-4 loader but also the PSR-0. We know that this gives a significant boost for Nextcloud processes: https://github.com/nextcloud/server/pull/36114 https://github.com/nextcloud/documentation/pull/9553.

## Preview

Bildschirmfoto vom 2023-01-20 16-57-46
Bildschirmfoto vom 2023-01-20 16-57-36

### DIFF
--- a/developer_manual/digging_deeper/classloader.rst
+++ b/developer_manual/digging_deeper/classloader.rst
@@ -10,7 +10,7 @@ The classloader is provided by Nextcloud and loads all your classes automaticall
 PSR-4 autoloading
 -----------------
 
-Nextcloud uses a PSR-4 autoloader. The namespace **\\OCA\\MyApp**
+Nextcloud uses a  :ref:`PSR-0 autoloader<psr4>`. The namespace **\\OCA\\MyApp**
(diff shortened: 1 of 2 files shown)


karlitschek:
wow. very interesting. This filesystem finds can generate a lot of IOPs which can cause bigger slowdowns especially when Nextcloud is stored in a remove network filesystem.

juliushaertl (approved):

PR #9575, opened 2023-01-22
Add removal of `OC.addTranslations` to NC26 upgrade guide

susnux:
Belongs to server PR: https://github.com/nextcloud/server/pull/36287


ChristophWurst (approved):
Thanks a lot!

Strictly speaking "APIs" in `OC` were never really public API but it's good to document the change nevertheless :+1:

PR #130, opened 2016-08-19
Developer Doc wrt Admin page split

blizzz:
forward port  of #128 

btw, merging current stable branch to master does not seem to be working easily anymore? At least i got too many conflicts that it liked to deal with.

PR #9583, opened 2023-01-24
Document `occ status` and `-e` option (Related to #35873)

leegarrett:


leegarrett:
The top of admin_manual/configuration_server/occ_command.rst has a tutorial-like description where `status` is used. It might make sense to remove that for the sake of brevity.

admin_manual/configuration_server/background_jobs_configuration.rst
```diff
@@ -127,11 +127,14 @@ This approach requires two files: **nextcloudcron.service** and **nextcloudcron.
 
   [Service]
   User=www-data
+  ExecCondition=php -f /var/www/nextcloud/occ status -e
```
kesselb, line 130: Good idea :+1: 

https://github.com/nextcloud/server/blob/ad00a149ea2bb262e62fbecf11051fd035f90ec4/cron.php#L45-L52

But isn't the check in cron.php not enough?

leegarrett: Not completely. For cron, this will still trigger an email sent to the admin address. But there are also many other reasons you'd want to skip it like in line #130.  Upgrading my instance to use apcu triggered this mail:
```
OCP\HintException: [0]: Memcache OC\Memcache\APCu not available for local cache (Is the matching PHP module installed and enabled?)
```
A temporarily unavailable DB will fail with a stack trace a lot earlier. Manually up/downgrading the instance might temporarily make some files unavailable and trigger a stacktrace. Usually the issue is clear, so the extra noise in the cron/logging should be avoided IMHO.

I've also noticed that I didn't add the check to cron, should I add it?

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1100,6 +1100,67 @@ Remove a certificate::
...
+     "version": "25.0.2.3",
+     "versionstring": "25.0.2",
+     "edition": "",
+     "maintenance": false,
+     "needsDbUpgrade": false,
+     "productname": "Nextcloud",
+     "extendedSupport": false
+ }
+
+And finally, the ``-e`` (for exit code) parameter can be used to check
```
nickvergessen, line 1132: I wouldn't mind a heading above this so we can link to it 🤔

leegarrett: I've added a heading, is this the way you'd like it?

nickvergessen (approved):

nickvergessen:
Looks good:

grafik

kesselb:
Hi @leegarrett,

Could you sign off the commits?

```
git rebase HEAD~2 --signoff
git push --force-with-lease origin master
```

leegarrett:
> Could you sign off the commits?

Done.

Issue #9582, opened 2023-01-24
`make html` fails with missing composer


leegarrett:
Trying to build the docs according to the section `Building HTML` fails due to missing composer. First tripfall:

```
$ pipenv --python 3.9
Creating a virtualenv for this project…
Using /usr/bin/python3.9 (3.9.2) to create virtualenv…
⠋created virtual environment CPython3.9.2.final.0-64 in 180ms
  creator CPython3Posix(dest=/home/randall/.local/share/virtualenvs/documentation-Z2RAW0Gc, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/randall/.local/share/virtualenv)
    added seed packages: pip==20.3.4, pkg_resources==0.0.0, setuptools==44.1.1, wheel==0.34.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

Virtualenv location: /home/randall/.local/share/virtualenvs/documentation-Z2RAW0Gc
Warning: Your Pipfile requires python_version 3.8, but you are using 3.9.2 (/home/r/.local/share/v/d/bin/python).
  $ pipenv check will surely fail.
```

I had to edit `./Pipfile` to `python_version = "3.9"` for it to then build (this part is trivial to fix).

However, now build fails due to missing `composer`. It is unclear to me why composer is needed to generate the static HTML files. If it is required, it should be documented in the build instructions.
```
$ make html
[...]
build succeeded, 16 warnings.

The HTML pages are in _build/html/zh_TW.
Build finished for LANG=zh_TW. The HTML pages are in _build/html/zh_TW.
mv _build/html/en/_static/ .
mv _build/html/en/_images/ .
cd _build/html/ && rm -rf `find -type d -name _static`
cd _build/html/ && rm -rf `find -type d -name _images`
mv ./_static _build/html/
mv ./_images _build/html/
cd _build/html/ && find ./ -type f -exec sed -i -e 's/_images/..\/_images/g' {} \;
cd _build/html/ && find ./ -type f -exec sed -i -e 's/_static/..\/_static/g' {} \;
make[2]: Leaving directory '/home/randall/Coding/nextcloud/documentation/user_manual'
make[1]: Leaving directory '/home/randall/Coding/nextcloud/documentation/user_manual'
User manual build finished; HTML is updated
[...]

PR #9584, opened 2023-01-24
Fix python version to match documented

leegarrett:


tflidd:
You mean the one documented in the README? https://github.com/nextcloud/documentation/commit/4c344bdcf5bb9570fb17710bdfefa8219b140aaa

and it is to fix https://github.com/nextcloud/documentation/issues/9582?

leegarrett:
This only fixes the discrepancy of python version to use between documentation and build. Previously, the documentation required 3.9, but the Pipfile required 3.8.

Regarding #9582, this does fix the comment about Python I made there, but not the issue that composer is required and undocumented.

PR #9591, opened 2023-01-26
Update externalapi.rst

lil5:
The api does not actually return status 100 on an **ok** response

Close https://github.com/nextcloud/server/issues/30163

@szaimen 

Signed-off-by: Lucian I. Last <li@last.nl>


szaimen:
Thanks for the PR! LGTM 👍

ChristophWurst:
I think there is a distinction between OCS V1 and V2

lil5:
@ChristophWurst maybe on paper but on nextcloud both link to the same functions.

ChristophWurst (approved):
:+1:

PR #9592, opened 2023-01-27
Remove docs for svg api, it's not provided anymore + hint for material-design-icons

datenangebot:
Signed-off-by: Florian <florian@nextcloud.com>


susnux:
Maybe instead of linking to the svg wrapper (or additonally) suggest using https://www.npmjs.com/package/vue-material-design-icons
This is the common way to use MDI on vue apps.

PR #9274, opened 2022-11-01
64-bit is required for NC26

szaimen:
Signed-off-by: szaimen <szaimen@e.mail.de>

PR #9615, opened 2023-02-07
Revert "64-bit is required for NC26"

come-nc:
Reverts nextcloud/documentation#9274
See https://github.com/nextcloud/server/issues/36072

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -15,7 +15,7 @@ For best performance, stability and functionality we have documented some recomm
 | Platform         | Options                                                               |
 +==================+=======================================================================+
 | Operating System | - **Ubuntu 22.04 LTS** (recommended)                                  |
-| (64bit)          | - Ubuntu 20.04 LTS                                                    |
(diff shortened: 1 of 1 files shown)


admin_manual/installation/system_requirements.rst
```diff
@@ -39,7 +39,7 @@ See :doc:`source_installation` for minimum PHP-modules and additional software f
 
 CPU Architecture and OS
 ^^^^^^^^^^^^^^^^^^^^^^^
-A 64-bit CPU and 64-bit OS is required for Nextcloud 26 and higher.
+A 64bit CPU and 64bit OS is required for Nextcloud to run well.
```
szaimen, line 42:
```suggestion
A 64-bit CPU and 64-bit OS is required for Nextcloud to run well.
```

szaimen:
Maybe remove the `(64-bit)` in the table?

come-nc:
> Maybe remove the `(64-bit)` fron the table?

You can open another PR for improvements but the point here is only to revert the 32bit drop information from documentation.

szaimen (approved):
LGTM

PR #9629, opened 2023-02-13
Clarify shares situation for disabled users

come-nc:
Signed-off-by: Côme Chilliet <91878298+come-nc@users.noreply.github.com>

### DIFF
--- a/admin_manual/configuration_user/user_configuration.rst
+++ b/admin_manual/configuration_user/user_configuration.rst
@@ -193,8 +193,8 @@ Hover your cursor over their name on the **Users** page until the "..."-menu ico
 appears at the far right. After clicking on it, you will see the **Disable** option.
 
 The user will not longer be able to access their Nextcloud until you enable them again.
-Keep in mind that the files, which were publicly shared by link by this user will no longer be accessible.
-Files shared internally with other users will stay accessible by those users.
+Also all external shares, via public link or email, will not be accessible.
+Internal shares will still be working, so that other users on Nextcloud can continue working.
 
 You will find all disabled users in the **disabled**-section on the left pane.
 Enabling users is as easy as disabling them. Just click on the "..."-menu, and
--- a/admin_manual/configuration_user/user_configuration.rst
+++ b/admin_manual/configuration_user/user_configuration.rst
@@ -193,7 +193,8 @@ Hover your cursor over their name on the **Users** page until the "..."-menu ico
 appears at the far right. After clicking on it, you will see the **Disable** option.
 
 The user will not longer be able to access their Nextcloud until you enable them again.
-Keep in mind that the files, which were shared by this user will not longer be accessible.
+Keep in mind that the files, which were publicly shared by link by this user will no longer be accessible.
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_user/user_configuration.rst
```diff
@@ -193,7 +193,8 @@ Hover your cursor over their name on the **Users** page until the "..."-menu ico
 appears at the far right. After clicking on it, you will see the **Disable** option.
 
 The user will not longer be able to access their Nextcloud until you enable them again.
-Keep in mind that the files, which were shared by this user will not longer be accessible.
+Keep in mind that the files, which were publicly shared by link by this user will no longer be accessible.
+Files shared internally with other users will stay accessible by those users.
```
szaimen, line 197:
```suggestion
Files shared internally with other users will stay accessible and internal sharing will still be possible.
```

admin_manual/configuration_user/user_configuration.rst
```diff
@@ -193,7 +193,8 @@ Hover your cursor over their name on the **Users** page until the "..."-menu ico
 appears at the far right. After clicking on it, you will see the **Disable** option.
 
 The user will not longer be able to access their Nextcloud until you enable them again.
-Keep in mind that the files, which were shared by this user will not longer be accessible.
+Keep in mind that the files, which were publicly shared by link by this user will no longer be accessible.
```
szaimen, line 196: what about email?

admin_manual/configuration_user/user_configuration.rst
come-nc: What does that mean?

admin_manual/configuration_user/user_configuration.rst
come-nc: No idea :shrug: 
Email sharing is just link sharing with an email sending added no?

admin_manual/configuration_user/user_configuration.rst
szaimen: I think the user is still listed in internal sharing so even sharing to it should still be possible?

admin_manual/configuration_user/user_configuration.rst
szaimen: Yes but maybe make it more explicit

admin_manual/configuration_user/user_configuration.rst
tobiasKaminsky: When user is disabled they do not show up in search list when sharing.
But they are shown within Talk.

admin_manual/configuration_user/user_configuration.rst
tobiasKaminsky: "keep in mind" sounds like something unexpected that user/admin have to remember.
But actually that is the expected behaviour.
So maybe phrase it like this:
> The user will not longer be able to access their Nextcloud until you enable them again. Also all external shares, via public link or email, will not be accessible.
Internal shares will still be working, so that other users on Nextcloud can continue working.

ChristophWurst (approved):

PR #9639, opened 2023-02-16
Add note about Group Folders

solracsf:
Using Group Folders, files are visible in the Users trash, but such files are not restored by this command.

### DIFF
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -1109,7 +1109,7 @@ Trashbin
 ::
 
  trashbin
-  trashbin:cleanup  [--all-users] [--] [<user_id>...]  Remove deleted files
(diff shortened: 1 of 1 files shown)


come-nc (approved):

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1146,6 +1146,10 @@ This example restores the deleted files of all users::
 This example restores the deleted files of users molly and freda::
 
  sudo -u www-data php occ trashbin:restore molly freda
+ 
+.. note::
+  This command only restores Users deleted files.
+  If you're using Group Folders app, such files are not restored.
```
come-nc, line 1152: Is this intentional BTW?
Is there another command to restore groupfolder deleted files?

solracsf: I don't think so, but yeah, this is a problem beacause ATM only WebUI can restore files (and browser == timeouts for big Trashbin) 😢

PR #9646, opened 2023-02-21
Typo in class loader `PSR-4` instead of `PSR-0`

susnux:


ChristophWurst (approved):
Excellent catch

ChristophWurst:
Added with https://github.com/nextcloud/documentation/pull/9573. No further backport required.

PR #9657, opened 2023-02-23
Update calendar.rst

KoMa1012:
Changed P1D to PT1D, P1D causes an error.


tflidd:
Normally the "T" is just if

> If the duration contains time elements, that portion of the specification is preceded by the letter T.

and they differentiate time elements and date elements, e.g. P1M is 1 month, and PT1M is 1 minute.

And I don't get an error, if I pass the command:
`Config value calendarSubscriptionRefreshRate for app dav set to P1D`

kesselb:

I also think that `P1D` is correct.

PR #9658, opened 2023-02-24
Add note about dependency limitation for apps

AndyScherzinger:
... with respect to supporting Nextloud server versions.

### DIFF
--- a/developer_manual/app_development/info.rst
+++ b/developer_manual/app_development/info.rst
@@ -270,6 +270,9 @@ dependencies/nextcloud
     * if absent white-listed owncloud versions will be taken from the owncloud element (see below)
     * must contain a **min-version** attribute (maximum 3 digits separated by dots)
     * can contain a **max-version** attribute (maximum 3 digits separated by dots)
+	
+.. note:: Dependencies `dependencies/php`, `dependencies/database` and `dependencies/lib` are checked at installation time (not on update time), hence applications need to stick to the dependencies supported by a major version of Nextcloud the moment an app releases support for that version, i.e. app needs to support the same PHP version-range the supported Nextcloud version supports.
+	
 background-jobs/job
     * optional
     * must contain a php class which is run as background jobs


ChristophWurst (changes requested):
`dependencies/php`, `dependencies/database` and `dependencies/lib` should be noted to not have any effect. Once info.xsd is updated to remove these xml elements we should drop their documentation here

AndyScherzinger:
@ChristophWurst better this way?

nickvergessen:
> dependencies/database … should be noted to not have any effect. Once info.xsd is updated to remove these xml elements we should drop their documentation here

database is needed, to indicate oracle as supported or not (most community apps don't). It is also okay to support more databases going forward, but you can not drop any basically.

And e.g. specifying DBs but not oci will allow you to bypass column lengths and other things:
https://github.com/nextcloud/server/blob/d151999fc8f31f7c9e76e9aa07a0ca50a9d5bdd6/lib/private/DB/MigrationService.php#L84-L96

nickvergessen:
> Do we have a way to deprecate them else when dropped any use would be "illegal" while I can't tell if we actually schema-validate the info.xml

If we remove them from the XSD all apps will break and you can no longer upload to the appstore. So not sure we want to do that. At least not now.
I think we should talk about this at the next occasion instead and draft a proper plan.
But mostlikely it makes sense to just keep them (at least database and lib) and just make sure authors understand that it can break things on updates if you change requirements and they don't cover the stock anymore.

ChristophWurst:
Then let's do like any other API removal

1) Make it a noop and mention it in the upgrade docs of 26 and on the metadata page
2) Remove it in the 26+3=Nextcloud 29 release

AndyScherzinger:
In any way, fine to merge @nickvergessen @ChristophWurst ?

nickvergessen (approved):

ChristophWurst (approved):

PR #9742, opened 2023-03-08
feat(bruteforce): Document the new BruteForceProtection attribute

nickvergessen:
grafik

### DIFF
--- a/developer_manual/basics/controllers.rst
+++ b/developer_manual/basics/controllers.rst
@@ -881,35 +881,69 @@ Nextcloud supports brute-force protection on an action basis. By default control
 
 The native brute-force protection will slow down requests if too many violations have been found. This slow down will be applied to all requests against a brute-force protected controller with the same action from the affected IP.
 
-To enable brute force protection the following *Annotation* can be added to the controller:
+To enable brute force protection the following *Attribute* can be added to the controller:
 
-* **@BruteForceProtection(action=string)**: "string" is the name of the action. Such as "login" or "reset". Brute-force attempts are on a per-action basis; this means if a violation for the "login" action is triggered, other actions such as "reset" or "foobar" are not affected.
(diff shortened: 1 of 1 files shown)


nickvergessen:
Server PR https://github.com/nextcloud/server/pull/36928 is long in, so ready for review

nickvergessen:
Added `:emphasize-lines` :)

developer_manual/basics/controllers.rst
```diff
@@ -881,36 +881,70 @@ Nextcloud supports brute-force protection on an action basis. By default control
...
+* ``#[BruteForceProtection(action: 'string')]``: "string" is the name of the action. Such as "login" or "reset". Brute-force attempts are on a per-action basis; this means if a violation for the "login" action is triggered, other actions such as "reset" or "foobar" are not affected.
 
-Then the **throttle()** method has to be called on the response in case of a violation. Doing so will increase the throttle counter and make following requests slower.
+.. note::
+
+    The attribute is only available in Nextcloud 27 or later. In older versions the ``@BruteForceProtection(action=string)`` annotation can be used, but that does not allow multiple assignments to a single controller method.
+
+Then the **throttle()** method has to be called on the response in case of a violation. Doing so will increase the throttle counter and make following requests slower, until a slowness of roughly 30 seconds is reached and the controller returns a ``429 Too Many Requests`` status is returned instead of avoid further requests.
```
come-nc, line 892:
```suggestion
Then the **throttle()** method has to be called on the response in case of a violation. Doing so will increase the throttle counter and make following requests slower, until a slowness of roughly 30 seconds is reached and the controller returns a ``429 Too Many Requests`` status to avoid further requests.
```
Not sure my fix is the right one but the sentence is definitely broken

nickvergessen: Replacing with:
```suggestion
Then the **throttle()** method has to be called on the response in case of a violation. Doing so will increase the throttle counter and make following requests slower, until a slowness of roughly 30 seconds is reached and the controller returns a ``429 Too Many Requests`` status without further processing the request.
```

I guess that makes the most sense :)

come-nc (approved):

PR #9906, opened 2023-03-16
docs(devel): Add note about updated PSR-11 dependency

susnux:
Documentation for nextcloud/server#36617

### DIFF
--- a/developer_manual/digging_deeper/psr.rst
+++ b/developer_manual/digging_deeper/psr.rst
@@ -45,6 +45,11 @@ PSR-11: Container Interface
 
 The dependency injection container follows the `PSR-11`_ container interface, so you may type-hint ``\Psr\Container\ContainerInterface`` whenever you want an instance of a container and use ``has($id)`` to check for existence and ``get($id)`` to retrieve an instance of a service. See the :ref:`dependency injection docs <dependency-injection>` for details.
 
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

developer_manual/digging_deeper/psr.rst
```diff
@@ -45,6 +45,9 @@ PSR-11: Container Interface
 
 The dependency injection container follows the `PSR-11`_ container interface, so you may type-hint ``\Psr\Container\ContainerInterface`` whenever you want an instance of a container and use ``has($id)`` to check for existence and ``get($id)`` to retrieve an instance of a service. See the :ref:`dependency injection docs <dependency-injection>` for details.
 
+.. versionchanged:: 27
+    Nextcloud ships version 2.0.2
```
ChristophWurst, line 49:
```suggestion
    Package updated from v1.1.1 to v2.0.2
```

ChristophWurst (approved):
Nit: also tell what version was used before

developer_manual/digging_deeper/psr.rst
susnux: Like with the logger interface I have now added the version history.

ChristophWurst (approved):

PR #9913, opened 2023-03-17
Fix a wrong „use“

7408647n:
The namespace is another: https://github.com/nextcloud/server/blob/acca2bdf20a32a4e6113fbab03861aa6639163e8/lib/public/Http/Client/IClientService.php


ChristophWurst:
wrong button?

7408647n:
Yes, sorry, can I undo this or do I need to create a new commit?

ChristophWurst:
It's fine :)

PR #9958, opened 2023-03-28
Update linux_database_configuration.rst

wilfile:
Modification to configuration settings for SSL certs for MySQL servers.

### DIFF
--- a/admin_manual/configuration_database/linux_database_configuration.rst
+++ b/admin_manual/configuration_database/linux_database_configuration.rst
@@ -170,12 +170,13 @@ This just covers the SSL database configuration on the Nextcloud server. First y
 
 ::
 
-  'dbdriveroptions' => [
-    \PDO::MYSQL_ATTR_SSL_KEY => '/../ssl-key.pem',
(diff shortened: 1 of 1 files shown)


tflidd:
The array keyword looks a bit strange, it is not required in other configurations lists/arrays:

https://github.com/nextcloud/server/blob/master/config/config.sample.php

Also you would need to change the above file, @juliushaertl did the last changes to this part. Perhaps he can help to clear this up.

juliushaertl:
The `array()` syntax would also work but I'd prefer the shorthand `[]`

kesselb:

For mysql the pdo_mysql adapter is used: https://github.com/nextcloud/server/blob/2269f3584245bd6d361daf4acf33ffd054f96b60/lib/private/DB/ConnectionFactory.php#L58

Therefore, `PDO::MYSQL_ATTR_SSL_KEY` is the right value: https://www.php.net/manual/en/ref.pdo-mysql.php#pdo.constants.mysql-attr-ssl-key

PR #9959, opened 2023-03-28
[RFC] Style guide and README extensions and improvements

silopolis:
Hi,

So, as I'm new to this project, I sincerely hope you won't mind this proposal... This is actually what's got out of a rabit hole I felt into, originaly just coming to make few translation triggered edits !

I know there are a couple of changes that may be bold or controversial at best, like choice of Google style guide as reference and use of semantic line breaks, but I believe they bring enough to the table to be worth taking the risk of being rejected...

For the rest, I hope the structure rework, added content and formatting will suit you all.

TY
J

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,167 +2,302 @@
 Nextcloud Documentation
 =======================
 
-Documentation is published on `<https://docs.nextcloud.com>`_.
-To edit it yourself, you need to tinker a bit with Git and Sphinx.
-See the `Style Guide <https://github.com/nextcloud/documentation/blob/master/style_guide.rst>`_ for formatting and style conventions.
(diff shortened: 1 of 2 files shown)


silopolis:
> well, if there is a suggested line break after 80 characters in the style guide and the style guide does not respect that, it's a bit strange.
> 

The proposal here is to use semantic line breaks, see line 88 of the style guide. The 80 char wrap line is there in comment only until approval of this, or as fallback.

> The google style guide is not bad, it gives some best practices with a lot of examples.
> 
> I didn't read through all the details, for most parts it looks that everything is restructured and cleaned up rather than changing things (except the parts you mentioned)?

That is indeed most of the work, and a bit more :)

Issue #9964, opened 2023-03-30
Document deprecation of PHP8.0


Fenn-CS:
PHP8.0 deprecation is [completed](https://github.com/nextcloud/server/pull/37451) and now has to be documented here.

PR #9966, opened 2023-03-30
Note PHP8.0 deprecation

Fenn-CS:
PHP8.0 is now deprecated in Nextcloud 27.

Resolves: #9964 
Ref : https://github.com/nextcloud/server/pull/37451

### DIFF
--- a/admin_manual/release_schedule.rst
+++ b/admin_manual/release_schedule.rst
@@ -19,7 +19,7 @@ Maintenance releases are scheduled in a 4 week cycle with one week before the re
 Critical changes
 ----------------
 
-* PHP 7.4 is not supported anymore. Please upgrade to PHP 8.0 or higher.
+* PHP 8.0 is now deprecated. Please upgrade to PHP 8.1 or higher.
 * PHP 8.2 is now supported.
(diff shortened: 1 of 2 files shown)


admin_manual/release_schedule.rst
```diff
@@ -19,7 +19,7 @@ Maintenance releases are scheduled in a 4 week cycle with one week before the re
 Critical changes
 ----------------
 
-* PHP 7.4 is not supported anymore. Please upgrade to PHP 8.0 or higher.
+* PHP 8.0 is now deprecated. Please upgrade to PHP 8.1 or higher.
```
Fenn-CS, line 22: I wonder if this is enough documentation for the deprecation and or if this is the only worthy of nothing this but I guessed this would be a good place to start!

nickvergessen: I would put it here as well:
https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html
Add a note like `**(deprecated)**` on the PHP 8.0 one

Fenn-CS: Right! Thanks!

nickvergessen (approved):

Pytal (approved):

PR #10001, opened 2023-04-04
feat(devmanual): Add go.php shortcut for the backports doc

ChristophWurst:
This will allow the URL https://docs.nextcloud.com/server/stable/go.php?to=developer-backports


christianlupus (approved):
Statically, it looks good, although I have not managed to test this change E2E in a test environment.

Issue #7162, opened 2021-10-01
icons documentation file broken


code1x1:
at the moment the [icon.rst](https://github.com/nextcloud/documentation/blob/86b85d18451821573e848d009521c0acac6dd3b5/developer_manual/design/icons.rst) file looks pretty emtpy. the generator for that file seams to be broken [generateIconsDoc.php](https://github.com/nextcloud/documentation/blob/e73414079cfc933b0c5d7b010359576d009bdb5d/build/generateIconsDoc.php)

code1x1:
https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/icons.html

StCyr:
still empty 😞

provokateurin:
I gave it a try, but even after fixing the scss compilation the resulting file only contained one single seemingly valid icon which must be wrong.

provokateurin:
I'd vote for removing the documentation for now as it is completely broken and stands in the way for other things.

provokateurin:
Or we give this a higher priority and fix it.

joshtrichards:
See https://github.com/nextcloud/documentation/issues/7649#issuecomment-1740892764

PR #10006, opened 2023-04-05
Add new OCS API documentation

provokateurin:
I wanted to link to the static file directly from the TOC, but it's not possible :/
Please don't merge yet, as the spec itself is not finalized.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "build/openapi-extractor"]
+	path = build/openapi-extractor
+	url = https://github.com/nextcloud/openapi-extractor
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ user-manual-html:
 	cd user_manual && make html
 	@echo "User manual build finished; HTML is updated"
 
-developer-manual-html: icons-docs
+developer-manual-html: openapi-spec icons-docs
 	rm -rf developer_manual/_build/html/com
 	cd developer_manual && make html
 	@echo "Developer manual build finished; HTML is updated"
@@ -26,8 +26,22 @@ user-manual-pdf:
 	cd user_manual && make latexpdf
 	@echo "User manual build finished; PDF is updated"
 
-icons-docs: clean-icons-docs
+get-server-sources:
 	cd build && sh get-server-sources.sh $(DRONE_BRANCH)
+
+openapi-spec: get-server-sources
+	git submodule update --init
+	cd build/openapi-extractor && composer install
+	cd build && ./openapi-extractor/merge-specs \
+		--core server/core/openapi.json \
+		--merged ../developer_manual/_static/openapi.json \
+		$$(ls server/apps/*/openapi.json)
+	cd developer_manual/_static && \
+		wget https://unpkg.com/@stoplight/elements@7.7.17/web-components.min.js -O stoplight-elements.js && \
+		wget https://unpkg.com/@stoplight/elements@7.7.17/styles.min.css -O stoplight-elements.css
+
+
+icons-docs: clean-icons-docs get-server-sources
 	cd build && composer install && composer update
 	cd build && php generateIconsDoc.php
 
--- a/developer_manual/_static/openapi.html
+++ b/developer_manual/_static/openapi.html
@@ -1,0 +1,10 @@
+<html>
+    <head>
+        <title>OCS API</title>
+        <script src="stoplight-elements.js"></script>
+        <link rel="stylesheet" href="stoplight-elements.css">
+    </head>
+    <body>
+        <elements-api apiDescriptionUrl="openapi.json" router="hash" hideTryIt="true" logo="logo-blue.png"></elements-api>
+    </body>
+</html>
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,11 @@ _build
 developer_manual/html_css_design/img
 developer_manual/html_css_design/icons.txt
 
+# OpenAPI file
+developer_manual/_static/openapi.json
+developer_manual/_static/stoplight-elements.js
+developer_manual/_static/stoplight-elements.css
+
 # Exclude Eclipse project
 .project
 
--- a/developer_manual/client_apis/OCS/index.rst
+++ b/developer_manual/client_apis/OCS/index.rst
@@ -4,6 +4,10 @@
 OCS API
 ===============================
 
+To browse the new OCS API documentation please go `here <../../_static/openapi.html>`_.
+
+The old documentation is still kept as it provides some additional documentation that is not completely covered in the new documentation:
+
 .. toctree::
 
    ocs-api-overview
--- a/developer_manual/conf.py
+++ b/developer_manual/conf.py
@@ -126,7 +126,7 @@ html_short_title = "Developer Manual"
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['../_shared_assets/static']
+html_static_path = ['../_shared_assets/static', '_static']
 
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.


developer_manual/client_apis/activity-api.rst
```diff
@@ -1,5 +0,0 @@
-============
-Activity API
-============
-
-Find the documentation in the `Activity app repo <https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md>`__.
```
nickvergessen, line 5: This was added intentional as people could not find it otherwise.

developer_manual/client_apis/RemoteWipe/index.rst
```diff
@@ -1,52 +0,0 @@
...
-then the client should proceed to wipe the device.
-
-
-Wiping the actual device
-------------------------
-
-The client must remove all user data linked to the account. This includes:
-  * caches 
-  * offline files
-  * the actual account itself
```
nickvergessen, line 40: This is a lot more information then normal API documentation. Are you sure it's reflected in the openapi stuff?

developer_manual/client_apis/OCS/ocs-api-overview.rst
```diff
@@ -1,239 +0,0 @@
...
-
-	POST /ocs/v2.php/apps/dav/api/v1/direct
-
-With the :code:`fileId` in the body (so :code:`fileId=42` for example).
-This will then return you the link to use to obtain the file.
-
-Notifications
--------------
-
-There is also the `Notifications API <https://github.com/nextcloud/notifications/blob/master/docs/ocs-endpoint-v2.md>`_
```
nickvergessen, line 208: Should also be kept

developer_manual/_static/openapi.json
nickvergessen, line 1: 👎🏼 

Similar to the icons this needs to be generated from the server source at build time, otherwise it does not scale.

nickvergessen (changes requested):

developer_manual/client_apis/OCS/ocs-api-overview.rst
provokateurin: The notifications API is well documented using OpenAPI. I can add all the extra information from the markdown docs to the code so it'll show up.

developer_manual/client_apis/RemoteWipe/index.rst
provokateurin: Depends on https://github.com/stoplightio/elements/issues/2267

developer_manual/_static/openapi.json
provokateurin: Done, but currently depends on a private submodule.

developer_manual/client_apis/OCS/ocs-api-overview.rst
provokateurin: Notifications is not part of the spec yet, so I'll revert it (it is documented but not generated).

Makefile
```diff
@@ -26,11 +26,17 @@ user-manual-pdf:
...
-icons-docs: clean-icons-docs
+get-server-sources:
 	cd build && sh get-server-sources.sh $(DRONE_BRANCH)
+
+icons-docs: clean-icons-docs get-server-sources
 	cd build && composer install && composer update
 	cd build && php generateIconsDoc.php
 
+openapi-spec: get-server-sources
+	cd build/openapi-extractor && composer install
```
juliushaertl, line 37: Should the makefile maybe initialize the git submodule as well?

developer_manual/_static/openapi.json
```diff
@@ -0,0 +1,16975 @@
+{
```
juliushaertl, line 1: This file seems to be the build asset of the merge step, right? Just double checking if we actually should commit it or rely on always building?

Makefile
juliushaertl, line 37: I think we have a conflicting requirement for php versions now between
- build/composer.json (requiring 7.4)
- build/openapi-extractor/composer.json (requiring 8.0+)

We can probably just drop these two dependencies in build/composer.json as they are no longer used:

     11         "juliushaertl/phpdoc-to-rst": "dev-php7.0",
     12         "phpdocumentor/reflection": "dev-php7.0",



I noticed this when just running `make developer-manual-html` in the root, which is not catched by CI it seems. Maybe we can make sure that there is actually a github action triggering the icon-docs and openapi-spec makefile targets.

developer_manual/_static/openapi.json
provokateurin: Ah yeah better to not commit it.

Makefile
provokateurin: Good catch. They are indeed not needed, but upgrading scssphp to a php 8.x version brings a lot of troubles :/ Still working on that...

provokateurin: Even installing PHP7.4 and running the current code doesn't work for me...

provokateurin: @juliushaertl I simply disabled generating the icons because it is broken anyway https://github.com/nextcloud/documentation/issues/7162. I don't like to leave it like that but for the tasks this seems to be the easiest because the icons-docs seem extremely broken.N ot only php7 and outdated dependencies, even with those fixed the output is complete garbage and there must be something else that is wrong.

Makefile
```diff
@@ -13,7 +13,8 @@ user-manual-html:
 	cd user_manual && make html
 	@echo "User manual build finished; HTML is updated"
 
-developer-manual-html: icons-docs
+# Should also build icons-docs, but due to https://github.com/nextcloud/documentation/issues/7162 it is disabled
+developer-manual-html: openapi-spec #icons-docs
```
nickvergessen, line 17:
```suggestion
developer-manual-html: openapi-spec icons-docs
```

*cough*

build/generate-openapi.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -euxo pipefail
```
nickvergessen, line 2: generate-openapi.sh: 2: set: Illegal option -o pipefail

What requirement changed? New pipenv, python or something required?

developer_manual/client_apis/OCS/ocs-translation-api.rst
```diff
@@ -1,84 +0,0 @@
...
-+--------------+--------+--------------------------------------------------------------+
-|``to``        | string | ISO code of the "to" language                                |
-+--------------+--------+--------------------------------------------------------------+
-|``toLabel``   | string | Name of the "to" language that should be shown to the user   |
-+--------------+--------+--------------------------------------------------------------+
-
-Translate a string
-------------------
-
-.. versionadded:: 26
```
nickvergessen, line 54: Min version is not available in openapi

developer_manual/client_apis/OCS/ocs-translation-api.rst
```diff
@@ -1,84 +0,0 @@
...
-|``text``         | string      | The text to be translated                                                                                                                                                    |
-+-----------------+-------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-|``fromLanguage`` | string/null | The ISO code of the "from" language, when null is given and a translation provider allows detecting the source language it will be tried to guess it from the ``text`` input |
-+-----------------+-------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-|``toLanguage``   | string      | The ISO code of the "to" language                                                                                                                                            |
```
nickvergessen, line 69: "ISO code" information is missing in openapi

nickvergessen (changes requested):
Only checked the translation thing.

Maybe we can just keep the "old docs" in parallel until the transition of all the available information to the openapi spec was done?

Makefile
provokateurin: I know, but the icons-docs are broken. Easier to disable this and keep the php8 from openapi-extractor.

developer_manual/_static/openapi.html
```diff
@@ -0,0 +1,10 @@
+<html>
+    <head>
+        <title>OCS API</title>
+        <script src="https://unpkg.com/@stoplight/elements/web-components.min.js"></script>
+        <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements/styles.min.css">
```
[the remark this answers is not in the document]
provokateurin: Sure, best to install and link via npm?

Makefile
nickvergessen: Not sure what the builder that does it for the releases has installed

developer_manual/_static/openapi.html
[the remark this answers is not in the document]
nickvergessen: Install with pip/python so it still only requires one stack? Otherwise download and put into a static folder?

developer_manual/digging_deeper/translation.rst
```diff
@@ -18,7 +18,7 @@ To consume the Translation API, you will need to :ref:`inject<dependency-injecti
  * ``translate(string $text, ?string $fromLanguage, string $toLanguage)`` This method provides the actual translation functionality. Note that, depending on the length of the text you want to translate, this may take longer than the HTTP request timeout or the PHP execution time limit.
  * ``canDetectLanguage()`` This method returns a boolean indicating whether language auto-detection is possible. If this is true, you can pass ``null`` as a ``$fromLanguage`` parameter to ``translate`` and it will automatically figure out the source language.
 
-If you would like to use the translation functionality in a client, there are also OCS endpoints available for this: :ref:`OCS Translation API<ocs-translation-api>`
+If you would like to use the translation functionality in a client, there are also OCS endpoints available for this: :ref:`OCS Translation API<ocsapiindex>`
```
nickvergessen, line 21: Revert as well?

developer_manual/_static/openapi.html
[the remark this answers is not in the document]
provokateurin: If just downloading the file is fine then that is the easiest option

developer_manual/digging_deeper/translation.rst
provokateurin: Ah forgot that

developer_manual/_static/openapi.html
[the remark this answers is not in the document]
nickvergessen: At least for now to get it live privacyaware asap I guess so.
A packaged and therefore update-checked version in the future would be preferred I guess, so raise a follow up for after 27? 😎

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "build/openapi-extractor"]
```
nickvergessen, line 1: New git submodule and things.

@blizzz can you dive the docker thing from the builder a test run to make sure it doesn't break the next RC docs?

nickvergessen:
For me it does not load:

## Firefox
```
Uncaught SyntaxError: illegal character U+20AC
[stoplight-elements.js:2:422677](http://localhost:63342/documentation/developer_manual/_build/html/com/_static/stoplight-elements.js)
```

### Chrome

```
caught SyntaxError: Invalid or unexpected token
```

provokateurin:
How are you loading it? I think it's a problem with the web server.
Using `python -m http.server` I also get this error, but using https://www.npmjs.com/package/serve works.

juliushaertl (approved):

juliushaertl:
> How are you loading it? I think it's a problem with the web server. Using `python -m http.server` I also get this error, but using [npmjs.com/package/serve](https://www.npmjs.com/package/serve) works.

Also worked fine for me with `php -S 0.0.0.0:8081`

.gitmodules
blizzz: > New git submodule and things.
> 
> @blizzz can you dive the docker thing from the builder a test run to make sure it doesn't break the next RC docs?

at least building the documentation does not fail and the output structured is as expected (as before). The developer manual is not being distributed. So, all is looking fine.

provokateurin:
@nickvergessen Is there something still missing?

come-nc (approved):

come-nc:
I’m failing to build this:
```
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires phpdocumentor/reflection dev-php7.0 -> satisfiable by phpdocumentor/reflection[dev-php7.0].
    - phpdocumentor/reflection dev-php7.0 requires php ^7.0 -> your php version (8.1.2) does not satisfy that requirement.
  Problem 2
    - Root composer.json requires leafo/scssphp dev-master -> satisfiable by leafo/scssphp[dev-master].
    - leafo/scssphp dev-master requires php ^5.4.0 || ^7 -> your php version (8.1.2) does not satisfy that requirement.
  Problem 3
    - phpdocumentor/reflection-docblock 4.3.0 requires php ^7.0 -> your php version (8.1.2) does not satisfy that requirement.
    - juliushaertl/phpdoc-to-rst dev-php7.0 requires phpdocumentor/reflection-docblock 4.3.0@stable -> satisfiable by phpdocumentor/reflection-docblock[4.3.0].
    - Root composer.json requires juliushaertl/phpdoc-to-rst dev-php7.0 -> satisfiable by juliushaertl/phpdoc-to-rst[dev-php7.0].

make: *** [Makefile:45 : icons-docs] Erreur 2
```

provokateurin:
That's the icons, they are broken since a long time. Unrelated to this change. You can comment the `icons-docs` target in the Makefile to skip it.

come-nc:
I tried instead of `make developer-manual-html` to do `make openapi-spec`, then I cd into developer_manual and run `make SPHINXBUILD=sphinx-autobuild html`, but still the link to http://127.0.0.1:8000/_static/openapi.html gives a blank page.

Web console shows:
```
stoplight-elements.js:2 Uncaught SyntaxError: Invalid or unexpected token
favicon.ico:1 GET http://127.0.0.1:8000/favicon.ico 404 (Not Found)
```

provokateurin:
See comments above. Try @juliushaertl's way to serve the documentation.

come-nc:
I tried `php -S 0.0.0.0:8081` but in the main folder I only get the index with links poiting to docs.nextcloud.com, in the developer_manual folder I get `The requested resource / was not found on this server.` When opening the web page.

`make SPHINXBUILD=sphinx-autobuild html` is the documented way if that does not work anymore the README needs to be updated and explain how to test the documentation locally.

juliushaertl:
`make SPHINXBUILD=sphinx-autobuild html` currently only works within the individual documentation folders developer_manual, user_manual, admin_manual. As those are separate sphinx projects I don't see a way to use the autobuild tooling.

While not a nice experience it is documented as
>   2. Enter the documentation section ``cd user_manual``

However the document sections Makefile doesn't cover the developer documentation pre-build steps which are in the main root makefile.

Steps for testing manually:
- run `make developer-manual-html` in the root
- run `php -S 0.0.0.0:8081` in developer_manual/_build/html/com/


I've pushed two commits to unblock this PR, but it would be nice to follow up with a cleaner solution so that either way of building involves the required build steps.

come-nc (approved):

come-nc:
Good enough for me, at least I could test it and it works.

Merge?

provokateurin:
I think @nickvergessen's concerns have been all addressed, but I don't really want to merge @juliushaertl commits in this PR. @juliushaertl could you make a separate PR? I'll happily approve that, but the commits really don't belong here.

juliushaertl:
Removed the commits and pushed them to #10584 as they were based on this one.

provokateurin:
Thanks! :heart:

provokateurin:
Meh, the problem @nickvergessen and @come-nc had with the illegal character thing also happen on the deployed version. Not sure what is going on there, I will investigate it (seems like some character set problem?)

PR #10029, opened 2023-04-09
Fix syntax highlighting of http method

felixschndr:
Add syntax highlighting to the `DELETE` http-method in the API docs

Signed-off-by: Felix Schneider <fs@felix-schneider.org>


kesselb (approved):
Thank you

PR #10032, opened 2023-04-10
Update installation_wizard.rst

kacperpaczos:
Added basic solving problems


tflidd:
I think the advice with the logfiles and linking to that part of the documentation is valid for any problem.

Perhaps we could imagine to put a list of common problems (once we have one).

However, for the reverse proxy there is a dedicated page in the documentation:
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html
wouldn't that be better for your tip?

The installation wizard does not mention a proxy or reverse proxy, it would be strange to mention further down on debugging. Like this, it probably confuses more people.

PR #10035, opened 2023-04-11
Fixes migrations:execute description in migrations.rst.

fsamapoor:
"Version" is the "prefix" of the class name, not the "postfix".


kesselb (approved):
Thank you

PR #10036, opened 2023-04-12
Fix an error in sync_android documentation

ls-1N:
It could be updated more to match the current copywriting of DAVx5, but this fixes a confusing outright mistake.

| Before | After |
|--------|--------|
| Screenshot from 2023-08-03 21-31-53 | Screenshot from 2023-08-03 21-31-02 |


kesselb (approved):
Thank you, well spotted :+1:

PR #10044, opened 2023-04-15
fix webdav fstab entry

dulvui:
The fstab entry of the command-line webdav mount is split in 2 lines, but needs to be a one liner to be mounted correctly.
I followed all the steps to mount Nextcloud as webdav on my Raspberry Pi 1 model B with Raspbian OS Lite but only with a one liner it mounted correctly.


kesselb (approved):
Thank you

PR #10062, opened 2023-04-18
Update access_webdav.rst

Matthionine:
Trying to use ``net use`` while the ``WebClient`` service in Windows is not running will cause the command to fail. This should be mentioned in a note.


joshtrichards:
Hi @Matthionine - Thanks for the submission (on behalf of, well, me, but I suppose also the community).

I ran across someone else with this issue once too. It's unclear if it some sort of default state under some limited circumstances or what.

This location in the doc is as good as any at the moment, but it made it clear to me that some re-organizing of the WebDAV docs would probably be helpful at some point. :-) When that happens this may or may belong as a note in the *Known Problems* section or *Troubleshooting* or something. No reason to hold back this PR on that though (and it'll be easier to reorganize if there aren't outstanding PRs anyhow). 

There's some additional official reference from MS backing this up. Just linking it here in case want to come back later and make any further changes: https://learn.microsoft.com/en-us/iis/publish/using-webdav/using-the-webdav-redirector#troubleshooting-the-webdav-redirector

Hopefully someone with approval status will be along for an official review.

tflidd (approved):

tflidd:
I agree that this should be cleaned up a bit, perhaps also a separation for Win10/11. But that's a different issue.

tflidd:
Let's merge this improvement. It doesn't mean we can't do further improvements in separate PR.
It solves https://github.com/nextcloud/documentation/issues/10207

PR #10137, opened 2023-04-24
chore: add untracked user manual files to gitignore

miaulalala:


nickvergessen:
Unluckily that would break the sync with transifex.
Locally I'm cheating around this and hard coded the Makefile to specifically only build the english language locally (append -lang-en on the user manual call)

Issue #10272, opened 2023-05-05
Building wheel for pycairo (pyproject.toml) did not run successfully.


ChristophWurst:
The documentation can't be built anymore. Neither on CI nor locally nor for deployment.

```
Collecting sphinx<5.0,>=4.5.0
  Using cached Sphinx-4.5.0-py3-none-any.whl (3.1 MB)
Collecting rst2pdf
  Downloading rst2pdf-0.100-py3-none-any.whl (176 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 176.9/176.9 kB 3.2 MB/s eta 0:00:00
Collecting pillow
  Downloading Pillow-9.5.0-cp310-cp310-manylinux_2_28_x86_64.whl (3.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3.4/3.4 MB 4.4 MB/s eta 0:00:00
Collecting sphinx-rtd-theme==1.2.0
  Using cached sphinx_rtd_theme-1.2.0-py2.py3-none-any.whl (2.8 MB)
Collecting sphinxcontrib-phpdomain
  Downloading sphinxcontrib_phpdomain-0.11.0-py3-none-any.whl (10 kB)
Collecting jinja2<3.1.0
  Using cached Jinja2-3.0.3-py3-none-any.whl (133 kB)
Collecting sphinxcontrib-jquery!=3.0.0,>=2.0.0
  Using cached sphinxcontrib_jquery-4.1-py2.py3-none-any.whl (121 kB)
Collecting docutils<0.19
  Using cached docutils-0.18.1-py2.py3-none-any.whl (570 kB)
Collecting requests>=2.5.0
  Downloading requests-2.30.0-py3-none-any.whl (62 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 62.5/62.5 kB 3.6 MB/s eta 0:00:00
Collecting docutils<0.19
  Downloading docutils-0.17.1-py2.py3-none-any.whl (575 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 575.5/575.5 kB 4.4 MB/s eta 0:00:00
Collecting babel>=1.3
  Downloading Babel-2.12.1-py3-none-any.whl (10.1 MB)
[...]

PR #10275, opened 2023-05-05
fix(deps): Require pkgconfig

ChristophWurst:
### ☑️ Resolves

* Fix https://github.com/nextcloud/documentation/issues/10272

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 sphinx>=4.5.0, <5.0
 rst2pdf
 pillow
+pkgconfig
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> ModuleNotFoundError: No module named 'sphinx_rtd_theme'

(╯°□°）╯︵ ┻━┻

mgallien:
@ChristophWurst from https://pypi.org/project/pkgconfig/ it seems the native binary is needed for the python module to succeed

PR #10355, opened 2023-05-10
feat(devmanual): Add user managers doc

ChristophWurst:
### ☑️ Resolves

* For https://github.com/nextcloud/server/pull/38013

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/digging_deeper/users.rst
+++ b/developer_manual/digging_deeper/users.rst
@@ -173,3 +173,82 @@ Then users can be logged in by using:
         }
 
     }
+
+User objects
+------------
+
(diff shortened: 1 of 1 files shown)


julien-nc (approved):
Ouch it's quite confusing to use `IUserManager` to get user objects that have the `getManagerUids` and `setManagerUids` methods.

ChristophWurst:
What would be the alternative? The managers are a property of the user, like their email addresses, location, etc.

ChristophWurst:
Do you mean the *user manager* and the *user's manager*  to be confusing? Then yes :s

PR #10362, opened 2023-05-11
fix: Remove outdated "windows enablers"

ChristophWurst:
### ☑️ Resolves

Remove 10 year old wget binary and ca bundle. First and last update was https://github.com/nextcloud/documentation/commit/598153e0fbc5164a90f5e42ddb4cccfa59d1c59e.

If this is required for a docs build on Windows we should find a better way than shipping a binary that doesn't receive updates.

### 🖼️ Screenshots

N/a


nickvergessen (approved):
🙈

nickvergessen (approved):

PR #10375, opened 2023-05-11
[admin] Nginx: Change mimetype for javascript

susnux:
Change mimetype from `application/javascript` to `text/javascript`
See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript

### ☑️ Resolves

* Fix https://github.com/nextcloud/server/pull/36057#issuecomment-1539658967

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -57,7 +57,7 @@ server {
     gzip_comp_level 4;
     gzip_min_length 256;
     gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
-    gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/wasm application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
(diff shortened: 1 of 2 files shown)


admin_manual/installation/nginx-root.conf.sample
```diff
@@ -57,7 +57,7 @@ server {
     gzip_comp_level 4;
     gzip_min_length 256;
     gzip_proxied expired no-cache no-store private no_last_modified no_etag auth;
-    gzip_types application/atom+xml application/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/wasm application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
+    gzip_types application/atom+xml text/javascript application/json application/ld+json application/manifest+json application/rss+xml application/vnd.geo+json application/vnd.ms-fontobject application/wasm application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/bmp image/svg+xml image/x-icon text/cache-manifest text/css text/plain text/vcard text/vnd.rim.location.xloc text/vtt text/x-component text/x-cross-domain-policy;
```
ChristophWurst, line 60: this only changes which mime have gzip enabled, right? Nextcloud still sends (some) files as application/javascript, e.g. in `\OC\Core\Controller\JsController::getJs` and `\OCA\Files\Controller\ApiController::serviceWorker`.

I'd propose to change the gzip config to compress both `text/javascript` **and** `application/javascript`.

admin_manual/installation/nginx-root.conf.sample
```diff
@@ -86,7 +86,7 @@ server {
     # only for Nextcloud like below:
     include mime.types;
     types {
-        application/javascript js mjs;
+        text/javascript js mjs;
```
ChristophWurst, line 89: this seems to be the actual change? :+1:

susnux: Yes but we need the other one as well to gzip compress it.

admin_manual/installation/nginx-root.conf.sample
susnux: Yes I will adjust it to include both MIME types.

susnux: Done :heavy_check_mark:

ChristophWurst (approved):

PR #10468, opened 2023-05-26
feat(adminmanual): Create critical changes section for admins

ChristophWurst:
* Moves critical changes from *Release schedule* to *Release notes*
* Update all update pages to link to the changes for admins to see before the upgrade

This is the state for 27. We'll need to write the critical changes for 28 still.

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/release_notes.rst
+++ b/admin_manual/release_notes.rst
@@ -2,4 +2,24 @@
 Release notes
 =============
 
-See `the official changelog <https://nextcloud.com/changelog/>`_ for release notes.
+.. _critical-changes:
+
+Critical changes
+----------------
+
+Here you find all important infos about the new release before you upgrade.
+
+System requirements
+^^^^^^^^^^^^^^^^^^^
+
+* PHP8.2 is recommended over PHP8.1.
+
+Exposed system address book
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Nextcloud 27 exposes the :ref:`system address book<system-address-book>`. Restrict the enumeration settings if your users should not see other users.
+
+Changelog
+---------
+
+See `the official changelog <https://nextcloud.com/changelog/>`_ for a complete list of changes.
--- a/admin_manual/maintenance/manual_upgrade.rst
+++ b/admin_manual/maintenance/manual_upgrade.rst
@@ -2,6 +2,10 @@
 Upgrade manually
 ================
 
+.. seealso::
+
+   If you upgrade from a previous major version please see :ref:`critical changes<critical-changes>` first.
+
 Always start by making a fresh backup and disabling all 3rd party apps.
 
 1. Back up your existing Nextcloud Server database, data directory, and 
--- a/admin_manual/maintenance/package_upgrade.rst
+++ b/admin_manual/maintenance/package_upgrade.rst
@@ -58,6 +58,10 @@ using Snappy Base 16.04 as it's currently unreleased.
 Upgrade tips
 ------------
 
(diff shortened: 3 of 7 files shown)


admin_manual/release_notes.rst
```diff
@@ -2,4 +2,24 @@
...
+
+System requirements
+^^^^^^^^^^^^^^^^^^^
+
+* PHP8.2 is recommended over PHP8.1.
+
+Exposed system address book
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Nextcloud 26 exposes the :ref:`system address book<system-address-book>`. Restrict the enumeration settings if your users should not see other users.
```
nickvergessen, line 20: I thought 27 only?

ChristophWurst: Typo

admin_manual/release_notes.rst
nickvergessen, line 1: Should we have this file versioned like the dev changes, so sysadmins can just read the last manual instead of having to jump through all of them?

ChristophWurst: I'm fine with either. I don't remember why I chose to keep all versions for the dev changes. We can also just clear those pages after branch-off of the predecessor.

nickvergessen: I'd prefer to have them all available in the latest version like in the dev manual.
Just handy, gives search results, etc.

ChristophWurst: Ugh. *Release notes* is a standalong page, not a directory with an index. I can't add subpages to it without breaking the URL `/release_notes.html`

ChristophWurst: I tried that too but it makes the left navigation jump due to the inconsistent page table. I think this would also be problematic for the generated PDF. Pages/directories outside the tree might not be rendered.

nickvergessen: > I think this would also be problematic for the generated PDF. Pages/directories outside the tree might not be rendered.

That's fine, the unlisted page is only redirecting to the treed index in the new folder which has the proper tree and list?

ChristophWurst: Approved

nickvergessen (approved):
So merge and backport like this and then add an empty 28 for master as second PR?

ChristophWurst:
Bildschirmfoto vom 2023-05-26 19-01-18

@nickvergessen your trick almost works, but the navigation picks up the standalone file, not the index.

1) Open the docs
2) Click on *Release notes*

Expected -> new index.rst
Actual -> release_notes.rst with the move notice

admin_manual/contents.rst
```diff
@@ -6,7 +6,7 @@ Table of contents
     :maxdepth: 2
 
     index
-    release_notes
+    release_notes/index
```
nickvergessen, line 9:
```suggestion
    upgrade_guide/index
```
Then let's try renaming the folder.
Not sure why it worked locally thou

ChristophWurst: there must be some caching in place. it still points to the old file :see_no_evil:

ChristophWurst: wiping the _build directory and rebuilding makes it work with your changes

ChristophWurst:
Changelog moved back up into the index because it's one link for all versions

admin_manual/contents.rst
nickvergessen: 🎉

nickvergessen (approved):

PR #10479, opened 2023-05-26
docs(dev): remove deprecated createEventSource

kesselb:
### ☑️ Resolves

- [x] Needs https://github.com/nextcloud/server/pull/38386

### 🖼️ Screenshots
Screenshot from 2023-06-01 18-08-15


ChristophWurst (approved):

ChristophWurst (commented):
Could we have a dedicated page for event source and how it's used in an app? :pray:

kesselb:
> Could we have a dedicated page for event source and how it's used in an app?

I think we should review the current implementation first, as the code is ancient. 

To inject IRequest into OC_EventSource was actually a side project for https://github.com/nextcloud/server/pull/38445. 
I did not expect to end up with EventSourceFactory ;)

PR #10517, opened 2023-05-30
fix: add actions/setup-python to build workflow

skjnldsv:


ChristophWurst:
What does this fix?

skjnldsv:
> What does this fix?

self hosted runners. Discussion is ongoing on talk

PR #10554, opened 2023-06-01
feat(devmanual): Document IMailer and inline attachments

ChristophWurst:
### ☑️ Resolves

* For https://github.com/nextcloud/server/pull/36118

### 🖼️ Screenshots

Bildschirmfoto vom 2023-06-01 18-01-21


ChristophWurst:
Note how I didn't set a subject. Turns out `\OCP\Mail\IMessage` doesn't support that.

PR #10575, opened 2023-06-07
fix(admin_manual): Add module js (`mjs`) as static resource to nginx config

susnux:
### ☑️ Resolves

Not sure how this slipped through, but the nginx config was missing `mjs` as static files, adding this so we can finally make use of it.

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -165,7 +165,8 @@ server {
         fastcgi_max_temp_file_size 0;
     }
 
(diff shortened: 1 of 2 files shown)


susnux:
(was missing here: https://github.com/nextcloud/documentation/pull/9909 )

szaimen (approved):
LGTM

szaimen:
@susnux probably this should also be added to some upgrade guide or so?

ChristophWurst:
Yes, we should tell admins to check their config in https://docs.nextcloud.com/server/latest/admin_manual/release_notes/upgrade_to_27.html

PR #10581, opened 2023-06-08
(hardening and security) Add disabling debug mode recommendation

joshtrichards:
### ☑️ Resolves

Enabling `debug` mode at all (and certainly leaving it on) in production environments has security implications. This adds a section about it to the Hardening and security guidance chapter.

### 🖼️ Screenshots

image


juliushaertl (approved):
Great, thanks a lot 👍

PR #10585, opened 2023-06-08
Document missing events

juliushaertl:
Documenting the missing event from https://github.com/nextcloud/server/pull/34799 and https://github.com/nextcloud/server/pull/33818


juliushaertl:
At some point it might also be nice to generate this list from phpdoc annotations in OCP.

ChristophWurst (approved):

PR #10602, opened 2023-06-10
Fix .htaccess configuration to actually work

Jakuje:
The .htaccess rules should not have the starting slash otherwise they will never get matched. This is done correctly when the nextcloud is installed in the root directory, but not in the documentation which proposes solving this common use case.

I actually copied these instructions to my root directory ages ago and wondered why they still do not work. Now I had to spend another hour figuring out why it wont work.

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -268,15 +268,14 @@ document root of your Web server and add the following lines::
 
 Make sure to change /nextcloud to the actual subfolder your Nextcloud instance is running in.
 
-.. note:: Note, that if you are putting the above configureation directly to
-   the apache configuration (generaly in ``/etc/apache2/sites-enabled/``),
-   you will need to prepend the first argument of each of the above
-   ``RewriteRule`` option with a forward slash ``/``, for example
+.. note:: If you put the above directives directly into an Apache
+   configuration file (usually within ``/etc/apache2/``)
+   instead of ``.htaccess``, you need to prepend the first argument of
+   each ``RewriteRule`` option with a forward slash ``/``, for example
    ``^/\.well-known/carddav``.
-   This is because Apache already normalizes the values for the use in
-   ``.htaccess`` by dropping any number of leading slashes, but it does
-   not for the use in its main site configuration where the forward slash
-   is still present.
+   This is because Apache normalizes paths for the use in ``.htaccess``
+   files by dropping any number of leading slashes, but it does not
+   do so for the use in its main configuration files.
 
 If you are running NGINX, make sure ``location = /.well-known/carddav {`` and ``location = /.well-known/caldav {`` are properly configured as described in :doc:`../installation/nginx`, adapt to use a subfolder if necessary.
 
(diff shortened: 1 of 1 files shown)


Jakuje:
> Would you mind taking a look at https://stackoverflow.com/a/64448903 and update the patch accordingly?

Can you clarify what you want to adjust in the patch in that is described in the posted answer?

MichaIng:
Simply skipping the leading slash also matches the `.htaccess` shipped by Nextcloud itself:
```
<IfModule mod_rewrite.c>
  RewriteEngine on
  RewriteCond %{HTTP_USER_AGENT} DavClnt
  RewriteRule ^$ /remote.php/webdav/ [L,R=302]
  RewriteRule .* - [env=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
  RewriteRule ^\.well-known/carddav /remote.php/dav/ [R=301,L]
  RewriteRule ^\.well-known/caldav /remote.php/dav/ [R=301,L]
  RewriteRule ^remote/(.*) remote.php [QSA,L]
  RewriteRule ^(?:build|tests|config|lib|3rdparty|templates)/.* - [R=404,L]
  RewriteRule ^\.well-known/(?!acme-challenge|pki-validation) /index.php [QSA,L]
  RewriteRule ^(?:\.(?!well-known)|autotest|occ|issue|indie|db_|console).* - [R=404,L]
</IfModule>
```
So as this parts of the docs is about adding things to `.htaccess`, it is cleaner and consistent to simply remove it instead of having it optional.

Jakuje:
@kesselb what do you think? Would it make sense to go back to remove the leading slashes and add a note that if the configuration is used from the main apache configuration, they need to be added?

Jakuje:
Updated with a note. Please, let me know if the wording is ok or you would improve something.

admin_manual/issues/general_troubleshooting.rst
```diff
@@ -260,14 +260,24 @@ document root of your Web server and add the following lines::
...
 
+.. note:: Note, that if you are putting the above configureation directly to
+   the apache configuration (generaly in ``/etc/apache2/sites-enabled/``),
+   you will need to prepend the first argument of each of the above
+   ``RewriteRule`` option with a forward slash ``/``, for example
+   ``^/\.well-known/carddav``.
+   This is because Apache already normalizes the values for the use in
+   ``.htaccess`` by dropping any number of leading slashes, but it does
+   not for the use in its main site configuration where the forward slash
+   is still present.
```
MichaIng, line 279: Some typos fixed and wording suggestion:
```suggestion
.. note:: If you put the above directives directly into an Apache
   configuration file (usually within ``/etc/apache2/``)
   instead of ``.htaccess``, you need to prepend the first argument of
   each ``RewriteRule`` option with a forward slash ``/``, for example
   ``^/\.well-known/carddav``.
   This is because Apache normalizes patjs for the use in ``.htaccess``
   files by dropping any number of leading slashes, but it does not
   do so for the use in its main configuration files where hence the
   forward slash is needed.
```
In other words: In `.htaccess` one must always use patterns for paths relative to its location, while in central config files one must use patterns which match the absolute path. Somehow makes sense.

Jakuje:
Thanks. Applied your suggestion. I will squash the changes tomorrow if needed.

kesselb:
image

Thanks :+1:

kesselb:
When you replace "If you put" with "To put" and "you need to prepend" with "prepend" it's still correct, but does not sound so german ;)

kesselb (approved):

PR #10603, opened 2023-06-10
(primary+external storage) S3 bucket naming requirements

joshtrichards:
* Adds note about S3 bucket naming requirements needing to be compliant with AWS S3 rules even if using a third-party S3 provider/platform (in order to work with the aws-sdk-php we use).
* Adds note about S3 endpoint usage for hostname to External Storage doc (to be consistent with Primary Storage doc change recently made in #10595)

### ☑️ Resolves

* Fix nextcloud/server#35364


artonge (approved):
Sounds good

PR #10621, opened 2023-06-13
desktop files client version 3.9 is the latest one

mgallien:
### ☑️ Resolves

* Fix #…


ChristophWurst (approved):

mgallien:
> does this also apply to 27 and older releases? if so, trigger backports

backports are not needed as the server doc uses a link to `latest` version of the client doc
this is just to update the landing page that is always generated from master branch

PR #10637, opened 2023-06-16
feat(adminmanual): Document how to use occ with sudo and env vars

ChristophWurst:
This might be obvious to real sysadmins. It was new to me.

### 🖼️ Screenshots

image


SystemKeeper (approved):

SystemKeeper:
I think it makes sense to mention this indeed.

PR #10644, opened 2023-06-17
add a section to server-tuning about asset compression

szaimen:
After an intense debugging session, I came to the conclusion that javascript, css and svg are not compressed by default because they are not served via Nextcloud's PHP code but instead via the web server. So enabling compression for these file types makes sense. See https://github.com/nextcloud/all-in-one/pull/2576#issuecomment-1595831206

### DIFF
--- a/admin_manual/installation/server_tuning.rst
+++ b/admin_manual/installation/server_tuning.rst
@@ -50,6 +50,14 @@ Caching improves performance by storing data, code, and other objects in memory.
 Memory cache configuration for the Nextcloud server must be installed and configured.
 See :doc:`../configuration_server/caching_configuration`.
 
+Compression
+-----------
+
+Enabling compression in your web server for plain filetypes like javascript, css and svg files
+improves the performance because less bytes need to be transferred to the clients. Only these 
+file types are necessary because all other plain filetypes are usually served by Nextcloud PHP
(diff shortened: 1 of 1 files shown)


admin_manual/installation/server_tuning.rst
```diff
@@ -50,6 +50,14 @@ Caching improves performance by storing data, code, and other objects in memory.
 Memory cache configuration for the Nextcloud server must be installed and configured.
 See :doc:`../configuration_server/caching_configuration`.
 
+Compression
+-----------
+
+Enabling compression in your web server for plain filetypes like javascript, css and svg files
```
kesselb, line 56: > Enabling compression in your web server for JavaScript, CSS, and SVG files improves performance because fewer bytes need to be transferred to the clients.

Simpler version.

> all other plain filetypes are usually served by Nextcloud PHP

I'm not certain. We have the compression middleware for some requests: https://github.com/nextcloud/server/blob/master/lib/private/AppFramework/Middleware/CompressionMiddleware.php

If we don't know for sure, we better remove the second sentence.

szaimen: Done with https://github.com/nextcloud/documentation/pull/10644/commits/c9914dc9c1ae084fe54e1b0c9b71335c3697ce6b

kesselb (approved):

PR #10676, opened 2023-06-27
add a section regarding showing free space on windows

szaimen:
As discussed with @icewind1991


icewind1991 (approved):

icewind1991:
As some extra context: http://support.microsoft.com/kb/2386902 [seems to have said that](https://stackoverflow.com/questions/12855859/windows-7-webdav-network-drive-size-shows-local-drive-size-does-not-let-me-copy) but the link is now dead.

PR #10701, opened 2023-06-30
Fix reference to download page in DE locale

JSchoeck:
### ☑️ Resolves

* Fix reference to Nextcloud download page

### 🖼️ Screenshots

image


tflidd:
Not sure if we are supposed to change that here. These files might be created automatically and the fix needs to be somewhere else.
@rakekniven might be able to help us.

PR #10703, opened 2023-06-30
Add OpenAPI tutorial

provokateurin:
Create a tutorial on how to add OpenAPI to your app.
Will stay WIP for a while until I have gathered some feedback by testing this out with some people.

### DIFF
--- a/developer_manual/client_apis/OCS/ocs-openapi.rst
+++ b/developer_manual/client_apis/OCS/ocs-openapi.rst
@@ -1,0 +1,731 @@
+====================
+OCS OpenAPI tutorial
+====================
+
+This page explains you how you can add OpenAPI support to your app such that you can automatically generated an OpenAPI specification from your server code.
+
+Please read the whole tutorial before starting to adapt your app.
+
+Do not be afraid that you do not know everything from the start.
+The openapi-extractor tool gives you many warnings and fails if there is something utterly broken that would not work.
+Let the tool run and it will tell you if there is something wrong.
+Psalm will also help you validate your changes to ensure that nothing is broken.
+
+Requirements and prerequisites
+------------------------------
+
+App supports at least Nextcloud 28
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Nextcloud 28 is the first version that has the required changes to get OpenAPI support going.
+
+Psalm is setup and configured
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Install psalm in your app as explained in `<https://psalm.dev/docs/running_psalm/installation>`_.
+
+Install and enable the required extensions as explained in :ref:`Required PHP extensions <psalm-php-extensions>`.
+
+Setup the following issue handlers in the psalm config (see `<https://psalm.dev/docs/running_psalm/dealing_with_code_issues>`_ for a tutorial):
+
+.. code-block:: xml
+
+    <LessSpecificReturnStatement errorLevel="error"/>
+    <LessSpecificReturnType errorLevel="error"/>
+    <LessSpecificImplementedReturnType errorLevel="error"/>
+    <MoreSpecificReturnType errorLevel="error"/>
+
+Install openapi-extractor in your app as explained in `<https://github.com/nextcloud/openapi-extractor>`_.
+
+Tips and tricks
+---------------
+
+The openapi-extractor tool forces you to set descriptions everywhere.
+To ease the adoption process you can set the ``--allow-missing-docs`` flag to ignore those problems.
+
+The tool also only shows errors one by one by default, but you can also let it show you all problems at once using the ``--continue-on-error``.
+
+It is advised to use neither of the mentioned flags to generate the final specification.
+They hide problems that your code has.
+Especially the later one is dangerous because it will never fail when any problem exists in your code.
+They should only be used to speed up the initial adaption process.
+
+Best practices
+--------------
+
+Note that you can find a step-by-step tutorial after this section.
+You can also read the tutorial before reading the best practices.
+
+PREFER to expose your APIs using OCS
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It provides a more standardized and easier way to write your APIs.
+Other methods are considered legacy.
+For details take a look at :ref:`OCS <ocscontroller>`.
+
+.. collapse:: Examples
+
+    .. code-block:: php
+        :caption: Bad
+
+        class SomeController extends ApiController {
+            ...
+
+            public function someControllerMethod(): JSONResponse {
+                ...
+                return JSONResponse(...);
+            }
+        }
+
+    .. code-block:: php
+        :caption: Good
+
+        class SomeController extends OCSController {
+            ...
+
+            public function someControllerMethod(): DataResponse {
+                ...
+                return DataResponse(...);
+            }
+        }
+
+DO type controller and helper methods as explicit as possible
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The closer you narrow down a type without violating any constraints the better the resulting specification will be.
+Psalm will catch these problems for you if you configured the issue handlers mentioned above correctly.
+
+.. collapse:: Examples
+
+    .. code-block:: php
+        :caption: Bad
+
+        /**
+         * @return array
+         */
+        public function someHelperMethod() {
+            ...
+            return [
+                "id" => id,
+                "name" => name,
+            ];
+        }
+
+    .. code-block:: php
+        :caption: Good
+
+        /**
+         * @return array{id: int, name: string}
+         */
+        public function someHelperMethod() {
+            ...
+            return [
+                "id" => id,
+                "name" => name,
+            ];
+        }
+
+PREFER to use ``null`` to represent empty data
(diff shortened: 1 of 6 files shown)


developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,329 @@
+===================
+OCS OpenAPI support
+===================
+
+This page explains you how you can add OpenAPI support to your app such that you can automatically generated an OpenAPI specification for your server code.
```
DaphneMuller, line 5: generated > generate typo

DaphneMuller:
The writing and structure is crystal clear, thank you for your work Kate to share your knowledge. I'm looking forward to app developers trying out your tutorial to see what are the missing gaps if they want to apply this knowledge to their own apps. I like that you outlined the theory and best practises and gave a simple example to follow along. We now just need to know what is missing for people to apply it to their apps.

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,329 @@
...
+
+- Psalm has been setup for the app
+- The app supports at least Nextcloud 28
+- Your APIs are exposed via OCS (see :ref:`OCS <ocscontroller>`). It is possible to also expose APIs in different ways, but those will not be explained here because they are considered legacy.
+
+Best practices
+--------------
+
+- Be as explicit and concrete as possible about types. The closer you narrow down a type without violating any constraints the better the resulting specification will be.
+- For empty data you should always use `null` if possible. If your API is well designed it should always work. If for some reason you can not do that, you should use `new \stdClass()` and not `[]`. The problem with `[]` in PHP is that will be converted into `[]` in JSON, but you want to have `{}` in JSON if the data you return is empty. For empty arrays that should be arrays it is not a problem.
```
Leptopoda, line 25: I can see people not knowing the issue getting confused here.

The second part starting with `The problem with ...` has a lot of inline code where words might be better.
It's not the best but maybe something like:

> Keep in mind that JSON distinguishes between empty lists and empty objects so a PHP `[]` should only be used to indicate an array with no items.

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,329 @@
...
+Psalm will also help you validate your changes.
+
+Requirements and prerequisites
+------------------------------
+
+- Psalm has been setup for the app
+- The app supports at least Nextcloud 28
+- Your APIs are exposed via OCS (see :ref:`OCS <ocscontroller>`). It is possible to also expose APIs in different ways, but those will not be explained here because they are considered legacy.
+
+Best practices
```
Leptopoda, line 21: I don't know how easy it is to replicate here but I really like how pest practices are communicated by the dart language (https://dart.dev/effective-dart/documentation).

They all start with a key word like `DO`, `DON'T`, `PREFER`, `AVOID` `CONSIDER` and then also give a clear example why this is good or bad.

I think that the items should at least always start with a key word as it's already done with the `Do not`

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,329 @@
...
+    #[NoAdminRequired]
+    public function delete(int $id): DataResponse {
+        ...
+    }
+
+How to add response definitions to share type definitions
+---------------------------------------------------------
+
+In the previous steps we have been re-using the same data structure multiple times, but it was copied every time.
+This is tedious and error prone, therefore we want create some shared type definitions.
```
Leptopoda, line 218: Missing to
we want create some -> we want to create some

developer_manual/client_apis/OCS/ocs-openapi.rst
provokateurin: I like this idea lot. I will try to implement something similar

developer_manual/client_apis/OCS/ocs-openapi.rst
provokateurin: Is the new version better?

developer_manual/client_apis/OCS/ocs-openapi.rst
provokateurin: I think the best practices are a lot better now, especially with the code examples. What do you think?

Leptopoda (commented):
Looks very clear now.
Maybe you can also remove the duplications that now happened in some places the main information is duplicated.
this can be seen in  the `PREFER to expose your APIs using OCS`  case.

Leptopoda (approved):
:+1:

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,731 @@
...
+Setup the following issue handlers in the psalm config (see `<https://psalm.dev/docs/running_psalm/dealing_with_code_issues>`_ for a tutorial):
+
+.. code-block:: xml
+
+    <LessSpecificReturnStatement errorLevel="error"/>
+    <LessSpecificReturnType errorLevel="error"/>
+    <LessSpecificImplementedReturnType errorLevel="error"/>
+    <MoreSpecificReturnType errorLevel="error"/>
+
+Install openapi-extractor in your app as explained in `<https://github.com/nextcloud/openapi-extractor>`_. TODO add instructions there
```
provokateurin, line 38: Still missing installation instructions. See https://github.com/nextcloud/openapi-extractor/issues/7

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,731 @@
...
+            ];
+        }
+    }
+
+The capabilities will automatically appear in the generated specification.
+
+How to generate the specification
+---------------------------------
+
+TODO Run the `generate-spec` script from openapi-extractor inside the root folder of your app and if you did everything right you will have a new file called `openapi.json`.
```
provokateurin, line 729: Still not done, depends on the installation instructions.

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,731 @@
...
+Your API should be designed in a way that represents empty data with `null`.
+
+There is a problem with PHP and arrays that get converted to JSON.
+JSON has lists and objects while PHP only has arrays.
+If you were to return an empty array in PHP it will always turn into `[]` in JSON.
+This is not a problem for endpoints that always return lists, but most endpoints return a single JSON object.
+For those endpoints returning `[]` in PHP is a problem because the consumer will either get `[]` or `{...}` which is hard to handle.
+
+If you are not able to use `null` for whatever reason, use `new \stdClass()` instead.
+It will get correctly converted into `{}` in the JSON response.
```
nickvergessen, line 140: But it breaks the XML response, making your OCS not fully OCS compatible.
See https://github.com/nextcloud/spreed/blob/c4a0f78e478fdca0353d3c47a6fcaaa0a0b30711/lib/Model/Message.php#L162-L167 + https://github.com/nextcloud/spreed/blob/8225b4443b5678c1034757d6b5a989054e4377f4/lib/Controller/AEnvironmentAwareController.php#L64-L83 for a painful learned work around

provokateurin: It doesn't anymore since https://github.com/nextcloud/server/pull/38745

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -0,0 +1,731 @@
...
+
+Return valid responses with an error message instead.
+
+.. collapse:: Examples
+
+    .. code-block:: php
+        :caption: Bad
+
+        /**
+         * @throws OCSBadRequestException
```
nickvergessen, line 195: Those `OCS…Exception`s are OCP public API. If usage is discouraged, it should be marked deprecated with explanation and the "Good" be included in the deprecation.

developer_manual/client_apis/OCS/ocs-openapi.rst
nickvergessen: Should then mention 28+ as requirement and the change should make it into the dev-docs:
https://github.com/nextcloud/documentation/blob/master/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.rst#changed-apis-1

That being said, let's hope no one uses a non-empty stdClass:
https://www.php.net/manual/en/class.stdclass.php#stdclass.examples

provokateurin: 28+ requirement is mentioned at the beginning of the document

nickvergessen: Well there is a difference between "Supports at least 28+" and "Requires at least 28+"

Just add a simple comment:
```suggestion
If you are not able to use `null` for whatever reason, use `new \stdClass()` instead.
It will get correctly converted into `{}` in the JSON response on Nextcloud 28 and later.
```

provokateurin:
@nickvergessen I guess the minimum required version will be 27.1 now? I doubt that apps will target a minor version though, so better stay with 28?

developer_manual/client_apis/OCS/ocs-openapi.rst
provokateurin: This is only advice for writing good APIs. You can still use those exceptions and there are valid use cases (especially for keeping API compatibility).

bigcat88 (approved):

marcelklehr:
@provokateurin could you squash all those fixup commits?

marcelklehr (approved):

PR #10816, opened 2023-07-26
feat: more specific Apache recommendations

miaulalala:
Due to https://httpd.apache.org/security/vulnerabilities_24.html#CVE-2022-22720 recommend versions higher than 2.4.53.

| Before | After |
|--------|--------|
| image| image |


ChristophWurst (approved):

nickvergessen:
And due to https://httpd.apache.org/security/vulnerabilities_24.html#CVE-2023-25690 I recommend 2.4.56?
I think recommendation is to always use the latest version, we never track security issues of 3rdparty?

PR #10832, opened 2023-07-27
fix: apply transifex updates to pot files as well

max-nextcloud:
Transifex switched to using `.po` file extensions about a year ago.
Copy the `.po` files on top of the outdated `.pot` files.
This will use the latest translations that were available when stable27 was branched off.


max-nextcloud:
I'll clean up the unused `.po` files in a separate PR afterwards.

nickvergessen (approved):
🐘

PR #10835, opened 2023-07-27
cleanup: user_manual .po files

max-nextcloud:
They are duplicates of the .pot files.
Transifex used to create `.pot` files and then switched to `.po` about one year ago. We only need one version and now update the `.pot` files.


max-nextcloud:
Last PR to cleanup the transifex sync mess.

tflidd:
@rakekniven can you help or delegate to someone to check?

nickvergessen (approved):

PR #10841, opened 2023-07-27
Update Debian version in system_requirements.rst

tflidd:
If we recommend a Debian operating system, it would be better to use a version that already ships a php version that is supported by NC.

https://wiki.debian.org/PHP#Available_versions

Not sure if you do some specific tests on a Debian setup and to claim to support it, you would need to adjust the version as well.

Perhaps back-port to NC26 as well (first version to support php 8.2).


kesselb (approved):

kesselb:
> If we recommend a Debian operating system,

We don't, it's just listed under operating systems. 
However, the Linux distribution is not so relevant for us. The PHP version is.

PR #10817, opened 2023-07-26
Update middlewares.rst

fsamapoor:
The plural of middleware is middleware.

https://english.stackexchange.com/questions/257120/middleware-vs-middlewares

PR #10904, opened 2023-08-02
[stable-25] Update middlewares.rst

fsamapoor:
The plural of middleware is middleware.

Manual backport of #10817

### DIFF
--- a/developer_manual/basics/middlewares.rst
+++ b/developer_manual/basics/middlewares.rst
@@ -4,7 +4,7 @@ Middlewares
 
 .. sectionauthor:: Bernhard Posselt <dev@bernhard-posselt.com>
 
-Middleware is logic that is run before and after each request and is modelled after `Django's Middleware system <https://docs.djangoproject.com/en/dev/topics/http/middleware/>`_. It offers the following hooks:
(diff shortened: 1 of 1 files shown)


developer_manual/basics/middlewares.rst
```diff
@@ -4,7 +4,7 @@ Middlewares
 
 .. sectionauthor:: Bernhard Posselt <dev@bernhard-posselt.com>
 
-Middleware is logic that is run before and after each request and is modelled after `Django's Middleware system <https://docs.djangoproject.com/en/dev/topics/http/middleware/>`_. It offers the following hooks:
```
tflidd, line 7: That is only in US English: https://www.oxfordlearnersdictionaries.com/definition/english/model_2

fsamapoor: Good catch! My Grammarly is set to American English indeed.
So the docs should be based on British English?

ChristophWurst: Good question. I think that has never been defined explicitly.

fsamapoor: Maybe it can be added to the to-do list of the PR template?
Either way and regardless of personal preferences, I'm always in favor of consistency.

tflidd: Would be good to clear this up. For this backport it's probably too late and the same change is already in the master branch. So let's finish this here.

tflidd (approved):

PR #10984, opened 2023-08-11
fix(CI): Make warnings fail the CI build 

nickvergessen:
### ☑️ Resolves

* Fix #…

PR #10987, opened 2023-08-11
fix(developer): Generate icon docs again

nickvergessen:
### ☑️ Resolves

Together with this beauty: https://github.com/nextcloud/server/pull/39811 this helps to generate the icon docs again

### 🖼️ Screenshots

grafik

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ user-manual-html:
 	cd user_manual && make html
 	@echo "User manual build finished; HTML is updated"
 
-developer-manual-html: openapi-spec # icons-docs
+developer-manual-html: openapi-spec icons-docs
 	rm -rf developer_manual/_build/html/com
 	cd developer_manual && make html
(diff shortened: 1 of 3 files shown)


skjnldsv (changes requested):
They're deprecated, so I still think we should not have them.
Didn't they got removed on purpose?

nickvergessen:
No, the generating code broke with the change from scss to compiling with node sass.

We can also add a deprecation warning, fine by me. I'm just trying to get rid of the warning/error so we can enable https://github.com/nextcloud/documentation/pull/10984

The missing file is still referenced in https://github.com/nextcloud/documentation/blob/master/developer_manual/html_css_design/icons.rst?plain=1#L15

So either we remove the full block, or we fix the docs by generating the file again

skjnldsv:
I'm up for removal, especially with 28

nickvergessen:
Removal part of https://github.com/nextcloud/documentation/pull/10984

PR #11021, opened 2023-08-14
(general_troubleshooting) Add Tideways as a problematic PHP extensions

joshtrichards:
Tideways come up twice in the last couple weeks with the same impact: Files pane is completely blank in the Web UI and minimal to no indication of the cause.

It's also had some other [interactions with different PHP versions](https://support.tideways.com/documentation/setup/troubleshooting/debugging-segfaults-in-php.html) specifically involving Symfony.

This adds it to the (existing) list of potentially problematic PHP extensions in the Admin Manual for admins to check for if they're experiencing problems. 

### ☑️ Resolves

* Fix nextcloud/server#39535
* https://help.nextcloud.com/t/fyi-on-php-8-2-the-php-module-tideways-can-break-nextcloud/168198

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -230,6 +230,7 @@ these modules:
 
 3. PHP
 
+* Tideways
 * eAccelerator
 
 .. _trouble-webdav-label:


szaimen:
Shouldnt there rather be an admin check for it in the admin overview directly?

joshtrichards:
I like the idea of doing a check, but I don't know of any others we currently check for. An interesting enhancement for the future. We do, however, expose the list of active PHP extensions via *Admin->System*.

szaimen:
Here is a PR that could be used as inspiration: https://github.com/nextcloud/server/pull/31470/files

susnux (approved):
we should do both

PR #11023, opened 2023-08-15
Added single space

Reispflanze:
### ☑️ Resolves

* Fix #…


kesselb:
Thank you for sending a pull request  :+1: 

I'm closing your pull request because there should be no space.

> The password of the MySQL account used for connecting to the server. The password value is optional. If not given, [mysql](https://dev.mysql.com/doc/refman/8.0/en/mysql.html) prompts for one. If given, there must **be no space** between [--password=](https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_password) or -p and the password following it. If no password option is specified, the default is to send no password.[^1]

[^1]: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_password

PR #11032, opened 2023-08-16
Increase discoverability of background jobs documentation

Pytal:
Add section to "Installation on Linux"

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -296,6 +296,18 @@ To use ``occ`` see :doc:`command_line_installation`.
 
 To use the graphical Installation Wizard see :doc:`installation_wizard`.
 
+.. _background_jobs_label:
+
+Setting up background jobs
+--------------------------
+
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

admin_manual/installation/server_tuning.rst
```diff
@@ -2,12 +2,6 @@
 Server tuning
 =============
 
-Using cron to perform background jobs
```
nickvergessen, line 5: Keep it here and just link to it from more places?

Pytal: At this point it's not just tuning, I would say that it is a requirement to run time sensitive background jobs regularly but yeah we can keep it here too

The "Installation on Linux" page is likely the most visited page but I don't have the metrics, where else would be appropriate?

nickvergessen: I would just have it in both. It is still a tuning for existing installations that did not do it on setup

PR #11093, opened 2023-09-05
docs: add repair-filecache when mimetype has been updated

LouisSung:
### ☑️ Resolves
reflect the discussion at nextcloud/server#30566 that shows the need to repair the filecache after updating the mimetype
(e.g., to open files with a customized extension in text editor instead of downloading them)


kesselb (approved):
Make sense, thank you :+1:

PR #11120, opened 2023-09-17
rename Core to Server development

privatemaker:
### ☑️ Resolves

Currently in the `Developer Manual` on both the sidebar and page is the title:

- [Core development](https://docs.nextcloud.com/server/latest/developer_manual/core/index.html)

My understanding is this is an artifact from Owncloud when the `server` was called `core`

My pull request is a simple 6 character change which resolves this.

### 🖼️ Screenshots

nextcloud-docs-core-to-server

### DIFF
--- a/developer_manual/core/index.rst
+++ b/developer_manual/core/index.rst
@@ -1,8 +1,8 @@
 .. _coreindex:
 
-=================
-Core development
-=================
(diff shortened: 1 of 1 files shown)


susnux (approved):
:sparkle:

PR #11122, opened 2023-09-17
Clarify the use of a master key during setup of encryption

func0der:
Add notice about encryption having a master key setup, since it has been the default since Nextcloud 13 without any notice.

### ☑️ Resolves

Kinda resolves discussion in this issue: https://github.com/nextcloud/server/issues/8283


susnux (approved):

susnux:
Sorry for the very long delay! Somehow this slipped through our triaging process!
But thank you very much!

PR #11136, opened 2023-09-21
move dir developer_manual/core/  to developer_manual/server/

privatemaker:
In tandem with #11120 where we renamed the user facing word `Core` to `Server` this pull-request migrates the file in the directory `core/*.rst` to the new semantically correct path `server/*.rst` which causes the current URLs like:

```/server/latest/developer_manual/core/*.html```

Being redirected to the following new URLs:

```/server/latest/developer_manual/server/*.html```

I implemented [sphinx-reredirects](https://documatt.gitlab.io/sphinx-reredirects) so that any URLs "in the wild" still pointing to the `core/*` path will be bounced to appropriate `server/*` path. 

I tested this locally and they all redirect as expected. I added this to the dependencies.

---

Unrelated but 9d44cf7c0fd476fde68229cf290493029e19b424 fixes an inconsistency where the `build/server/` repo was previously cloned (at `--depth 1`) was downloading the whole many GB tree.
 
---

Also updated some `core` to `server` words with cf81672d84882d206fe59660310d2d8f9809b82f


susnux (approved):
✨

privatemaker:
@susnux thanx 😄

PR #11138, opened 2023-09-21
change Nextcloud to Interface & interaction design

privatemaker:
### ☑️ Resolves

The current wording `Nextcloud design guidelines` is:

- Redundant, as this is obviously about Nextcloud
- Vague as it's not obviously about visual UI / UX design 

documentation-nextcloud-design

This two word change addresses both issues. It also is easier to identify the section in search results.

documentation-interface-design
 
documentation-interface-design-search

### DIFF
--- a/developer_manual/design/index.rst
+++ b/developer_manual/design/index.rst
@@ -1,12 +1,12 @@
-===========================
-Nextcloud design guidelines
-===========================
+==============================
+Interface & interaction design
+==============================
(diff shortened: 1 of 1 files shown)


jancborchardt (approved):
Yep, this is more clear and searchable. Another option could be "Design system" which is a term getting more common use these days – albeit only in design circles tbh.

PR #11144, opened 2023-09-24
Mobile sync: Add note about 2-way sync support status in Android client

GTP95:
Added link where to find information about 2-way-sync on Android

### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots
Screenshot_20230924_220033


joshtrichards:
Hi @GTP95 - First, thanks for your contribution. :+1:  Let's hold on this for a moment:

1. I'm in the midst of cleaning up / consolidating old Issues in that repo. There may be a better one to link to, at a minimum. I'll try to look over the related ones in the next week or so. I'm also hesitant to link to a random issue for a key feature in an offhanded way from the User Manual. If there's something to be said in the Manual(s) about the 2-way sync functionality as it stands today I'd rather just say it (and update the manual as things evolve).
2. I'd like to get feedback from some of the active devs in that project. They have some better ideas about how to describe the current state so we can avoid linking to a random issue entirely. I've asked them to review this thread to get some feedback.

GTP95:
I linked to that specific issue because it's the oldest one about the topic, contains some links to other related issues and received recent attention, so I assumed it was a good place for a user to understand the current status of the issue. 
But I also understand your point about not linking to an issue entirely and just explain the situation directly in the manual. What I think it's important is that, one way or another, users are informed about the feature's status.

tobiasKaminsky (approved):

AndyScherzinger (approved):

PR #11154, opened 2023-09-28
fix indefinite articles

dartcafe:
### ☑️ Resolves

* Fix indefinite articles fom 'an' to 'a' in case of articles in front of URL and user
* Uppercase URL as  used in all other places


ChristophWurst (approved):

ChristophWurst:
Well, no fixes for stable26 and 27 :D

PR #11177, opened 2023-10-04
Update nginx-subdir.conf.sample &  nginx-root.conf.sample

noci2012:
prevent 405 error combined with pretty URL's

### ☑️ Resolves

* Issue with f.e. shifts (shiftplan)  that uses URL:
https://server/admin/settings     with PUT
location /   (without proxy_pass) returns error 405 on PUT/POST...

Solution a redirect in case of 405 errors.

Example patched.

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -186,6 +186,12 @@ server {
     location /remote {
         return 301 /remote.php$request_uri;
     }
+    # fix 405 error on posts that are passwd as pritty URL
+    error_page 405 =200 @post_static ;   #/index.php/$request_uri;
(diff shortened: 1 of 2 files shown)


noci2012:
Not sure this is the way to go now, it breaks some other things, some discussion appreciated.

kesselb:
Thanks :+1: 

I fail to understand your change. 

Did you discuss this change with the shifts developers?

noci2012:
never mind,.
There is a class of apps that doesn't work...
As there is a redirect to a directory. Shift plan is such an app.

noci2012:
This is the issue btw.
https://www.bonbon.io/articles/how-to-enable-post-to-static-pages-running-via-nginx

PR #11200, opened 2023-10-14
fix(l10n): Enumerations are displayed correctly

rakekniven:
Seen at docs.nextcloud.com while reviewing strings.

See https://docs.nextcloud.com/server/latest/user_manual/de/groupware/mail.html


kesselb (approved):

kesselb:
Thank you :+1:

PR #11209, opened 2023-10-14
fix(i18n): Fixed typo

rakekniven:
Simple typo and languages specific changes.


susnux:
Isn't the default language US English? At least for code it is (as British English is en-gb and defauöt US is en) so it should be "synchronization".

tflidd (approved):

PR #11221, opened 2023-10-17
Update settings.rst: Admin privileges

Rello:
Add info on Admin privileges

### DIFF
--- a/developer_manual/digging_deeper/settings.rst
+++ b/developer_manual/digging_deeper/settings.rst
@@ -294,3 +294,12 @@ This is rather straightforward as you can see in the code snippet below
     </settings>
     ...
 
+Administration privileges 
+-------------------------
+
+By default only members of the admin group can access and edit the admin settings. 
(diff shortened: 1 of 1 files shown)


joshtrichards:
Looks like we have some significant overlap between the *Basic concepts->Settings* and *Digging deeper->Settings* sections. And `IDelegatedSettings` is covered a bit in Basics already. 

- https://docs.nextcloud.com/server/latest/developer_manual/basics/setting.html
- https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/settings.html

I wonder if we should take this opportunity to see if there's really any need for the *Digging deeper* version for *Settings*? And if there is still, maybe reduce some of the overlap and/or have them reference each other to minimize redundancy?

kesselb:
I second Josh's suggestion.

blizzz:
Please don't forget to set milestone and labels when opening a PR :smiley:

PR #11231, opened 2023-10-19
reverse_proxy_configuration.rst: IPv6 CIDR is supported

jivanpal:
Confirmed that IPv6 CIDR notation works on server v27.0.0. Apparently, this has been supported since v25.0.0-beta1; see  https://github.com/nextcloud/server/pull/32615

### DIFF
--- a/admin_manual/configuration_server/reverse_proxy_configuration.rst
+++ b/admin_manual/configuration_server/reverse_proxy_configuration.rst
@@ -18,16 +18,14 @@ configured in :file:`config/config.php`
 
 Set the :file:`trusted_proxies` parameter as an array of:
 
-* IPv4 addresses, 
+* IPv4 addresses 
 * IPv4 ranges in CIDR notation
 * IPv6 addresses
+* IPv6 ranges in CIDR notation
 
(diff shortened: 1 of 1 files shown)


tflidd (approved):

joshtrichards (approved):
:+1:  And looks like the config sample got taken care of in nextcloud/server#35956.

jivanpal:
@joshtrichards Would we like to add the example addresses to match https://github.com/nextcloud/server/pull/35956 ?

tflidd:
If you want to add an example, I'd rather use a subnet size of at least /64. But if you add it, wouldn't you need to put ipv4 examples as well?

kesselb (approved):

kyrofa:
This is so timely, thank you @kesselb. I was just horrified to discover that ipv6 cidrs weren't supported, and started digging into why, only to learn that they are!  The v27 docs are just out of date. Should we backport this to other supported releases?

joshtrichards:
Done, @kyrofa. Thanks for the reminder!

kyrofa:
My pleasure, thank YOU!

PR #11254, opened 2023-11-01
Fix(ldap): Config is not active by default

### LINKED ISSUE #912: LDAP "Configuration Active" should have more emphasys
Hi,

I believe there should be more emphasis about the need to have "Configuration Active" checked in order to have LDAP working.

I spent one day troubleshooting a new installation (v 14) because all LDAP configuration was correct and still got "wrong password" when trying to authenticate an LDAP account.

The current information is buried in the LDAP section of Admin Manual. I believe, right in the beginning of the section, there should be a warning.

My suggestion is to replace 

"
Configuration
First enable the LDAP user and group backend app on the Apps page in Nextcloud. Then go to your Admin page to configure it.
"

by

"
Configuration
First enable the LDAP user and group backend app on the Apps page in Nextcloud. Then go to your Admin page to configure it. **Before testing, make sure that Advanced / Configuration Active is checked.**
"

I'll also open two issues to Nextcloud Server:

1. The "Configuration Active" item wasn't checked, despite the configuration was tested successfully multiple times

2. The error message shouldn't be "wrong password", but "no active LDAP configuration found".

Regards,
Roger

c199

joshtrichards:
### ☑️ Resolves

* Fix #912 

### 🖼️ Screenshots

Before:

image

After:

image

### DIFF
--- a/admin_manual/configuration_user/user_auth_ldap.rst
+++ b/admin_manual/configuration_user/user_auth_ldap.rst
@@ -36,6 +36,9 @@ Configuration
 First enable the ``LDAP user and group backend`` app on the Apps page in
 Nextcloud. Then go to your Admin page to configure it.
 
+.. warning:: By default, a new LDAP configuration is inactive. See :ref:`advanced_settings_label`
+   to make your configuration active or test it. Until you do so, LDAP authentication will fail.
+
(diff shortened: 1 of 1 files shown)


blizzz:
Actually, the configuration is turned on automatically, when all required settings are done and tests succeed.

PR #11265, opened 2023-11-05
Remove old redis installation hints

tflidd:
Follow up on https://github.com/nextcloud/documentation/pull/11264.

There is more outdated documentation regarding redis:
- instruction to retired Ubuntu 14.04 manual
-  php-redis packages are now included in Ubuntu (since 20.04 LTS), older onces are or are about to expire by now (18.04 + 5 years)
- info about redis suppot for php7.1/2 compatibility is not relevant, these versions are not supported for quite some time.

Not sure about the fedora/CentOS part, but there is just an information how to retrieve package information on a setup, that does not really help.

We can probably backport these changes to all supported versions.

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/configuration_server/caching_configuration.rst
+++ b/admin_manual/configuration_server/caching_configuration.rst
@@ -307,31 +307,6 @@ More information on configuration of phpredis session handler can be found on th
 
 ..  _install_redis_label:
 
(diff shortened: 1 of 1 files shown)


kesselb (approved):
Thank you

PR #11266, opened 2023-11-05
Database installation - MariaDB use mariadb / no FLUSH PRIVILEGES

grooverdan:
MariaDB is deprecating off its mysql client usage (has been a symlink for a very long time).

FLUSH PRIVILEGES aren't required after any creation of users or grants. It has only two very specific meenings that aren't the case here; when credentials are directly changes in underlying tables using DML, or the server was started with --skip-grant-tables and this is ending that lack of priv scheme.

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/configuration_database/linux_database_configuration.rst
+++ b/admin_manual/configuration_database/linux_database_configuration.rst
@@ -127,6 +127,10 @@ To start the MySQL command line mode use::
 
   mysql -uroot -p
 
+When using MariaDB use::
+
+  mariadb -uroot -p
+
 Then a **mysql>** or **MariaDB [root]>** prompt will appear. Now enter the following lines and confirm them with the enter key:
 
 ::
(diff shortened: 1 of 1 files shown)


tflidd:
The same command is used for mysql setups, does this apply for mysql as well?

grooverdan:
yes, applies to mysql too.

per docs say don't say FLUSH PRIVILEGES is required:

https://dev.mysql.com/doc/refman/8.0/en/grant.html
https://dev.mysql.com/doc/refman/8.0/en/create-user.html
https://dev.mysql.com/doc/refman/8.0/en/create-database.html

As a second check, creating those object makes them immediately available without it.

nickvergessen:
We also don't flush in the code when we create the user:
https://github.com/nextcloud/server/blob/master/lib/private/Setup/MySQL.php#L123

tflidd (approved):

PR #11282, opened 2023-11-13
Document new API in OCP\Authentication

come-nc:
### ☑️ Resolves

* Document https://github.com/nextcloud/server/pull/41017

### 🖼️ Screenshots

Screenshot_20231113_151951

### DIFF
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.rst
@@ -144,6 +144,8 @@ Added APIs
 
   * ``wasm-unsafe-eval`` is `supported by most browsers <https://caniuse.com/mdn-http_headers_content-security-policy_script-src_wasm-unsafe-eval>`_
   * WebAssembly compilation and execution in worker threads is not affected by this directive (browsers allow compilation and execution of WebAssembly in worker threads by default)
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

ChristophWurst:
New APIs should have a persistent place, e.g. somewhere we already have docs about authentication.

The upgrade docs purpose is to help maintains of existing apps keep their app alive for the next major. They only need to know about stuff that was changed or removed.

In other words, someone potentially developing against this new API will have a hard time finding docs.

Hope this makes sense.

come-nc:
> New APIs should have a persistent place, e.g. somewhere we already have docs about authentication.
> 
> The upgrade docs purpose is to help maintains of existing apps keep their app alive for the next major. They only need to know about stuff that was changed or removed.
> 
> In other words, someone potentially developing against this new API will have a hard time finding docs.
> 
> Hope this makes sense.

In https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/security.html ?

Most of OCP is not documented outside of https://nextcloud-server.netlify.app, which already lists https://nextcloud-server.netlify.app/classes/ocp-authentication-token-iprovider and https://nextcloud-server.netlify.app/classes/ocp-authentication-token-itoken

ChristophWurst:
> In https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/security.html ?

Sounds good!

The current state is not great, yeah. We have to start somewhere …

PR #11303, opened 2023-11-22
Updated language_configuration.rst to proper URL

Krelyshy:
Transifex URL isn't working anymore, I found out that NC uses ISO-639-1 lang. codes

https://docs.nextcloud.com/server/19/admin_manual/configuration_server/config_sample_php_parameters.html

### ☑️ Resolves

* invalid URL

### DIFF
--- a/admin_manual/configuration_server/language_configuration.rst
+++ b/admin_manual/configuration_server/language_configuration.rst
@@ -33,8 +33,8 @@ in the :file:`config/config.php`.
 If users shall be unable to change their language, but users have different languages,
 this value can be set to ``true`` instead of a language code.
 
-.. note:: Please check `Transifex language codes
-   <https://explore.transifex.com/languages/>`_ for the list of valid language
+.. note:: Please check `ISO 639-1 language codes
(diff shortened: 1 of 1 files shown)


tflidd:
For me, the transifex url still works ...

Krelyshy:
> For me, the transifex url still works ...

Today I checked again, and all of the sudden, it somehow works again. It is unreliable though, since it did not work for about a month, so I would change the link to wikipedia, though I do not have any proof of the site not working... a few weeks ago, it pointed to transifex's homepage.

tflidd:
Some are different, the `@` notation is only in transifex:

>     Uyghur (ug)
>     Uighur (Arabic) (ug@Arab)
>     Uighur (Cyrillic) (ug@Cyrl)
>     Uighur (Latin) (ug@Latin)

Also you have different Englishs:

>     English (Chile) (en_CL)
>     English (Colombia) (en_CO)
>     English (Cyprus) (en_CY)
>     English (Czech Republic) (en_CZ)

not sure if we can use them all like this?

PR #11317, opened 2023-11-27
Add OpenAPI workflow to confirm it's working

nickvergessen:
### ☑️ Resolves

* CI for #11314

### DIFF
--- a/.github/workflows/openapi.yml
+++ b/.github/workflows/openapi.yml
@@ -1,0 +1,34 @@
+name: OpenAPI
+
+on: pull_request
+
+permissions:
+  contents: read
+
+concurrency:
+  group: openapi-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  openapi:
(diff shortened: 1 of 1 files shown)


nickvergessen:
Failing run on stable28 before the merge of #11315 
https://github.com/nextcloud/documentation/actions/runs/7005702281/job/19055942324

provokateurin:
This doesn't make much sense because it breaks silently and not due to PR changes. The only real solution is https://github.com/nextcloud/documentation/issues/11318

nickvergessen:
> This doesn't make much sense because it breaks silently and not due to PR changes.

It would still notify us that the next release build will be broken.
But yeah #11318 is much better.

provokateurin:
> It would still notify us that the next release build will be broken.

In that case running it as a cron would make more sense since none of the PRs will be the cause of the breakage and it would unnecessarly block merging fine PRs.

provokateurin (approved):

PR #11305, opened 2023-11-23
Add whitelist configuration on bruteforce_configuration.rst

quentinDupont:
### ☑️ Resolves

* Clarify settings for Whitelist
* Help admin that don't have the right app enabled

PR #11347, opened 2023-12-02
Add whitelist configuration on bruteforce_configuration.rst

kesselb:
### ☑️ Resolves

* Rebased version of https://github.com/nextcloud/documentation/pull/11305

### 🖼️ Screenshots

image


Altahrim (approved):

admin_manual/configuration_server/bruteforce_configuration.rst
```diff
@@ -22,6 +22,16 @@ The maximum delay is 25 seconds.
...
+Exclude IP addresses from brute force protection
+------------------------------------------------
+
+It's possible to exlude IP addresses from the brute force protection.
+
+- Enable the bruteforcesettings app
+- Login as admin and go to Administration settings -> Security
+
+Note that any excluded IP address can perform authentication attempts without any throttling.
+Its best to exclude as few IP addresses as you can, or even none at all.
```
nickvergessen, line 34: Are issues with proxies mentioned on this page and that instead of excluding a (reverse) proxy it should be configured as trusted proxy?

kesselb: Yep, it's mentioned in the troubleshooting section below: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/bruteforce_configuration.html#troubleshooting

Shall we flip troubleshooting and exclude?

nickvergessen: yeah I would say so

nickvergessen: don't know the exact syntax atm, but maybe we make this a warning box as well

```suggestion
.. warn::
    Its best to exclude as few IP addresses as you can, or even none at all. 
```

joshtrichards:
Thanks @quentinDupont and @kesselb for taking the initiative on this! :sunglasses:  Comes up often enough. :)

nickvergessen:
>  Comes up often enough. :)

And should basically never be the solution. In most cases you are doing something else wrong, see the warning box.

PR #11374, opened 2023-12-09
Correct the version of owncloud allowing migration

SystemKeeper:
### ☑️ Resolves

Upgrading from owncloud is only supported from 10.5, and not from 10.0.5 as stated in the documentation. See https://github.com/nextcloud/server/blob/stable20/version.php#L43 for reference.

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/maintenance/migrating_owncloud.rst
+++ b/admin_manual/maintenance/migrating_owncloud.rst
@@ -17,7 +17,7 @@ See the table below for a version map, where migrating is easily possible:
 +===================+==============================+
 | 10.13.x           | 25.0.x (but at least 25.0.2) |
 +-------------------+------------------------------+
-| 10.0.5 or later   | 20.0.x (but at least 20.0.5) |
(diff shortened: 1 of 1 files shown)


nickvergessen:
Hmm looking at https://github.com/nextcloud/documentation/commit/6034de3149fb1ac74d9eb22bbb5da4a29a6cb35f it seems it was intended or even more wrong

SystemKeeper:
Maybe it worked before, but after https://github.com/nextcloud/server/pull/23044 it seems that upgrade from below 10.5 is not possible anymore?
 
https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/lib/private/Updater.php#L208-L215
https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/lib/private/Updater.php#L223-L225

nickvergessen (approved):

PR #11406, opened 2023-12-19
Update config_sample_php_parameters.rst

eric-schleicher:
update to docs for config.php as the system does not check  `simpleSignUpLink.shown`, rather it checks `simpleSignUpLink` (with no sub property) to control this functionality.

### ☑️ Resolves

* unexplained confusion from forums because the docs are just out of sync with the current version.

### 🖼️ Screenshots

N/A

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -3302,13 +3302,13 @@ specifications. For those, have an opt-out.
 
 WARNING: only use this if you know what you are doing
 
-simpleSignUpLink.shown
+simpleSignUpLink
 ^^^^^^^^^^^^^^^^^^^^^^
 
 
 ::
 
-	'simpleSignUpLink.shown' => true,
+	'simpleSignUpLink' => true,
 
 By default, there is on public pages a link shown that allows users to
 learn about the "simple sign up" - see https://nextcloud.com/signup/


kesselb:
Hi and thanks for your pull request :+1: 

This page is generated from https://github.com/nextcloud/server/blob/6d557569f6106f00867c970eeae2d8daa960e8be/config/config.sample.php#L2277C2-L2277C24 and therefore you need to change the sample file in nextcloud/server.

However `simpleSignUpLink.shown` appears to be the right value.

eric-schleicher:
But alas `simpleSignUpLink.shown` is not used by the application and
doesn’t work.


I’ll go chase the right repo to resubmit, but making sure it’s clear that
the value in the sample config and the docs are ***incorrect*** (now
checking commit history)

---------------------------------
CONFIDENTIALITY NOTICE: The contents of this email message and any
attachments are intended solely for the addressee(s) and may contain
confidential and/or privileged information and may be legally protected
from disclosure. If you are not the intended recipient of this message or
their agent, or if this message has been addressed to you in error, please
immediately alert the sender by reply email and then delete this message
and any attachments. If you are not the intended recipient, you are hereby
notified that any use, dissemination, copying, or storage of this message
or its attachments is strictly prohibited.

kesselb:
> But alas `simpleSignUpLink.shown` is not used by the application and
doesn’t work.

https://github.com/nextcloud/server/blob/6d557569f6106f00867c970eeae2d8daa960e8be/lib/private/TemplateLayout.php#L204

PR #11428, opened 2024-01-08
fix(admin_manuel): Add note about shared items feature in contacts

JohannesGGE:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots
image

### DIFF
--- a/admin_manual/groupware/contacts.rst
+++ b/admin_manual/groupware/contacts.rst
@@ -42,3 +42,10 @@ Address Book Sync
 ^^^^^^^^^^^^^^^^^
 
 The address book is updated automatically with every added, modified, disabled or removed user. Admins can also trigger a full rewrite of the address book :ref:`with occ<occ-dav-sync-system-address-book>`.
+
+Shared items
+------------
+
+.. versionadded:: 5.5.0
+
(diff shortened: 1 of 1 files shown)


admin_manual/groupware/contacts.rst
```diff
@@ -42,3 +42,10 @@ Address Book Sync
 ^^^^^^^^^^^^^^^^^
 
 The address book is updated automatically with every added, modified, disabled or removed user. Admins can also trigger a full rewrite of the address book :ref:`with occ<occ-dav-sync-system-address-book>`.
+
+Shared items
+------------
+
+.. versionadded:: 5.5.0
+
+.. note:: `The related resources app <https://apps.nextcloud.com/apps/related_resources>`_ should be already in place to enable this feature. By default the app is shipped with Nextcloud.
```
ChristophWurst, line 51: Let's rephrase saying that the app is shipped but has to be *enabled* for the feature to be available

This new section has only a note element, no actual contents. Let's render the text without a note box.

JohannesGGE: :heavy_check_mark:

ChristophWurst (approved):

PR #11442, opened 2024-01-15
feat(developer_manual): Add Route attribute to routing introduction

provokateurin:
### ☑️ Resolves

Documents the new routing method available in Nextcloud 29.

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/basics/routing.rst
+++ b/developer_manual/basics/routing.rst
@@ -15,11 +15,22 @@ Routes map a URL and a method to a controller method. Routes are defined inside 
         ],
     ];
 
+.. versionadded:: 29
+
+You can also use attributes on the Controller method to define routes.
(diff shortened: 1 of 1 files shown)


developer_manual/basics/routing.rst
```diff
@@ -15,11 +15,20 @@ Routes map a URL and a method to a controller method. Routes are defined inside
         ],
     ];
 
+Since Nextcloud 29 you can also use an attributes on the Controller method to define routes.
```
ChristophWurst, line 18:
```suggestion
.. versionadded:: 29

You can also use an attributes on the Controller method to define routes.
```

works best if it has a dedicated heading/section

ChristophWurst (approved):

PR #11444, opened 2024-01-15
Remove bz2 from recommend extensions

J0WI:
### ☑️ Resolves

* pragmatic fix for [#42342](https://github.com/nextcloud/server/issues/42342)

### DIFF
--- a/admin_manual/installation/php_configuration.rst
+++ b/admin_manual/installation/php_configuration.rst
@@ -48,7 +48,6 @@ Database connectors (pick the one for your database):
 
 Required for specific apps:
 
-* PHP module bz2 (for apps packaged as bz2 archive)
 * PHP module ldap (for LDAP integration)
 * PHP module smbclient  (SMB/CIFS integration, see
   :doc:`../configuration_files/external_storage/smb`)
--- a/admin_manual/installation/php_configuration.rst
+++ b/admin_manual/installation/php_configuration.rst
@@ -42,13 +42,13 @@ Database connectors (pick the one for your database):
 
 *Recommended* packages:
 
-* PHP module bz2 (recommended, required for extraction of apps)
 * PHP module intl (increases language translation performance and fixes sorting
   of non-ASCII characters)
 * PHP module sodium (for Argon2 for password hashing. bcrypt is used as fallback, but if passwords were hashed with Argon2 already and the module is missing, your users can't log in.)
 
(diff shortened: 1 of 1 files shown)


Altahrim (approved):

admin_manual/installation/php_configuration.rst
```diff
@@ -42,13 +42,13 @@ Database connectors (pick the one for your database):
...
 *Recommended* packages:
 
-* PHP module bz2 (recommended, required for extraction of apps)
 * PHP module intl (increases language translation performance and fixes sorting
   of non-ASCII characters)
 * PHP module sodium (for Argon2 for password hashing. bcrypt is used as fallback, but if passwords were hashed with Argon2 already and the module is missing, your users can't log in.)
 
 Required for specific apps:
 
+* PHP module bz2 (for apps packaged as bz2 archive)
```
joshtrichards, line 51:
```suggestion
```

joshtrichards: I think we can safely remove it outright from the docs since { core + no shipped apps need it, no currently published apps for >=NC26 are packaged as bz2}, but I have no problem leaving it as you are suggesting as an interim path. :shrug: 

Your approach is consistent with how we're treating the checks for the other modules in that section (i.e. we don't check in the core setup checks and leave it to individual apps that need one of these to check for it themselves)

joshtrichards (approved):

come-nc:
The problem here is that from what I understood bz2 is documented to be a supported format for applications.
So we should deprecate/forbid it before removing it from recommended modules.

Does the appstore really support bz2? If not that means we can safely remove it without deprecation period I think.
If it does support it, can we get statistics on whether it’s actually used?

nickvergessen:
https://github.com/nextcloud/appstore/blob/09e2de7092eaac5b4d37d1cfee3c0d2ca880d6e5/nextcloudappstore/api/v1/tests/test_parser.py#L223-L239

https://github.com/search?q=repo%3Anextcloud%2Fappstore%20bz2&type=code

Seems like it never did support it?

come-nc (approved):
Looking into it I think the appstore does not support bz2 and it is not used, so let’s go with this.

bigcat88:
Currently we have no applications in the AppStore in `bz2` format and **afaik** it is not possible currently to upload in such format.

PR #11470, opened 2024-01-23
Update backup.rst

rickysixx:
Mention `custom_apps` directory.

### ☑️ Resolves

Custom apps too should be retained when fully backing up a Nextcloud instance.

### DIFF
--- a/admin_manual/maintenance/backup.rst
+++ b/admin_manual/maintenance/backup.rst
@@ -2,9 +2,10 @@
 Backup
 ======
 
-To backup a Nextcloud installation there are four main things you need to retain:
+To backup a Nextcloud installation there are five main things you need to retain:
 
 #. The config folder
+#. The custom apps folder (only if you have custom apps installed)
 #. The data folder
 #. The theme folder
 #. The database


joshtrichards:
Hi @rickysixx - First, thank you for the contribution!

Is there a specific situation you ran into that lead you to making this change? 

In theory, apps in alternative `apps_paths` should be handled the same way as any other app.

rickysixx:
Hi @joshtrichards!

Yes. In these days I have been experimenting different strategies to backup my Nextcloud instance.

First I performed a backup following the steps in that documentation page, but when I restored that backup I noticed that there were no apps in my restored Nextcloud instance (which was strange, because in my original instance I use Calendar and Task apps for example) even though I have not changed the default app installation path.

So I digged and I have found that the `/var/www/htnl/custom_apps` directory should be backed up too, if there are apps installed.

I thought about making this pull request because I felt the documentation was missing this little thing :)

skjnldsv (approved):

skjnldsv:
I'm ok with that, I think it make sense if you want to be able to restore specific versions without internet connection. Especially helpful for offline backups etc etc.

PR #11473, opened 2024-01-27
fix(admin): Update supported database support

susnux:
### ☑️ Resolves

* Resolves https://github.com/nextcloud-gmbh/server/issues/721
  * We need support for mariadb 10.11 as this is the only version provided by our supported Debian 12 version
  * mariadb 10.4 is not used by any supported or used distribution and EOL with Nextcloud 29
  * mariadb 10.3 is EOL but we need to keep support as long as we support Ubuntu 20.04
  * Postgres 10 and 11 are EOL and not used by any distribution we support or is used, all distributions also provide at least Postgres 12

BTW mariadb 10.11 is the current LTS version.

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -23,9 +23,9 @@ For best performance, stability and functionality we have documented some recomm
 |                  | - openSUSE Leap 15.5                                                  |
 |                  | - CentOS Stream                                                       |
 +------------------+-----------------------------------------------------------------------+
-| Database         | - **MySQL 8.0+** or MariaDB 10.3/10.4/10.5/**10.6** (recommended)     |
+| Database         | - **MySQL 8.0+** or MariaDB 10.3/10.5/**10.6** (recommended)/10.11    |
 |                  | - Oracle Database 11g (*only as part of an enterprise subscription*)  |
-|                  | - PostgreSQL 10/11/12/13/14/15                                        |
+|                  | - PostgreSQL 12/13/14/15/16                                           |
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

AndyScherzinger:
Fine by me, also looping in @sorbaugh @nickvergessen @ChristophWurst @juliushaertl FYI.

At some point we might want/need to restructure is since some versions/support are linked between different requirements, like OS and DB versions. Like mentioned by @susnux 10.4 is too high for Ubuntu 20 but 22 would your versions larger than 10.5 yet it is likely confusing people if we support 10.3 and 10.5 but not 10.4 - so maybe leave it in to not confuse people because an explanation might just be as confusing?

susnux:
Well but we then also support 10.6 an 10.11 but not 10.7/8/9/10 due to the fact they are EOL, right?

Or we might add a note like "only supported on Ubuntu 20.04" or similar? Not sure if we want to guarantee working solution for EOL like 10.4. So I am fine with every decision here (just would like to have 10.11 officially supported :wink: )

AndyScherzinger:
Well 10.11 we should add of course (if it has been tested, best if run in CI in our matrix). For the older versions, I am fine with removing or keeping the versions. My main issue is to align on a way that makes this clear to the readers :)

szaimen:
merge then or not merge?

nickvergessen (approved):

szaimen:
> /backport to stable28

But we dont test postgres16 and co. on stable28 currently?

nickvergessen:
yeah, right. But the removals and 10.11 still should go there

PR #11492, opened 2024-02-01
feat(devmanual): DB clusters and read/write split

ChristophWurst:
### ☑️ Resolves

* For https://github.com/nextcloud/server/pull/42345

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/digging_deeper/performance.rst
+++ b/developer_manual/digging_deeper/performance.rst
@@ -60,6 +60,52 @@ If you increase the long_query_time to 100 and add log-queries-not-using-indexes
   slow_query_log_file = /var/log/mysql/mysql-slow.log
   long_query_time=100
 
+Writing scalable transactions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Database queries and transactions have to work efficiently no matter the size of the Nextcloud installations. They have to work with simple SQLite, a single node Postgres and a MariaDB Galera cluster to give some examples.
+
+Database clusters
+~~~~~~~~~~~~~~~~~
+
+Single instance database offer strong consistency. Clusters can offer it too, but it comes with a performance penalty and admins occasionally choose to relax these guarantees to increase throughput. One of these allowed inconsistencies can happen on read/write split clusters, where some nodes handle only read (SELECT) operations while other can handle read and write (SELECT, INSERT, UPDATE and DELETE). Keeping replica nodes in sync with primary nodes is expensive. Allowing a few milliseconds of delay for the data to propagate from primary to replica allows the primary to process more queries.
+
+.. versionadded:: 29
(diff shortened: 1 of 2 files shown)


developer_manual/digging_deeper/performance.rst
```diff
@@ -60,6 +60,50 @@ If you increase the long_query_time to 100 and add log-queries-not-using-indexes
...
+
+There are two patterns to avoid the "dirty" read:
+
+  1. **Wrap the write+read operation in a transaction**. Nextcloud's read/write split, but also other database cluster load balancers will ensure that the queries of a transactions go to one single database node of a cluster. That ensures that data written is instantly available to be read back. This approach guarantees consistency, but puts additional load on the primary node because it has to execute the read operation.
+  2. **Avoid the read operation**. If the code allows it, avoid the read operation all together. You should know what was just written. If you need the auto increment ID, use the database's *last insert ID* feature. Proceed with this data, pass it to event listeners, etc. This approach guarantees consistency, too, but also improves overall performance.
+
+.. note::
+  Nextcloud can help you identify read after write without the need to set up a cluster for your development environment. If you change the loglevel to 0 (debug), dirty reads will trigger a log entry. Monitor the log when testing your code.
```
juliushaertl, line 91: Maybe add a comment that the log is based on the table so there might be e false positives

developer_manual/digging_deeper/performance.rst
```diff
@@ -60,6 +60,50 @@ If you increase the long_query_time to 100 and add log-queries-not-using-indexes
...
+It is important for Nextcloud developers to keep this in mind when writing database queries, especially when it is a series of queries. The next sections cover common anti-patterns and solutions.
+
+Reading data that has just been written
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A common pattern that works fine with small databases but falls apart on overloaded clusters are causal reads. These happen when a Nextcloud process INSERTs new data and reads that data right away. This can be obvious to spot in the code, but sometimes this is also obfuscated because of event listeners that react on new data.
+
+There are two patterns to avoid the "dirty" read:
+
+  1. **Wrap the write+read operation in a transaction**. Nextcloud's read/write split, but also other database cluster load balancers will ensure that the queries of a transactions go to one single database node of a cluster. That ensures that data written is instantly available to be read back. This approach guarantees consistency, but puts additional load on the primary node because it has to execute the read operation.
```
juliushaertl, line 87: Should we mention that depending on where the queries happen this might not be an option, as usually you'd want to keep the transaction time as low as possible?

juliushaertl (approved):
Nice writeup 💪

Two small comments inline, but looks good in general

juliushaertl (approved):

PR #11508, opened 2024-02-07
fix(dev): Remove duplicate debugging page

provokateurin:
### ☑️ Resolves

https://github.com/nextcloud/documentation/pull/2165 failed to remove the original page when moving it. The correct page is at https://github.com/nextcloud/documentation/blob/master/developer_manual/digging_deeper/debugging.rst.

@ChristophWurst how far should this be backported if at all?


ChristophWurst (approved):
Thanks!

I'd backport down to 26

PR #11522, opened 2024-02-08
fix(developer): Mention removal of global `autosize`

susnux:
For: https://github.com/nextcloud/server/pull/43466


ChristophWurst (approved):
nit: it is a removed global, not an API

https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_20.html#removed-globals

PR #11536, opened 2024-02-12
Speech-To-Text and Machine Translation providers with user context

andrey18106:
Describe provider with userId usage for Machine Translation (https://github.com/nextcloud/server/pull/42649) and Speech-To-Text (https://github.com/nextcloud/server/pull/42761), similar to Text Processing for consistency between them.

### Speech-To-Text

image

### Machine Translation
image

### DIFF
--- a/developer_manual/digging_deeper/speech-to-text.rst
+++ b/developer_manual/digging_deeper/speech-to-text.rst
@@ -109,6 +109,51 @@ The method ``transcribeFile`` transcribes the passed file and returns the transc
 
 The class would typically be saved into a file in ``lib/SpeechToText`` of your app but you are free to put it elsewhere as long as it's loadable by Nextcloud's :ref:`dependency injection container<dependency-injection>`.
 
+Provider with user context
+--------------------------
+
+.. versionadded:: 29.0.0
+
+Sometimes the processing of a the task may depend upon which user requested the task.
+You can now obtain this information in your provider by additionally implementing the ``OCP\SpeechToText\ISpeechToTextProviderWithUserId`` interface:
+
+.. code-block:: php
+    :emphasize-lines: 9,12,14,25,26,27
+
+    <?php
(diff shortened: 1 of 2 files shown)


nickvergessen:
Please also add to https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.html#id1

ChristophWurst:
> Please also add to https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.html#id1

IMO upgrade docs are not necessary for a *new* API. If you have an existing app you don't need to do anything about this to keep the app alive.

nickvergessen:
> IMO upgrade docs are not necessary for a new API. If you have an existing app you don't need to do anything about this to keep the app alive.

My understanding of that thing was to have a "quick overview" of things you could add to your app, without having to grep the docs repo for "New in version 29" or checking all files touch-date.
Maybe you considered implementing a STT provider but the missing context prevented you from doing so?

ChristophWurst:
Fair point

My inspiration was https://laravel.com/docs/9.x/upgrade. If we list everything, the important changes can be hard to find. We could have a dedicated section for announcements of new APIs?

bigcat88:
_Can we not add this time to the “new APIs”, we are not sure how long these classes will be there, this will be decided by team and Marcel, perhaps in the near future we will make a system where providers are immediately located in exApp and will move away from using these classes through the process of deprication._

nickvergessen:
> Can we not add this time to the “new APIs”, we are not sure how long these classes will be there, this will be decided by team and Marcel, perhaps in the near future we will make a system where providers are immediately located in exApp and will move away from using these classes through the process of deprication.

It's OCP, so our deprecation period is 3 years?

PR #11538, opened 2024-02-13
enh(utility): add timezone generator method to ITimeFactory

miaulalala:
Documentation for https://github.com/nextcloud/server/pull/41300

### 🖼️ Screenshots
image


ChristophWurst (changes requested):
Timezone is not part of https://www.php-fig.org/psr/psr-20/#21-clockinterface

Not sure where, but this should go somewhere else

ChristophWurst:
I would put it somewhere under https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/index.html. Upgrade docs are only there to highlight *changes*

ChristophWurst (approved):

PR #11585, opened 2024-02-26
feat: Add hint about renaming of circles to teams

juliushaertl:
Add hint about renaming of circles to teams to the developer upgrade guide for 29

### DIFF
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.rst
@@ -5,6 +5,8 @@ Upgrade to Nextcloud 29
 General
 -------
 
+* The Circles app will be renamed to Teams. Apps that use the terms Circle/Circles need to get adjusted to use Team/Teams instead. For example ``share to circle`` would become ``share to team``.
+
 info.xml
 ^^^^^^^^
 


ChristophWurst (commented):
Should go into general because there could also be strings in the backend that need adjustment

juliushaertl:
We would not rename APIs and internal references for now, that's why I moved it to frontend, but of course if backend provides string to the user then those also need adjustments, but I can move it.

developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.rst
```diff
@@ -19,6 +19,8 @@ Make sure your ``appinfo/info.xml`` allows for Nextcloud 29.
 Front-end changes
 -----------------
 
+* The Circles app will be renamed to Teams, meaning that any usage of the terms Circle/Circles need to get adjusted to Team/Teams for apps that make use of it, e.g. when sharing to a circle
```
max-nextcloud, line 22: Trying to break the long sentence into multiple: 
```suggestion
* The Circles app will be renamed to Teams. Apps that use the terms Circle/Circles need to get adjusted to use Team/Teams instead. For example ``share to circle`` would become ``share to team``.
```

max-nextcloud (approved):
Thanks for taking care of this. Content is fine. I'll propose a different wording but good to go from my side either way.

ChristophWurst:
> We would not rename APIs and internal references for now, that's why I moved it to frontend, but of course if backend provides string to the user then those also need adjustments, but I can move it.

I mean things like l10n strings in activities, notifications, php templates for settings

juliushaertl:
Moved and applied the wording suggestion.

PR #11587, opened 2024-02-26
enh(admin_manual): document count discrepancy for

miaulalala:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

| Before | After |
|--------|--------|
| image | image |


admin_manual/configuration_server/occ_command.rst
```diff
@@ -1434,6 +1434,10 @@ authentication servers such as LDAP::
  | user directories | 2  |
  +------------------+----+
```
come-nc, line 1435: Could you also update the example output? It does not contain active&disabled lines which seems to always be there.

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1434,6 +1434,10 @@ authentication servers such as LDAP::
...
 
+ There might be a discrepancy between the total number of users compared to the number of active users and the number of disabled users. 
+ Users that have never logged in before are not counted as active or disabled users. 
+ Some user backends also do not allow a count for the number of users.
```
come-nc, line 1439:
```suggestion
`active users` shows the number of users which logged in at least once.
`disabled users` shows the number of users which are disabled.
 There might be a discrepancy between the total number of users compared to the number of active users and the number of disabled users, as users that have never logged in before are not counted as active nor disabled users.
 Some user backends also do not allow a count for the number of users.
```

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1431,9 +1431,19 @@ authentication servers such as LDAP::
  |                  |    |
  | total users      | 98 |
  |                  |    |
+ | active users     | 15 |
+ | disabled users   | 0  |
+ |                  |    |
  | user directories | 2  |
```
come-nc, line 1437:
```suggestion
 | user directories | 2  |
 | active users     | 15 |
 | disabled users   | 0  |
```
It appears in this order in my command output on 28.

come-nc (approved):

joshtrichards (approved):

PR #11588, opened 2024-02-26
docs(admin): Emphasize recommended versions in db chapter

joshtrichards:
### ☑️ Resolves

More strongly emphasize that picking the *version* of your preferred database is also important. 

- Mention it
- Link to system requirements
- Add a note call-out
- Clarify the steps for db deployment as bullet points for the reader

Context: Some people merely use the `latest` tags their MariaDB Docker stacks or install the latest from vendor repositories without paying attention to the versions recommended by us. Not surprisingly they run into problems from time to time (like the latest breaking changes that were in MariaDB 11.3.x images -- which we don't even support since it's a short-term release). 

Every little nudge in the right direction helps.

Side note: Anyone know why the Database config section isn't just a chapter under *Installation and server configuration*? It's kind of weird that it's arbitrarily near the end of the Admin Manual, since it's a prerequisite to any deployment.

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/configuration_database/linux_database_configuration.rst
+++ b/admin_manual/configuration_database/linux_database_configuration.rst
@@ -10,14 +10,14 @@ Nextcloud requires a database in which administrative data is stored. The follow
 
 The MySQL or MariaDB databases are the recommended database engines.
 
-.. tip:: Not all versions of every supported database are recommended. Please review the Nextcloud :doc:`System Requirements`<../installation/system_requirements>`
+.. tip:: Not all versions of every supported database are recommended. Please review the Nextcloud :doc:`System Requirements`<../installation/system_requirements>
    before settling on a particular version.
 
 Requirements
 ------------
 
 * Decide whether you wish to use MySQL / MariaDB, PostgreSQL, or Oracle as your database
-* Pick a recommendeded version of your database by checking the Nextcloud :doc:`System Requirements`<../installation/system_requirements>`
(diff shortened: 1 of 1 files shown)


kesselb (approved):
Good recommendation, Thank you :pray:

zmweske:
should the mariadb version include v10.6? That is what is listed under https://github.com/nextcloud/docker?tab=readme-ov-file#running-this-image-with-docker-compose

nickvergessen:
DB versions are listed in https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html

zmweske:
Ok, it looks like it was updated since posting the comment: 
(image)
Just wanted to make sure that things were consistent

zmweske:
Looks like it is more updated now: 
image

nickvergessen:
https://github.com/nextcloud/documentation/commit/97910350748d872062dda41e728a098599d63963 2 months ago, but all good

zmweske:
Oh it looks like they are both listed as latest, even though one is definitely not. I'll post the links when I get back to my comp

zmweske:
Yeah, the problem was that both the page title and the menu in the bottom right said latest when it was not: https://docs.nextcloud.com/server/21/admin_manual/installation/system_requirements.html (obv v21 when looking through the url)

joshtrichards:
@zmweske Indeed. Been attempting to fix that (e.g. #11628), but it's a work in progress still. :)

zmweske:
No worries, been there hahaha

PR #11589, opened 2024-02-26
in text: "bruteforce" > "brute force"

alexanderdd:
In description text we should use correct spelling which is "brute force" or "brute-force", not "bruteforce". 

In other parts of the documentation, "brute force" is used: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/bruteforce_configuration.html

See also https://en.wikipedia.org/wiki/Brute-force_attack

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -606,7 +606,7 @@ auth.bruteforce.protection.enabled
 
 	'auth.bruteforce.protection.enabled' => true,
 
-Whether the bruteforce protection shipped with Nextcloud should be enabled or not.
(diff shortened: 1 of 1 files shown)


joshtrichards:
Makes sense, but these descriptions have to be changed in `server`. They're just imported from the distributed sample config

https://github.com/nextcloud/server/blob/master/config/config.sample.php

https://github.com/nextcloud/documentation/blob/508aa2f1efd74ebccdcdf040fabe16cce94490a1/admin_manual/configuration_server/config_sample_php_parameters.rst#L27

alexanderdd:
Thanks, I opened this PR https://github.com/nextcloud/server/pull/43962
and will close this one here

PR #7141, opened 2021-09-24
Fix legacy routes like ajax.php on nginx

ChristophWurst:
Suggestions from @jlehtoranta, taken from https://github.com/nextcloud/documentation/pull/2197#issuecomment-721432337

Without this you can't use the LDAP wizard. With the changes it works.

PR #11596, opened 2024-02-28
Added support for Built-in CODE Server (ARM64)

dgiebert:
Needed to support the arm64 Code Server, also used here https://github.com/nextcloud/helm/pull/540

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -142,7 +142,7 @@ server {
     # to the URI, resulting in a HTTP 500 error response.
     location ~ \.php(?:$|/) {
         # Required for legacy support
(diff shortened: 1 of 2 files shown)


provokateurin (approved):

joshtrichards:
Holy crap! How did we overlook this? Thanks for fixing this! 

Best as I can tell, way back in #2197 the arm64 handling was noted, but the legacy routes were removed outright so it was a non-issue, but then when #7141 added back the legacy route handling it got overlooked.

I've looked at that config probably hundreds of times. :disappointed: 

Needs to be added here too in (via the same PR or one immediately following): https://github.com/nextcloud/documentation/blob/4eca582c9e337e53edace0d872973d1abb2dcaf8/admin_manual/installation/nginx-root.conf.sample#L145

provokateurin (changes requested):
@joshtrichards thanks for catching this, it needs to be fixed in both config files.

provokateurin (approved):

PR #11606, opened 2024-03-04
`mail_domain` clarification re: SMTP EHLO

monochromec:
### ☑️ Resolves

* Fixes misleading `mail_domain` config parameter description (cf. https://github.com/nextcloud/server/issues/43955 ).

### DIFF
--- a/admin_manual/configuration_server/config_sample_php_parameters.rst
+++ b/admin_manual/configuration_server/config_sample_php_parameters.rst
@@ -997,7 +997,7 @@ addresses starting with ``10.0.0.`` and ending with 1 to 3:
 
 Defaults to ``''`` (empty string)
 
-.. _overwrite.cli.url
+overwrite.cli.url
 ^^^^^^^^^^^^^^^^^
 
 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/config_sample_php_parameters.rst
```diff
@@ -736,7 +736,9 @@ mail_domain
 
 The return address that you want to appear on emails sent by the Nextcloud
 server, for example ``nc-admin@example.com``, substituting your own domain,
-of course.
+of course. Please note that this is *not* the domain used for the SMTP protocol handhshake.
+So if you want to have the server responding with an EHLO reply identifiying itself properly,
+please configure ``overwrite.cli.url`` accordingly.
```
tflidd, line 741: I'd rather say that it uses the value of ``overwrite.cli.url``. Because like that, you might get the impression to set this variable accordingly to make the mail delivery work. But it has other consequences as well (it's not use exclusively for the mail EHLO).

joshtrichards:
This is already noted in the actual E-mail configuration chapter:

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/email_configuration.html#configuring-an-smtp-server

> [!NOTE] 
> The `overwrite.cli.url` parameter from `config.php` will be used for the `SMTP EHLO`.

I don't think it needs to be included in the config.sample.php personally (as proposed in this PR) since it's a fairly obscure customization, but if one is going to do that... the `config.sample.php` in this repository is actually pulled from the `server` repository [here](https://github.com/nextcloud/server/blob/master/config/config.sample.php). The PR would have to be against that.

skjnldsv:
Closing as per Josh's comment :)

PR #11607, opened 2024-03-04
Using`mysql.utf8mb4` in via `occ config:system:set` while in maintenance mode

asmith15:
Point out mistake in instructions

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/configuration_database/mysql_4byte_support.rst
+++ b/admin_manual/configuration_database/mysql_4byte_support.rst
@@ -52,7 +52,12 @@ The result should look like this::
     ALTER DATABASE nextcloud CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 
 5. Set the ``mysql.utf8mb4`` config to true in your config.php::
-
+>>>>>>>>>>>>>>>>>>>>
+This command doesn't work if nextcloud is in maintenance mode. It says:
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_database/mysql_4byte_support.rst
```diff
@@ -52,7 +52,12 @@ The result should look like this::
     ALTER DATABASE nextcloud CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 
 5. Set the ``mysql.utf8mb4`` config to true in your config.php::
-
+>>>>>>>>>>>>>>>>>>>>
```
nickvergessen, line 55: This does not work in rst.

Anyway, for me the command still works, even with maintenance mode enabled:
```
$ occ config:system:set 'mysql.utf8mb4' --type boolean --value="true"
Nextcloud is in maintenance mode, no apps are loaded.
Commands provided by apps are unavailable.
System config value mysql.utf8mb4 set to boolean true
```

joshtrichards:
I think this is only applicable to outdated installations. Current patch levels should behave as described in nextcloud/server#37925.

- v26 branch: >=v26.0.2
- v27 branch: >=v27.0.0

kesselb:
Thanks for your report. 

I changed the wording a while ago with https://github.com/nextcloud/documentation/issues/11914.

Should be more clear now.

Issue #11170, opened 2023-10-02
2FA `occ` commands missing from admin manual + outdated use of `twofactor` vs `twofactorauth` keyword


joshtrichards:
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/occ_command.html#two-factor-authentication

```
www-data@06ed34a4d084:~/html$ ./occ twofactor                        

                                       
  Command "twofactor" is not defined.  
                                       
  Did you mean one of these?           
      twofactorauth:cleanup            
      twofactorauth:disable            
      twofactorauth:enable             
      twofactorauth:enforce            
      twofactorauth:state              
                                       

www-data@06ed34a4d084:~/html$ ./occ twofactorauth:                   

                                                                                                                  
  Command "twofactorauth:" is ambiguous.                                                                          
  Did you mean one of these?                                                                                      
      twofactorauth:cleanup Clean up the two-factor user-provider association of an uninstalled/removed provider  
      twofactorauth:enforce Enabled/disable enforced two-factor authentication                                    
      twofactorauth:enable  Enable two-factor authentication for a user                                           
      twofactorauth:disable Disable two-factor authentication for a user                                          
      twofactorauth:state   Get the two-factor authentication (2FA) state of a user.                              
```

skjnldsv:
Fixed — the commands are now documented in occ_users.rst using the correct `twofactorauth` prefix.

PR #11634, opened 2024-03-11
feat(admin): occ twofactor:cleanup/disable

ChristophWurst:
### ☑️ Resolves

* Ref https://github.com/nextcloud/documentation/issues/11170

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -1,3 +1,5 @@
+.. _occ:
+
 =====================
 Using the occ command
 =====================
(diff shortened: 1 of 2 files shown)


joshtrichards (approved):

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1729,13 +1731,13 @@ In the case of a user losing access to the second factor (e.g. lost phone with
 two-factor SMS verification), the admin can try to disable the two-factor
```
joshtrichards, line 1731: Maybe we go ahead and add the full command list to the top of the *Two-factor authentication* section while we're at it - e.g.

The ``twofactorauth`` commands are used to configure/manage various aspects of two-factor authentication::
```
      twofactorauth:cleanup Clean up the two-factor user-provider association of an uninstalled/removed provider  
      twofactorauth:enforce Enabled/disable enforced two-factor authentication                                    
      twofactorauth:enable  Enable two-factor authentication for a user                                           
      twofactorauth:disable Disable two-factor authentication for a user                                          
      twofactorauth:state   Get the two-factor authentication (2FA) state of a user.                              
```

(unfortunately GitHub comments don't let one select lines outside the PR AFAIK so I can't make this a suggestion)

ChristophWurst: I would like that. But for the sake of a minimal backport, I would do it separately.

Issue #9466, opened 2022-12-02
Multiple domains reverse SSL proxy


Ramalama2:
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html#multiple-domains-reverse-ssl-proxy

This is absolutely not understandable, because:
Its called "Multiple domains reverse SSL proxy"
-> So you think you could host nextcloud through a reverse proxy with multiple domains.
-> This is not correct, because:
--> If you set: 'overwritehost' => 'mydomain.com',
---> Nextcloud will redirect 302 every domain to mydomain.com
So in the end this is only a redirection rule, but you cannot host nextcloud with multiple domains.

https://help.nextcloud.com/t/nginx-reverse-proxy-multiple-domains/135730
There are people that asking the same, i think they have the exact same issue, because that snippet above is not understandable.

<br><br>
```
Multiple domains reverse SSL proxy
If you want to access your Nextcloud installation http://domain.tld/nextcloud via a multiple domains reverse SSL proxy https://ssl-proxy.tld/domain.tld/nextcloud with the IP address 10.0.0.1 you can set the following parameters inside the config/config.php.

<?php
$CONFIG = array (
  'trusted_proxies'   => ['10.0.0.1'],
  'overwritehost'     => 'ssl-proxy.tld',
  'overwriteprotocol' => 'https',
  'overwritewebroot'  => '/domain.tld/nextcloud',
  'overwritecondaddr' => '^10\.0\.0\.1$',
);
```
Lets take that documentation and split it up:
- trusted_proxies
-- Absolutely understandable, this is fine, since your proxy has an ip
- overwritehost
-- To What? your proxy internal dns name? (don't forget nextcloud is hosted externally)
-- There is no internal dns name, and if you set this to your nextcloud external dns name of one of your domains, nextcloud will always redirect 302 --> As result this is not "Multiple domains reverse SSL proxy"
-- This is a single domain hosted nextcloud behind a proxy, with redirects.
- overwriteprotocol
-- This is fine and understandable, tho nextcloud will never need this behind a reverse proxy, only in rare circuumstances.
- overwritewebroot
-- This is not fully understandable either. Because same as with the host, so the domain part is actually probably not needed.
- overwritecondaddr
-- I don't understand this either, tho it has the longest description text.

So basically, i could make a PR, with a better description and example.
But someone has to explain to me, what on this "Multiple domains reverse SSL proxy" is actually Multiple?

Cheers

[...]

PR #11638, opened 2024-03-11
fix(configuration): [reverse proxy] Correct errors in `Multiple domains reverse SSL proxy` example

ixtrader:
The ' was missing. Maybe the url off the ssl proxy also?

### DIFF
--- a/admin_manual/configuration_server/reverse_proxy_configuration.rst
+++ b/admin_manual/configuration_server/reverse_proxy_configuration.rst
@@ -167,7 +167,7 @@ you can set the following parameters inside the :file:`config/config.php`.
     'overwriteprotocol' => 'https',
     'overwritewebroot'  => '/domain.tld/nextcloud',
     'overwritecondaddr' => '^10\.0\.0\.1$',
-    'overwrite.cli.url' => 'https://domain.tld/,
+    'overwrite.cli.url' => 'https://ssl-proxy.tld/domain.tld/',
   );
 
 .. note:: If you want to use the SSL proxy during installation you have to


admin_manual/configuration_server/reverse_proxy_configuration.rst
```diff
@@ -167,7 +167,7 @@ you can set the following parameters inside the :file:`config/config.php`.
     'overwriteprotocol' => 'https',
     'overwritewebroot'  => '/domain.tld/nextcloud',
     'overwritecondaddr' => '^10\.0\.0\.1$',
-    'overwrite.cli.url' => 'https://domain.tld/,
+    'overwrite.cli.url' => 'https://ssl-proxy.tld/domain.tld/',
```
joshtrichards, line 170:
```suggestion
    'overwrite.cli.url' => 'https://ssl-proxy.tld/domain.tld/nextcloud',
```

joshtrichards (changes requested):
Good catch! I was looking at this area of the docs last week - seemed like something was wrong here. Thanks for looking closely enough to identify and fix it!

kesselb:
Thanks :+1:

kesselb (approved):

admin_manual/configuration_server/reverse_proxy_configuration.rst
kesselb: I wonder how relevant the above example block is nowadays. 
The time that someone would actually use a ssl proxy because the certificates are too expensive is long over.
We should probably remove the section because people confuse it with reverse proxies used today.

kesselb: See my more elaborated answer: https://github.com/nextcloud/documentation/issues/9466#issuecomment-1345339655

skjnldsv:
Superseded by https://github.com/nextcloud/documentation/pull/14671

PR #11648, opened 2024-03-13
Updated reStructuredText Primer link

Owslla:
The current URL in the documentation does not function. Replaced with the current URL to the related primer.

### ☑️ Resolves

* Not related to a specific issue.
* Fix: Broken link to the ReStructuredText Primer page.

### 🖼️ Screenshots

The appearance is the same. Only the URL was altered.

### DIFF
--- a/style_guide.rst
+++ b/style_guide.rst
@@ -6,7 +6,7 @@ Nextcloud manuals style guide
 
 See the `Documentation README <https://github.com/nextcloud/documentation/blob/master/README.rst>`_ for information on setting up your documentation build environment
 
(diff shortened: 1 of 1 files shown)


Owslla:
I searched and did not see an Issue related to this. Should I create an issue first, then do a pull request against that?  I greatly appreciate your guidance.

joshtrichards (approved):
Thanks. Works for me. Only other variation I can see is maybe linking to the parent instead:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html

But I think your link is fine too.

Owslla:
I had the same thought about linking to the parent, but went with one that matched the original, since I didn't know about any past rational.

PR #11649, opened 2024-03-13
docs(adminmanual): LLM supported languages

hamza221:
### 🖼️ Screenshots
image

### DIFF
--- a/admin_manual/groupware/mail.rst
+++ b/admin_manual/groupware/mail.rst
@@ -93,6 +93,8 @@ LLM Processing
 
 The Mail app can optionally use large language models to process emails and offer assistance features like thread summaries, smart replies and event agendas.
 
+.. note:: The supported languages depend on the used large language model.
+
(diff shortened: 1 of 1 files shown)


admin_manual/groupware/mail.rst
```diff
@@ -93,6 +93,8 @@ LLM Processing
 
 The Mail app can optionally use large language models to process emails and offer assistance features like thread summaries, smart replies and event agendas.
 
+.. note:: The supported languages depend on the used large language model. Our prompts are written in english
```
ChristophWurst, line 96:
```suggestion
.. note:: The supported languages depend on the used large language model. Our prompts are written in English.
```

Not sure if we even need the second sentence. In my testing I had some German emails that worked fine. Just because the prompt is English might not necessarily mean that the data has to be English too.

hamza221: Idk if it's useful information either, I tested with french and arabic and it works.

ChristophWurst (approved):

PR #11682, opened 2024-03-27
feat(dev): Add an example for transient db entity attributes

ChristophWurst:
### 🖼️ Screenshots

image

Example taken from https://www.baeldung.com/jpa-transient-ignore-field


marcelklehr (approved):

mejo- (approved):
Wow, thanks @ChristophWurst :pray:

PR #11684, opened 2024-03-28
conf.py: Set navigation_with_keys to allow navigating documentation through arrow keys

C0rn3j:
Feature from https://github.com/sphinx-doc/sphinx/pull/2064.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -25,6 +25,7 @@ release = version
 # RTD theme options
 html_theme_options = {
 	'logo_only': True,
+	'navigation_with_keys': True,
 	'style_external_links': True,
 	'version_selector': False,
 }


joshtrichards:
Nice idea. Seems to work reasonably. 

The only concern I've come up with: it disables the ability to use the keyboard to scroll sideways when there are big tables (e.g. https://docs.nextcloud.com/server/latest/admin_manual/ai/overview.html). The horizontal scrollbar still works there of course. 

I don't consider that a showstopper, as I consider those wide tables already to be in need of some other clarity to indicate they can be scrolled (since the scrollbar doesn't even appear - to indicate they're scrollable - unless you roll your cursor in front of the table). 

So I'm okay with this. Not sure how others feel about it so I'll leave this PR open a bit before merging so that others can weigh in first.

joshtrichards (approved):

PR #11706, opened 2024-04-05
docs(tuning): add warning for Imaginary+HEIC

pulsejet:
Autorotation is (still) broken [and disabled](https://github.com/nextcloud/server/blob/b5c1766d336e376319ac77289cbc8f6f299e3f20/lib/private/Preview/Imaginary.php#L92-L93
) on HEIC for Imaginary (it just stalls on certain images); this leads to unexpectedly rotated previews scattered around which is hard to debug (e.g. https://github.com/pulsejet/memories/issues/1120).

While this PR just documents the issue, is there still any point recommending Imaginary? I ran some tests and found no evidence to suggest it did any better than Imagick or GD for preview generation in terms of performance (I understand vips is likely better for image manipulation, but that's not really what we're doing here). Actually it's worse for HEIC at least (funny considering IIRC both Imagick and Imaginary use libheif).

Either way it seems counterproductive to recommend something that has known serious issues with a format used by half the mobile devices in the world (at least; even my Samsung shoots HEIC now).

### DIFF
--- a/admin_manual/installation/server_tuning.rst
+++ b/admin_manual/installation/server_tuning.rst
@@ -168,6 +168,11 @@ external microservice: `Imaginary <https://github.com/h2non/imaginary>`_.
    Imaginary is currently incompatible with server-side-encryption. 
    See https://github.com/nextcloud/server/issues/34262
 
+.. warning::
+
+   Imaginary is currently known to have issues with HEIC images.
+   See https://github.com/nextcloud/server/issues/35643
+
 We strongly recommend running our custom docker image that is more up to date than the official image.
 You can find the image at `docker.io/nextcloud/aio-imaginary:latest`.
 


szaimen (approved):
A waning in the docs is fine for now

AndyScherzinger (approved):

AndyScherzinger:
> A waning in the docs is fine for now

I think so too. A reason to use imaginary is that you can externalize that by running it on a completely different server and thus take away server load from the Nextcloud tier in your setup.

pulsejet:
> for me that implies a bit, that it is a known issue that is worked on (and solved in the future). But then the issue should still be open, no? Or we just say it does not support it.

It's a bug either in libheif or vips, not sure. Don't know what the status is.

> A reason to use imaginary is that you can externalize that by running it on a completely different server and thus take away server load from the Nextcloud tier in your setup.

In spirit I agree. The two things seem separable though; it's quite trivial to build an external (PHP-based?) service that uses Imagemagick to generate the previews. This gives the sandboxing security advantage and works on all files too ... just a thought though, not sure if it's worth the effort at the moment.

PR #11724, opened 2024-04-09
fix(developer): Add note about breaking UI change due to heading styles

susnux:
### ☑️ Resolves

* For: https://github.com/nextcloud/server/pull/43640

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.rst
@@ -8,6 +8,13 @@ General
 Front-end changes
 -----------------
 
+Font sizes
+^^^^^^^^^^
+
+Nextcloud now provides meaningful default styles for heading elements.
+This can cause visual regressions if your code does not explicitly set font size and weight.
+If you need to use heading elements outside of text content, you might need to adjust their styles.
+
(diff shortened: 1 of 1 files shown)


developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.rst
```diff
@@ -8,6 +8,13 @@ General
 Front-end changes
 -----------------
 
+Breaking changes
+^^^^^^^^^^^^^^^^
+
```
ChristophWurst, line 13: should this get its own heading?

susnux: Would you put it directly into the one above?

ChristophWurst: replace *Breaking changes* with *Font sizes* and the structure looks good imo

there might be other breaking changes in 30, that's why I would just not put the paragraph as *Breaking changes*

susnux: Done :white_check_mark:

ChristophWurst (approved):
:sunglasses:

Issue #11623, opened 2024-03-08
Impact of using multiple hostnames (on share links)


joshtrichards:
This isn't documented in the spots I'd expect it to be mentioned.

joshtrichards:
Related: nextcloud/server#35942

skjnldsv:
It's severely lacking context 🙈 
Let me see...

skjnldsv:
The linked server issue (nextcloud/server#35942) explains the root cause: Nextcloud generates absolute share URLs based on the hostname it sees on the incoming request. In multi-hostname setups (e.g. internal DNS `cloud.local` + external DNS `cloud.example.com`), the generated URL reflects whichever hostname was used, so a share link created from the internal network breaks for external users.

**Workaround to document:**

Force a single canonical URL for all generated links via `config.php`:

```php
'overwrite.cli.url' => 'https://cloud.example.com',
'overwritehost'     => 'cloud.example.com',
'overwriteprotocol' => 'https',
```

- `overwrite.cli.url`: used by CLI-triggered jobs and as fallback base URL
- `overwritehost`: overrides the hostname in all server-generated URLs regardless of which `trusted_domain` the request arrived on
- `overwriteprotocol`: ensures `https` is used even when the internal request hits over `http`

All three together ensure share links, download URLs, and federation links always use the public hostname.

**Where to document:** The `trusted_domains` / `overwrite*` section of the admin manual (`admin_manual/configuration_server/config_sample_php_parameters.rst`) should cross-reference this, and the sharing chapter should note that share link URLs reflect the canonical host set via `overwritehost`.

PR #11732, opened 2024-04-14
Instructions for domain change

Sommerwiesel:
Add instructions for a domain change.

### ☑️ Resolves

* Gives users instructions on how to perform a domain change
* I have fully tested the method on my production nextcloud with many many apps and many many fiiles and there are no problems

### DIFF
--- a/admin_manual/configuration_server/domain_change.rst
+++ b/admin_manual/configuration_server/domain_change.rst
@@ -2,7 +2,21 @@
 Domain Change
 =====================
 
-Changing the domain after the first setup is currently not supported by Nextcloud. That is mainly because Nextcloud's apps don't support changing the domain after they are set up.
+Changing the domain after the first setup is quite easy. 
 
-.. note::
-    This documentation will get updated with steps what needs to be done if you want to change the domain as soon as this feature is available.
+First, you should have the new domain ready, including DNS, a valid SSL-certificate and the apropriate virtual host entry in the webserver of your choice (apache2, nginx, etc).
+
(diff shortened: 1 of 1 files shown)


joshtrichards:
Hi @Sommerwiesel - Thanks for the addition! I agree it's easy enough to add/change domains by merely adding them to `trusted_domains`, but I'm not sure about this because it gets more complicated when client apps are involved as well as for some server-side apps.

Server-side apps: Some will break and require re-configuration or config resets. `social` (requires a full config reset), `notify_push` (if not configured to use `config.php`), and `richdocuments` (if used with `richdocumentscode`) come to mind. Possibly some of the authentication apps. On the other hand, that's always the case I suppose and we can't cover all apps/scenarios. Plus admin's should expect there will be other spots they need to change domain references I guess (`theming` and `spreed`/talk?).

Clients (desktop, mobile): Will not accept a domain change. New account profiles must get created. Data will have to be re-synced. Profile options will have to be re-entered.

Also, AFAIK even with redirects in place existing shares will break (same as when multiple distinct domains are in-use I presume - #11623).

I just don't want to mislead people that it's just that easy since it's more along the lines of *it might be easy, but it depends*. :-)

Which doesn't mean we can't document precisely that I guess. :thinking: 

Those are my initial thoughts.

kesselb:
I changed the domain for my Nextcloud instance a while ago and found the domain name a couple of times in the database. phpMyAdmin's search feature was really helpful back then. 

image

The domain name is used in oc_circles_event (instance name) and oc_circles_member (note).

szaimen:
> The domain name is used in oc_circles_event (instance name) and oc_circles_member (note) at least.

There are even more places, e.g. look here: https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-change-the-domain

PR #11748, opened 2024-04-19
Added serverinfo configuration options.

mvtab:
Used https://github.com/nextcloud/serverinfo to add serverinfo parameters to the official sample configuration parameters.


mvtab:
System settings do not get propagated to respective apps, this doesn't work.

PR #8084, opened 2022-02-21
Add doc about dry-run option for add-missing-* cmd

artonge:
Documentation for https://github.com/nextcloud/server/pull/31304

PR #11768, opened 2024-04-29
Add option to db indices to occ_command.rst

### LINKED ISSUE #11767: Need to add option for db:add-missing-indices
There is a `--dry-run` option for this function to output the db queries that is not documented yet.

tflidd:
An option `--dry-run` for the adding db-indices seems to be available but not documented (https://github.com/nextcloud/server/blob/2c059dd606f923e5fd689815a09a263bed1c25fd/core/Command/Db/AddMissingIndices.php#L65C17-L65C24)

### ☑️ Resolves

* Fix #11767

### DIFF
--- a/admin_manual/configuration_server/occ_command.rst
+++ b/admin_manual/configuration_server/occ_command.rst
@@ -632,6 +632,8 @@ following command::
 
  sudo -u www-data php occ db:add-missing-indices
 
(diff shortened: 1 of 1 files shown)


joshtrichards (approved):
Looks like the original docs only ended up in the Upgrade section (via #8084).

PR #11801, opened 2024-05-09
fix(admin_manual): Explicitly specify which versions of MySQL

joshtrichards:
### ☑️ Resolves

Rather than say "8.0+" (which could lead to problems with deployments using untested versions of MySQL with Nextcloud) let's stick to the actual versions we test against. 

Currently (in `master`) that is, technically 8.0 and 8.3, but not the recently published 8.4. We are explicit with our other recommended databases so let's be consistent here too. 

Included 8.2 since, like for PostgreSQL, our automated test matrix skips it for efficiency (presumably) since it *likely* is covered by our 8.0 and 8.3 tests (it's out of scope to debate the pros/cons of that here).

### 🖼️ Screenshots

New:
image

Existing:
image

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -11,30 +11,30 @@ For best performance, stability and functionality we have documented some recomm
           reliable scaling) and support, we strongly recommend you to check out our `enterprise support
           <https://nextcloud.com/enterprise/>`_.
 
-+------------------+-----------------------------------------------------------------------+
-| Platform         | Options                                                               |
-+==================+=======================================================================+
-| Operating System | - **Ubuntu 22.04 LTS** (recommended)                                  |
-| (64-bit)         | - Ubuntu 20.04 LTS                                                    |
-|                  | - **Red Hat Enterprise Linux 9** (recommended)                        |
-|                  | - Red Hat Enterprise Linux 8                                          |
-|                  | - Debian 12 (Bookworm)                                                |
-|                  | - SUSE Linux Enterprise Server 15                                     |
-|                  | - openSUSE Leap 15.5                                                  |
-|                  | - CentOS Stream                                                       |
-+------------------+-----------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


admin_manual/installation/system_requirements.rst
```diff
@@ -11,30 +11,30 @@ For best performance, stability and functionality we have documented some recomm
...
+| Operating System | - **Ubuntu 22.04 LTS** (recommended)                                               |
+| (64-bit)         | - Ubuntu 20.04 LTS                                                                 |
+|                  | - **Red Hat Enterprise Linux 9** (recommended)                                     |
+|                  | - Red Hat Enterprise Linux 8                                                       |
+|                  | - Debian 12 (Bookworm)                                                             |
+|                  | - SUSE Linux Enterprise Server 15                                                  |
+|                  | - openSUSE Leap 15.5                                                               |
+|                  | - CentOS Stream                                                                    |
++------------------+------------------------------------------------------------------------------------+
+| Database         | - **MySQL** **8.0**/8.2/8.3 or MariaDB 10.3/10.5/**10.6**/10.11 (recommended)      |
```
susnux, line 26: 8.2 is EOL since 3 month.

susnux, line 26: I think we should only list the LTS version 8.0 and 8.4 (and add 8.4 to our testing matrix).

joshtrichards: Agreed.

kesselb (approved):

admin_manual/installation/system_requirements.rst
nickvergessen: https://schilljs.com

nickvergessen: The ` (recommended)` was inline, to clarify that we recommend the 10.6 instead of any mysql/mariadb
But it seems that got lost already for mysql on the last bump. So let's hope the bolding is enough.

skjnldsv:
Outdated

PR #11805, opened 2024-05-11
fix(trashbin): Add missing size/expire commands to occ commmand list

joshtrichards:
### ☑️ Resolves

Adds to the list of commands:
- nextcloud/server#21658
- nextcloud/server#1138

Doesn't actually show their usage, but at least they're listed now. I didn't even realize `occ trashbin:size` was a command. Kind of handy for setting a size when you aren't using quotas.

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/occ_command.rst
+++ b/admin_manual/occ_command.rst
@@ -1303,14 +1303,16 @@ units.
 Trashbin
 --------
 
-::
+These commands allow for manually managing various aspects of the trash bin (deleted files)::
 
  trashbin
-  trashbin:cleanup  [--all-users] [--] [<user_id>...]  Permanently remove deleted files
-  trashbin:restore  [--all-users] [--scope[=SCOPE]] [--since[=SINCE]] [--until[=UNTIL]] [--dry-run] [--] [<user_id>...]  Restore deleted files according to the given filters
+  trashbin:cleanup      Permanently remove deleted files
+  trashbin:expire       Expires the users trashbin
+  trashbin:size         Configure the target trashbin size
+  trashbin:restore      Restore all deleted files according to the given filters
 
 .. note::
-  This command is only available when the "Deleted files" app
+  These commands are only available when the "Deleted files" app
   (``files_trashbin``) is enabled.
 
 The ``trashbin:cleanup  [--all-users] [--] [<user_id>...]`` command removes the deleted files of the specified


admin_manual/configuration_server/occ_command.rst
```diff
@@ -1272,14 +1272,16 @@ units.
...
 --------
 
-::
+These commands allow for manually managing various aspects of the trash bin (deleted files)::
 
  trashbin
-  trashbin:cleanup  [--all-users] [--] [<user_id>...]  Permanently remove deleted files
-  trashbin:restore  [--all-users] [--scope[=SCOPE]] [--since[=SINCE]] [--until[=UNTIL]] [--dry-run] [--] [<user_id>...]  Restore deleted files according to the given filters
+  trashbin:cleanup      Permanently remove deleted files
+  trashbin:expire       Expires the users trashbin
```
tflidd, line 1279: The `expire` command is not really clear for me, you give a date or something after that it deletes the files?

admin_manual/configuration_server/occ_command.rst
```diff
@@ -1272,14 +1272,16 @@ units.
 Trashbin
 --------
 
-::
+These commands allow for manually managing various aspects of the trash bin (deleted files)::
 
  trashbin
-  trashbin:cleanup  [--all-users] [--] [<user_id>...]  Permanently remove deleted files
-  trashbin:restore  [--all-users] [--scope[=SCOPE]] [--since[=SINCE]] [--until[=UNTIL]] [--dry-run] [--] [<user_id>...]  Restore deleted files according to the given filters
```
tflidd, line 1279: Having the options is really nice, especially with filters because the GUI is limited in this respect.

joshtrichards: Currently in the `occ` command docs, in the list of commands for each area the options are never listed. This one was an oddball and it made the table challenging to read. I agree options can make sense in the docs, but not in that spot currently. Many of the command sections have examples and additional bits after the command listing. Perhaps something there in a follow-up PR.

admin_manual/configuration_server/occ_command.rst
joshtrichards: At present I'm just adding the missing command to the basic list that exists for each command per our current convention.

The description text there is pulled from the command output itself. Further details (the options) are available by running the command. Someone can do a follow-up PR in the spot down below that with examples if they want.

tflidd: Ok, I created a bug report for clarification: https://github.com/nextcloud/server/issues/45418

PR #11828, opened 2024-05-16
chore: remove documentation for user_external

kesselb:
### ☑️ Resolves

Users expect, because the app is mentioned in our admin documentation, that there should be a compatible release and are disappointed if not.

We do not maintain or support user_external, and therefore the mention in our documention should go.

### DIFF
--- a/admin_manual/configuration_user/index.rst
+++ b/admin_manual/configuration_user/index.rst
@@ -10,7 +10,6 @@ User management
     reset_user_password
     user_password_policy
     two_factor-auth
-    user_auth_ftp_smb_imap
     user_auth_ldap
     user_auth_ldap_cleanup
     user_auth_ldap_api
--- a/admin_manual/configuration_user/user_auth_ftp_smb_imap.rst
+++ b/admin_manual/configuration_user/user_auth_ftp_smb_imap.rst
@@ -1,9 +1,0 @@
-==================================================
-User authentication with IMAP, SMB, FTP and others
-==================================================
-
-You may configure additional user backends with the 
-External User Backends (user_external) app.
-
-See https://github.com/nextcloud/user_external#readme 
-for an up to date documentation.


szaimen (approved):

joshtrichards:
Woah, this goes *way* back. It's been around forever.

I'm okay with this, though it brings up a larger discussion of whether it's okay to mention things like this in the docs but somehow make it clearer they're community apps/not fully supported/etc. Sometimes we're going to want to do this I imagine.

That said, the focus of the Admin Manual should be on shipped plus... featured?I guess.

joshtrichards:
Backports desired?

joshtrichards (approved):

violoncelloCH:
> I'm okay with this, though it brings up a larger discussion of whether it's okay to mention things like this in the docs but somehow make it clearer they're community apps/not fully supported/etc. Sometimes we're going to want to do this I imagine.

I think the thing here is that user_external was once created by a core dev and was part of a repo with multiple apps which were (I think) shipped with Nextcloud. That bundled repo was then abandoned and not integrated any more. Then user_external was split out of this and I'm since trying to maintain it in my free time. However my workload for university is currently so high that I don't find time to properly test and feel confident about creating releases for this app, sorry :/

kesselb:
Thanks @violoncelloCH, for the update and maintaining user_external :+1: 

I think it's fine and understandable that university is your top priority.

Aaron-Ritter:
@kesselb Is this something which should get back added in to the core, or maybe at least parts of it?

Especially things like IMAP, HTTP BasicAuth and SSH which is very close to the functionality of nextcloud.

1. IMAP (email) is often defined for each nextcloud user anyway, and using the same login for email and nextcloud is therefore closely connected.
2. HTTP BasicAuth is especially in simple nextcloud setups always a topic
3. SSH especially when you already have a system setup with multiple ssh users on the same system as nextcloud.

Maybe the IMAP part could be integrated back in to nextcloud with the email overhaul?

kesselb:
I don't think so.

PR #11831, opened 2024-05-17
Add details about outgoing connections

Rello:
Adding more details about outgoing connections and the data which is being transferred


derigs (approved):
Checked, all looks good.

Issue #11447, opened 2024-01-17
Standardizing commonly referenced strings throughout the documentation


joshtrichards:
Throughout the documentation there are repeated references to things like:

- various filesystem paths
- `occ` commands
- various URLs for the reader to utilize

These generally (or at least often) vary in different installations/environments (e.g. the folder Nextcloud Server is installed in, nuances of syntax for executing `occ` based on install/runtime method and file/process ownership).

Sometimes these references are made without the proper context for the reader to realize they need to be adapted to their local environment. In other cases some context is provided, but it's inconsistent with how similar references are presented elsewhere in the same manual.

This creates an additional burden on the reader. It also creates a burden on the writer, which has to decide what - if any - context to provide each time. And these inconsistent references become a pain in the butt to monitor/maintain since making updates are not a simple "search and replace" throughout the manual.

I propose we make this better by:

- Coming up with some standards for some of the most common items that fall into this category
- Documenting a standard presentation (in the README/similar)
- Showing reader's how to adapt the standardized upon presentation their own environment somehow (and somewhere in the manual) [the specifics need to be hashed out]

This mostly applies to the Admin Manual (and to a lesser extent the User Manual).

The initial tasks I've come up with so far:

- [x] put this idea out there
- [x] initial a couple initial items that seem like good candidates for this
- [ ] discuss and identify the most common items in this category (comment if you think of one not noted above)
- [ ] explore existing functionality/recommendations for doing things like this with our current documentation platform
- [ ] come up with a standard way of presenting each item (one by one) both in terms of content and style [include the design team in this]
- [ ] determine whether the presentation needs to be explained within the manual in some way [ditto]
- [ ] document the standard way in the README (or, if necessary, an auxiliary document)
- [ ] start replacing references incrementally with the new standard presentation

If we decide to get fancy we could even have dynamically adjusted values (with reasonable defaults), but that's putting the proverbial cart before the horse. Let's figure what and how we want to present things first IMO.

joshtrichards:
# Brainstorming

## Filesystem paths

* Are always stated relative to Nextcloud installation folder and therefore installation folder agnostic
  - e.g. whether Server is installed in `/var/www/html` or `/home/joeuser/cloud.joeuser.com` the references to the following are tall the same:
    - Default data directory path: `/data`
    - Main log file: `/data/nextcloud.log`
    - Main config file: `/config/config.php`
* Where absolute paths are appropriate use them - e.g.
[...]

PR #11838, opened 2024-05-17
fix(admin): Remove "latest" archive reference during upgrades

joshtrichards:
### ☑️ Resolves

- adjusts step 2 to avoid breaking things
  - specifically: installing the latest tarball/zip file isn't always the correct choice; e.g. 27 -> 29 will break their installation
- also adds reminder about checking new system requirements and the standard upgrade requirements
 
### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/maintenance/manual_upgrade.rst
+++ b/admin_manual/maintenance/manual_upgrade.rst
@@ -4,82 +4,93 @@ Upgrade manually
 
 .. seealso::
 
-   If you upgrade from a previous major version please see :ref:`critical changes<critical-changes>` first.
+   If you upgrade from a previous major version please see :ref:`critical changes<critical-changes>`
+   first.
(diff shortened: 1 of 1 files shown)


tflidd (approved):
👍 for the improvements.
Only thing, I don't know if it is perhaps to obvious, that the `www-data` user is mainly for debian-based distributions. Most beginners use that, and other systems with different users might have enough experience to be aware of this. But could be a further improvement (other parts of the documentation are affected by this as well).

joshtrichards:
> Only thing, I don't know if it is perhaps to obvious, that the www-data user is mainly for debian-based distributions.

Yeah I agree. It gets confusing. That's going to take further passes later on (outside of this PR). The docs are littered with stuff like that. Sort of related to #11447 too.

PR #11843, opened 2024-05-20
fix(docs): Fix occ example for dashboard layout

SystemKeeper:
### ☑️ Resolves

The `occ` command for `user:setting` does not have a `--value` option:

```
The "--value" option does not exist.

user:setting [--output [OUTPUT]] [--ignore-missing-user] [--default-value DEFAULT-VALUE] [--update-only] [--delete] [--error-if-not-exists] [--] <uid> [<app> [<key> [<value>]]]
```

image


kesselb (approved):

joshtrichards:
Guess this feature missed getting into the 27 docs by a few days. Hah. Taking care of in nextcloud/documentation#11849 then we can do the backport here again too of this fix.

PR #11873, opened 2024-05-28
fix: 29 is the latest stable version while 28 is now just a normal version

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud/documentation/pull/11710#issuecomment-2135996398


AndyScherzinger (approved):

st3iny (approved):
Thanks

PR #11882, opened 2024-06-03
fix(admin/AI): llm2 & stt_whisper2 still don't support GPU yet

marcelklehr:


joshtrichards:
This might have some upcoming conflicts with #11907...

julien-nc (approved):

PR #11893, opened 2024-06-06
docs(admin): Add note about configuring encoding order for text

juliushaertl:

### ☑️ Resolves

* Fix #…


elzody (approved):
I think it is really good to include this 👍️

Issue #11915, opened 2024-06-20
Fix OpenAPI tutorial Exceptions


provokateurin:
The tutorial currently says that Exceptions should be avoided if possible, but this was on a wrong basis as I previously thought they were not really usable. This was fixed in https://github.com/nextcloud/openapi-extractor/pull/43, but never made it into the docs.

PR #11918, opened 2024-06-20
fix(dev): Adjust Exception explanations in OpenAPI tutorial

provokateurin:
### ☑️ Resolves

* Fix https://github.com/nextcloud/documentation/issues/11915


come-nc (approved):

come-nc:
I still wonder what bad it would do to have the middleware turn all uncatched throwables into json.

provokateurin:
It would certainly make it better, but it's technically a breaking change.

PR #11931, opened 2024-06-25
enable HSTS

wolegis:
according to [Nginx' documentation](http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header) `add_header` settings are inherited to deeper nested server or location blocks only if these deeper blocks do not contain their own `add_header` statements. In our case the relevant block has indeed its own `add_header` statements. Thus the HSTS settings from Nginx' main configuration file are not inherited and need to be reproduced.


susnux (approved):

susnux:
@wolegis Thank you! But could you please sign off your commit? (`git commit --amend -s`)

st3iny (changes requested):
The HSTS header config is now included twice. I see no difference between both versions and HSTS should not be enabled by default in my opinion. Enabling it has severe implications and should be considered thoroughly.

See the whole block:
```

    # HSTS settings
    # although already present in main configuration file this must be reproduced here
    # (see http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header)
    add_header Strict-Transport-Security "max-age=15768000" always;
    # WARNING: Only add the preload option once you read about
    # the consequences in https://hstspreload.org/. This option
    # will add the domain to a hardcoded list that is shipped
    # in all major browsers and getting removed from this list
    # could take several months.
    #add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload" always;
```

wolegis:
> The HSTS header config is now included twice. I see no difference between both versions and HSTS should not be enabled by default in my opinion. Enabling it has severe implications and should be considered thoroughly.

The two `add_header` statements in the cited block are alternatives. Probably, the comments should emphasize this fact.

IMHO, HSTS in itself is desirable. The problematic part is `preload` and the comment pretty clearly indicates the implications.

wolegis:
I've improved the comments (and additionally signed off the latest commit).

skjnldsv:
@st3iny mind giving this PR another look ? :)

PR #11934, opened 2024-06-25
fix: Adjust system requirement for upcoming Nextcloud 30

susnux:
### ☑️ Resolves

Clarify which versions we are testing and thus can support:
* MySQL 8.0 LTS + 8.4 LTS
* MariaDB 10.6 LTS, 10.11 LTS and 11.4 LTS
* *(other databases unchanged)*

See also https://github.com/nextcloud/server/pull/46121

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -23,7 +23,7 @@ For best performance, stability and functionality we have documented some recomm
 |                  | - openSUSE Leap 15.5                                                  |
 |                  | - CentOS Stream                                                       |
 +------------------+-----------------------------------------------------------------------+
-| Database         | - **MySQL 8.0+** or MariaDB 10.3/10.5/**10.6** (recommended)/10.11    |
+| Database         | - MySQL 8.0 / **8.4** or MariaDB 10.6/ **10.11** (recommended) / 11.4 |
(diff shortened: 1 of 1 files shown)


admin_manual/installation/system_requirements.rst
```diff
@@ -23,7 +23,7 @@ For best performance, stability and functionality we have documented some recomm
 |                  | - openSUSE Leap 15.5                                                  |
 |                  | - CentOS Stream                                                       |
 +------------------+-----------------------------------------------------------------------+
-| Database         | - **MySQL 8.0+** or MariaDB 10.3/10.5/**10.6** (recommended)/10.11    |
+| Database         | - MySQL **8.0** / 8.4 or MariaDB 10.6/ **10.11** (recommended) / 11.4 |
```
nickvergessen, line 26: Let's recommend the latest LTS:
```suggestion
| Database         | - MySQL 8.0 / **8.4** or MariaDB 10.6/ **10.11** (recommended) / 11.4 |
```

susnux: Reason I not did this: We have not yet have much experience with 8.4 as it is new to our test chain, but of course we can do so to recommend the latest LTS.
(8.0 has still ~2 years, 8.4 ~7 years)

nickvergessen (approved):

PR #11936, opened 2024-06-25
[stable29] fix(admin): Adjust postgres version to all tested versions

susnux:
### ☑️ Resolves

Nextcloud 29 still support PostgreSQL 10 and 11 (we still test this!)


nickvergessen (approved):

nickvergessen:
I think we usually drop what is EOL and not shipped by a major distro, but fine

PR #11947, opened 2024-06-30
In Files client's API, mention the use of a davClient to retrieve a Node from a file's name

StCyr:
See discussion with @susnux on cloud.nextcloud.com

### ☑️ Resolves

* Fix #…

### DIFF
--- a/developer_manual/client_apis/files.rst
+++ b/developer_manual/client_apis/files.rst
@@ -30,7 +30,7 @@ the ``event-bus`` and use the following events:
   * ``files:node:moved``: the node has been moved (and its data is already updated)
   * ``files:node:updated``: the node data has been updated
 
-All these events use a Node as argument.
+All these events use a Node as argument. You can use a webdav client (for example, by using the davGetClient function from `@nextcloud/files <https://nextcloud-libraries.github.io/nextcloud-files/>`__) to retrieve a node from a file's name.
 
 Examples
 -------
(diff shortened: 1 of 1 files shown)


developer_manual/client_apis/files.rst
```diff
@@ -30,7 +30,9 @@ the ``event-bus`` and use the following events:
   * ``files:node:moved``: the node has been moved (and its data is already updated)
   * ``files:node:updated``: the node data has been updated
 
-Example
+All these events use a Node as argument.
+
+Examples
```
susnux, line 35: Not sure if we really want to move documentation here, as we already have exactly this documentation **and** we already link to (17 lines above).
This functions might change in different versions that might all support the current version of Nextcloud.

So it might be better to just mention the documentation link above ( https://nextcloud-libraries.github.io/nextcloud-files/ )

StCyr: I understand your concern, and it makes sense.

What I could write here could rather be:

```
All these events use a Node as argument.

You can use a webdav client (for example, by using the davGetClient function from (@nextcloud/files)[https://nextcloud-libraries.github.io/nextcloud-files/] to retrieve a node from a file's name.
```

Then I would also update @nextcloud/files documentation to clarify that the webdav client returned is based on [webdav](https://www.npmjs.com/package/webdav)

what do you think?

StCyr: Hello @susnux, I've updated my PR accordingly

susnux (approved):

developer_manual/client_apis/files.rst
```diff
@@ -30,7 +30,9 @@ the ``event-bus`` and use the following events:
   * ``files:node:moved``: the node has been moved (and its data is already updated)
   * ``files:node:updated``: the node data has been updated
 
-Example
+All these events use a Node as argument. You can use a webdav client (for example, by using the davGetClient function from `@nextcloud/files <https://nextcloud-libraries.github.io/nextcloud-files/>`__) to retrieve a node from a file's name.
+
+Examples
 -------
```
susnux, line 36:
```suggestion
Examples
--------
```

PR #11957, opened 2024-07-04
docs(AI/LLM2): Update requirements and document model configuration

marcelklehr:
### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/ai/app_llm2.rst
+++ b/admin_manual/ai/app_llm2.rst
@@ -18,10 +18,15 @@ Requirements
 
 * This app is built as an External App and thus depends on AppAPI v2.3.0 or higher
 * Nextcloud AIO is supported
-* Using GPU is currently not supported
+* We currently support NVIDIA GPUs and x86_64 CPUs
+* GPU Sizing
+
+   * A NVIDIA GPU with at least 8GB VRAM
+   * At least 12GB of system RAM
 
 * CPU Sizing
 
+   * At least 12GB of system RAM
    * The more cores you have and the more powerful the CPU the better, we recommend 10-20 cores
    * The app will hog all cores by default, so it is usually better to run it on a separate machine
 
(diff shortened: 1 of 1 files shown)


admin_manual/ai/app_llm2.rst
```diff
@@ -42,6 +47,42 @@ This app allows supplying alternate LLM models as *gguf* files in the ``/nc_app_
...
+       "stop": ["<|im_end|>"]
+     }
+   }
+
+Here is an example configuration for Llama 3:
+
+.. code-block:: json
+
+   {
+     "prompt": "<|begin_of_text|><|start_header_id|>system<|end_header_id|>\n{system_prompt}<|eot_id|><|start_header_id|>user<|end_header_id|>\n{user_prompt}<|eot_id|>\n<|start_header_id|>assistant<|end_header_id|>\n",
```
kyteinsky, line 78: it's two `\n`s after `<|end_header_id|>`. Doesn't affect the output much but would be better still.
https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-3/

admin_manual/ai/app_llm2.rst
```diff
@@ -18,10 +18,15 @@ Requirements
...
+* We currently support NVIDIA GPUs and x86_64 CPUs
+* GPU Sizing
+
+   * A NVIDIA GPU with at least 8GB VRAM
+   * At least 12GB of system RAM
 
 * CPU Sizing
 
+   * At least 12GB of system RAM
    * The more cores you have and the more powerful the CPU the better, we recommend 10-20 cores
```
kyteinsky, line 30: gpt4all uses only 4 threads by default, so mentioning that would be good to make use of the extra cores.

marcelklehr: We switched to Llama.cpp though

admin_manual/ai/app_llm2.rst
marcelklehr: The notes on that page say: "Newlines are part of the prompt format, for clarity in the examples, they have been represented as actual new lines", so I believe, every empty line in their examples, is *on* `\n`

kyteinsky: hmm, it says that but see here in their repo they use the double `\n`: https://github.com/meta-llama/llama3/blob/main/llama/tokenizer.py#L206-L212

admin_manual/ai/app_llm2.rst
marcelklehr: Yeath, thought about that. Will need to be backward compatible though.

marcelklehr: Let's keep it like this for now, and add an issue on the llm2 repo, once that is changed we can update the docs again.

kyteinsky: sure!

kyteinsky (approved):

PR #12020, opened 2024-07-15
fix(Files): mention enabledPreviewProviders parameter in Previews chapter

joshtrichards:
### ☑️ Resolves

This parameter *is* documented in our config sample and people routinely use it (or ask questions that lead to it). We should mention it in the Previews section. Also mildly tidied up the overall language of the note.

P.S. Can't add direct link labels into config sample sections currently because of how we generate that section of the docs. That's out of scope for this PR.

### 🖼️ Screenshots

Now: 

image

Before:

image

### DIFF
--- a/admin_manual/configuration_files/previews_configuration.rst
+++ b/admin_manual/configuration_files/previews_configuration.rst
@@ -16,12 +16,13 @@ By default, Nextcloud can generate previews for the following filetypes:
 * Cover of MP3 files
 * Text documents
 
-.. note:: Technically Nextcloud can also generate the previews
-          of other file types such as PDF, SVG or various office documents.
-          Due to security concerns those providers have been disabled by
-          default and are considered unsupported.
-          While those providers are still available, we discourage enabling 
-          them, and they are not documented.
+.. note:: Nextcloud can also generate previews of other file types (such as PDF, SVG,
+   various Office document formats, and various video formats). Due to security and
+   performance concerns those providers are disabled by default. While those providers 
+   are still available, we discourage enabling them and they are considered unsupported. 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_files/previews_configuration.rst
```diff
@@ -16,12 +16,13 @@ By default, Nextcloud can generate previews for the following filetypes:
...
-          default and are considered unsupported.
-          While those providers are still available, we discourage enabling 
-          them, and they are not documented.
+.. note:: Nextcloud can also generate previews of other file types (such as PDF, SVG,
+   various Office document formats, and various video formats). Due to security and
+   performance concerns those providers are disabled by default. While those providers 
+   are still available, we discourage enabling them and they are considered unsupported. 
+   The full list of the preview providers that are enabled by default (as well as those 
+   disabled by default) can be found under the ``enabledPreviewProviders`` 
+   :doc:`configuration parameter </configuration_server/config_sample_php_parameters>`.
```
szaimen, line 25: Maybe? (Not sure if this is going to work though)

```suggestion
   :doc:`configuration parameter </configuration_server/config_sample_php_parameters#enabledpreviewproviders>`.
```

joshtrichards: No won't work because those aren't labels from RST's perspective. We can do that if we make a standard URL link (e.g. https://) but that'll create more problems down the road (though it's what we've done in a few other spots for this situation).

I'd like to turn on automatic labels based on headings (across the docs), but it'll require cleaning up various depth 2 & 3 headings. That should work for the config sample.

PR #12024, opened 2024-07-17
feat(admin): Document server autoloading and class maps

ChristophWurst:
### ☑️ Resolves

* Engineers asking how to do this

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/digging_deeper/classloader.rst
+++ b/developer_manual/digging_deeper/classloader.rst
@@ -7,10 +7,22 @@ Classloader
 
 The classloader is provided by Nextcloud and loads all your classes automatically. See :ref:`the composer section<app-composer>` if you want to include and autoload 3rd party libraries.
 
+Server autoloading
+------------------
+
+Classes of the Nextcloud server use an authoritative class map, and the class map is committed to git.
+
+When you add, move or delete a class file you have to update the autoloaders for the class to be found. The process is slightly different for classes that belong to the server itself, and apps that are included in the server repository:
+
+1. Use ``composer dump-autoload`` to update the server class map.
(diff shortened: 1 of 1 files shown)


developer_manual/digging_deeper/classloader.rst
```diff
@@ -7,10 +7,22 @@ Classloader
...
 
+Server autoloading
+------------------
+
+Classes of the Nextcloud server use an authoritative class map, and the class map is committed to git.
+
+When you add, move or delete a class file you have to update the autoloaders for the class to be found. The process is slightly different for classes that belong to the server itself, and apps that are included in the server repository:
+
+1. Use ``composer dump-autoload`` to update the server class map.
+2. Use ``composer -d apps/dav/composer dump-autoload`` to update the class map of the dav app.
```
miaulalala, line 18: Maybe add a sentence that this applies to all apps that are included?

```suggestion
2. Use ``composer -d apps/dav/composer dump-autoload`` to update the class map of the dav app, for example. All shipped apps have their own composer folder where you should run this command.
```

miaulalala (approved):

developer_manual/digging_deeper/classloader.rst
nickvergessen: Or run `bash build/autoloaderchecker.sh && git restore apps/**/installed.php` to update all

ChristophWurst: Works too, but is very slow. And with the many lines printed I have the feeling most engineers don't know what they are doing. So I prefer the more direct approach.

Altahrim (approved):

PR #12027, opened 2024-07-17
admin: Add Ubuntu 24.04 support

susnux:
### ☑️ Resolves

Add Ubuntu 24.04

### 🖼️ Screenshots
image


AndyScherzinger (approved):
is 20.04 not an option anymore with the other system requirements for v30?

Asking because if it is it could still be in there but marked _(deprecated)_

**Edit**: Fine to drop 20.04 - ignore the comment 👍

nickvergessen (approved):

PR #12028, opened 2024-07-17
feat(admin): Add note about `webp` static assets

susnux:
### ☑️ Resolves

* For https://github.com/nextcloud/server/pull/46543


ChristophWurst (approved):
Nice

PR #12044, opened 2024-07-19
fix: EoL 27 from index + move EoL desktop versions to dedicated section

joshtrichards:
### ☑️ Resolves

* Drop Server 27 as *last supported*
* Moves EoL Desktop versions to their own section, just as we do with EoL Server versions
* Adds a link to the *Maintenance and Release Schedule* at the top of the Server section

P.S. (desktop) I believe it's time to bump Desktop 3.13 to *latest* stable and 3.12 to *previous* stable. I don't believe there is an *upcoming* doc base deployed yet, but might be time for that too.

### DIFF
--- a/index.html
+++ b/index.html
@@ -247,7 +247,7 @@
 
 							<div class="section" id="nextcloud-28">
 								<h2>Nextcloud 28<a class="headerlink" href="#nextcloud-28" title="Permalink to this headline">¶</a></h2>
-								<p>This documents the <em>previous</em> version of Nextcloud.</p>
+								<p>This documents the <em>last supported</em> version of Nextcloud.</p>
 								<ul class="simple">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -287,92 +277,27 @@ <h2>Desktop 3.12<a class="headerlink" href="#desktop-312" title="Permalink to th
 										<li><a class="reference external" href="https://docs.nextcloud.com/desktop/3.12">Manual</a>
 									</ul>
 								</div>
-
-								<div class="section" id="desktop-311">
```
nickvergessen, line 291: previous should probably stay for now

nickvergessen: (like keep the last "3" that are still supported)

nickvergessen (approved):

PR #12073, opened 2024-07-29
Warn admins not to disable file locking when using object storage as primary storage

claucambra:
### ☑️ Resolves

Multiple admins, due to advice on the internet (e.g. https://autoize.com/s3-compatible-storage-for-nextcloud/), have disabled file locking when using object storage as primary storage. This note warns them not to do so


skjnldsv:
@claucambra what's the status here? Could you update the labels please

PR #12078, opened 2024-07-31
docs(AI): Add more details to requirements

marcelklehr:


admin_manual/ai/app_context_chat.rst
```diff
@@ -31,6 +31,10 @@ Requirements
 * CPU Sizing
 
    * At least 12GB of system RAM
+   * Below version 3, 10-20 Cores, the more cores the faster it will be
```
kyteinsky, line 34:
```suggestion
   * Below version 3, 10-20 Cores, the more cores (physical cores) the faster the prompt processing will be. Overall performance will increase with an increase in memory bandwidth (more memory sticks and/or higher DDR version)
```

PR #12100, opened 2024-08-10
Update nginx-root.conf.sample

michnovka:
Add version so its easy to check when updates are needed

Right now from time to time when sth breaks I have to open nextcloud nginx docs page and compare line by line with my local webserver setup to see where new changes were introduced. This will simplify it greatly, provided that people who change the documentation update the version date when relevant changes are made to the conf file.


michnovka:
Hi @susnux can you please have a look? every new upgrade I have to manually compare the nginx file revisions to make sure nothing is broken. This would make it way easier. Thanks!

susnux (approved):

PR #12120, opened 2024-08-23
feat(Admin/AI): add docs for translate2

kyteinsky:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

image
image


DaphneMuller:
Looks good. Under the known limitations:

- I'm not sure if the first point of "likely to generate false information" applies to a translation model. Is this thing likely to hallucinate and make up information? If not, you can remove this

Add:
- AI translations are not a replacement for human professional translations and in many cases post-editing is required. AI translations can be used for understanding the main content of a text but not for translations that require special knowledge (such as technical content or legal content), or translations that require specific writing style to convey style, deeper meaning, or emotions (such as marketing content or translating books). 
- While the quality of the output will be fine for the most common languages (English, French, Spanish) the quality will suffer for languages that have less coverage in the original training set.

PR #12135, opened 2024-08-28
feat(dev): Add algorithms supported by oc:checksum prop and OC-Checksum header

provokateurin:
### ☑️ Resolves

Derived from https://github.com/nextcloud/desktop/blob/master/src/common/checksumconsts.h.
Since the server doesn't do any validation the prop/header is basically free form and we can't really know what values to expect.

### DIFF
--- a/developer_manual/client_apis/WebDAV/basic.rst
+++ b/developer_manual/client_apis/WebDAV/basic.rst
@@ -252,7 +252,10 @@ Supported properties
 |                               |                                                 | |     ``<nc:type>0</nc:type>``                                                       |
 |                               |                                                 | | ``</nc:sharee>``                                                                   |
 +-------------------------------+-------------------------------------------------+--------------------------------------------------------------------------------------+
-| <oc:checksums />              | An array of checksums.                          | ``<oc:checksum>md5:04c36b75222cd9fd47f2607333029106</oc:checksum>``                  |
+| <oc:checksums />              | An array of checksums stored in the DB by other | ``<oc:checksum>md5:04c36b75222cd9fd47f2607333029106</oc:checksum>``                  |
+|                               | clients.                                        |                                                                                      |
+|                               | Currently used algorithms are ``MD5``, ``SHA1``,|                                                                                      |
+|                               | ``SHA256``, ``SHA3-256``, ``Adler32``.          |                                                                                      |
 +-------------------------------+-------------------------------------------------+--------------------------------------------------------------------------------------+
 | <nc:has-preview />            | Whether a preview of the file is available.     | ``true`` or ``false``                                                                |
 +-------------------------------+-------------------------------------------------+--------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


artonge:
> Since the server doesn't do any validation the prop/header is basically free form and we can't really know what values to expect.

So, is it really helpful to add a list here?

provokateurin:
I think so, since it is important for clients to know what algorithms to expect. I don't know if the other files clients do anything about the checksums, but if the desktop client supports these any other client supporting checksums should support at least these as well.
I added these docs because I was trying to understand it myself as there is very little documentation about this feature.

developer_manual/client_apis/WebDAV/basic.rst
```diff
@@ -253,6 +253,8 @@ Supported properties
 |                               |                                                 | | ``</nc:sharee>``                                                                   |
 +-------------------------------+-------------------------------------------------+--------------------------------------------------------------------------------------+
 | <oc:checksums />              | An array of checksums.                          | ``<oc:checksum>md5:04c36b75222cd9fd47f2607333029106</oc:checksum>``                  |
+|                               | Supported algorithms are ``MD5``, ``SHA1``,     |                                                                                      |
```
artonge, line 256: Making it sound less official is probably better.

```suggestion
|                               | Currently used algorithms are ``MD5``, ``SHA1``,     |                                                                                      |
```

developer_manual/client_apis/WebDAV/basic.rst
```diff
@@ -253,6 +253,8 @@ Supported properties
 |                               |                                                 | | ``</nc:sharee>``                                                                   |
 +-------------------------------+-------------------------------------------------+--------------------------------------------------------------------------------------+
 | <oc:checksums />              | An array of checksums.                          | ``<oc:checksum>md5:04c36b75222cd9fd47f2607333029106</oc:checksum>``                  |
```
artonge, line 255: Maybe this helps too?

```suggestion
| <oc:checksums />              | An array of checksums stored in the DB by other clients.                          | ``<oc:checksum>md5:04c36b75222cd9fd47f2607333029106</oc:checksum>``                  |
```

joshtrichards:
Interestingly, there is an entire section about checksum handling in the Desktop docs, but it's currently not being included when the docs get built so it's not visible. You can find the entire section here in the raw code:

https://github.com/nextcloud/desktop/blob/5b08c7aed7995dee90b4fd7038d1daaf8a7bbc8e/doc/architecture.rst?plain=1#L120

provokateurin:
Yeah I saw that as well. It's a bit confusing why it is commented out, but it also seems wrong. It tells you the server actually validates the checksums, but from my research and the docs here this is actually not the case.
The last change to that section of the desktop docs is a long time ago, so I doubt anyone still knows why it was done.

artonge (approved):

PR #12146, opened 2024-08-29
Update nginx-root.conf.sample

tofuSCHNITZEL:
### ☑️ Resolves

* prevents access to grunt, package and composer files that could help an attacker to get information about the system (used packages, version etc.)


tflidd:
You speak of code files? That is then also in the official repositories, no? You can just get the version of a setup from https://example.com/status.php

tofuSCHNITZEL:
> You speak of code files? That is then also in the official repositories, no? You can just get the version of a setup from https://example.com/status.php

not sure what you are refering to.
These additions tto the nginx.conf prevent information disclosure of the nextcloud installation and were recomended to me after a security audit of my nextcloud install.

provokateurin:
Hiding these files will not make your setup more secure. The versions of any part of Nextcloud can be found out in multiple ways which can then be used to figure out the content of these files since everything is publicly available.
Also see https://nextcloud.com/security/threat-model/#version-disclosure

tflidd:
Perhaps some of the packaging info can be removed after the packages are built. (most of the files you excluded already reside in folders that are excluded as well).

provokateurin:
Those files are kept for legal/license reasons to trace back dependencies.

skjnldsv:
@copilot does it make sense?
Do you think there are more to look for ?

skjnldsv:

skjnldsv:
Yeah, I agree with the previous comments too, let's close :)

PR #12147, opened 2024-08-29
fix(encryption): Link to current version of recovery tools

joshtrichards:
### ☑️ Resolves

* Fix nextcloud/server#47594

We were still linking to the archived version that isn't up-to-date with our current encryption implementation.


yahesh (approved):
LGTM.

PR #12169, opened 2024-09-06
chore: update available events

kesselb:
### ☑️ Resolves

Update list of available events.

Sorry about the cluttered list, the first version of the event finder did not order the result by class names.


provokateurin (approved):

provokateurin:
Would it be possible to generate this list automatically in the docs (either every time or using a cron workflow)? Then we no longer have to worry about updating the list

kesselb:
> Would it be possible to generate this list automatically in the docs (either every time or using a cron workflow)? Then we no longer have to worry about updating the list

Yep, that the idea.

Generating the rst from the nextcloud/ocp package is done: https://github.com/kesselb/nextcloud-ocp-events/ (that's also how I did this pr, update_events.php, cp event.rst _available_ocp_events.rst, git commit).

The next steps are:

- Make it a proper package
- Thinking how to handle the different stable branches and if that's necessary at all
- Script the update process to open a pull request, like we do for the psalm baseline update

I don't know when I have time to continue, but it's on my radar.

PR #12195, opened 2024-09-10
Feature/nginx conf

istvan-berenyi:
### ☑️ Resolves

http2 conf was changed so it conforms to new NGINX formatting requirements

* Fix #
Not aware that it was logged before


istvan-berenyi:
Updated both config files as per your suggestion @susnux . Thank you!

susnux (approved):

PR #12224, opened 2024-09-22
fix(upgrade_to_30): Note the drop of MariaDB 10.5 support

joshtrichards:
### ☑️ Resolves

* Fix nextcloud/server#48271

So apparently we dropped official support for MariaDB 10.5 (which is an LTS and still supported thru 5/2025 per https://mariadb.org/about/#maintenance-policy) in v30. Earlier in the v30 dev cycle I'd kept 10.5 in (for nextcloud/server#45241). Not sure if it was an oversight or deliberate, but looks like it got dropped outright in nextcloud/server#46121. The new cut-off became 10.6.

Anyhow, adding a note to the release notes section to further call it out (the system requirements are accurate/up-to-date with nextcloud/server#46121; this is just to call it out more prominently since this is a notable change).

I'd personally still vote for keeping 10.5 in, but either way let's keep the docs clear.

### DIFF
--- a/admin_manual/release_notes/upgrade_to_30.rst
+++ b/admin_manual/release_notes/upgrade_to_30.rst
@@ -8,6 +8,7 @@ System requirements
 * PHP 8.1 is now deprecated but still supported.
 * PHP 8.0 is no longer supported.
 * PostgreSQL 9.4 is no longer supported.
+* MariaDB 10.5 is no longer supported.
 
 Web server configuration
 ------------------------
--- a/admin_manual/release_notes/upgrade_to_30.rst
+++ b/admin_manual/release_notes/upgrade_to_30.rst
@@ -8,7 +8,7 @@ System requirements
 * PHP 8.1 is now deprecated but still supported.
 * PHP 8.0 is no longer supported.
 * PostgreSQL 9.4 is no longer supported.
-* MariaDB 10.5 is no longer supported.
+* MariaDB 10.3 and 10.5 are no longer supported.
 
 Web server configuration
 ------------------------


nickvergessen:
The PR at https://github.com/nextcloud/server/pull/46121 says 10.5 was never an LTS.
cc @susnux

joshtrichards:
It appears 10.5 *is* LTS and still supported into next year:

https://mariadb.org/about/#maintenance-policy

admin_manual/release_notes/upgrade_to_30.rst
```diff
@@ -8,6 +8,7 @@ System requirements
 * PHP 8.1 is now deprecated but still supported.
 * PHP 8.0 is no longer supported.
 * PostgreSQL 9.4 is no longer supported.
+* MariaDB 10.5 is no longer supported.
```
susnux, line 11:
```suggestion
* MariaDB 10.3 and 10.5 are no longer supported.
```

susnux:
> The PR at [nextcloud/server#46121](https://github.com/nextcloud/server/pull/46121) says 10.5 was never an LTS.

I think the confusion is coming from their LTS model change, so back then they there was only 10.11 listed as LTS (and 11.4 as upcomin). But all the other 10.x were just not listed as LTS. 10.6 was / is supported and required as it is part of many distributions.

There should be not real problem with dropping 10.5 as it is not included in any distribution but all have either <= 10.3 or 10.6+. Meaning 10.5 is really nowhere used except you manually install it (docker or custom package).

susnux (approved):

PR #12226, opened 2024-09-24

zenlord:
Correcting the information in the *note* in section "Service Discovery" - it appears to not be true in all circumstances, so I added a nuance and a reference to a scenario in which it is not true

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/issues/general_troubleshooting.rst
+++ b/admin_manual/issues/general_troubleshooting.rst
@@ -287,12 +287,12 @@ Make sure to change /nextcloud to the actual subfolder your Nextcloud instance i
 
 .. note:: If you put the above directives directly into an Apache
    configuration file (usually within ``/etc/apache2/``)
-   instead of ``.htaccess``, you need to prepend the first argument of
+   instead of ``.htaccess``, you may need to prepend the first argument of
    each ``RewriteRule`` option with a forward slash ``/``, for example
    ``^/\.well-known/carddav``.
    This is because Apache normalizes paths for the use in ``.htaccess``
    files by dropping any number of leading slashes, but it does not
-   do so for the use in its main configuration files.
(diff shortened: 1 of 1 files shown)


susnux:
Thank you for the pull request!
I am not sure here as the docs say in vhost configurations the leading slash is always included:
https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule

It sounds weird to me that Debian should patch their apache package to break with upstream behavior?

zenlord:
Hi,
I was not familiar with the official docs, but am I misreading the 5th bullet point under 'per-directory' settings? This looks like you would never need a forward slash at the start of your rewrite rules?

> The removed prefix always ends with a slash, meaning the matching occurs against a string which never has a leading slash. Therefore, a Pattern with ^/ never matches in per-directory context.

Anyway, I'm not an expert, so if you believe this PR is not correct, don't feel bad about rejecting it :).

susnux:
>  This looks like you would never need a forward slash at the start of your rewrite rules?

Only if in `.htaccess` or `<Directory>` context.

zenlord:
> >  This looks like you would never need a forward slash at the start of your rewrite rules?
> 
> Only if in `.htaccess` or `<Directory>` context.

That is correct, but I also read:

`Although rewrite rules are syntactically permitted in <Location> and <Files> sections (including their regular expression counterparts), this should never be necessary and is unsupported. A likely feature to break in these contexts is relative substitutions.`

I did indeed place the Rewrite-block in a directory section, which explains why it is working for me, but at the same time renders it clear that the PR is not entirely correct

skjnldsv:
I think we can close then. We lack expertise and a reproducible example. I haven't seen other complains about our default apache conf for debian users.
If that rise again, I guess we can revive this PR!

Thanks for the help and attention @zenlord !
Sorry we took too long to reply :disappointed:

PR #12231, opened 2024-09-26
fix(dev): Use non-ambigous array syntax in OpenAPI tutorial

provokateurin:
### ☑️ Resolves

https://github.com/nextcloud/openapi-extractor/issues/33
https://github.com/nextcloud/openapi-extractor/pull/168

This fix is fully backwards compatible, so we can already advertise it.

### DIFF
--- a/developer_manual/client_apis/OCS/ocs-openapi.rst
+++ b/developer_manual/client_apis/OCS/ocs-openapi.rst
@@ -143,7 +143,7 @@ For those endpoints returning ``[]`` in PHP is a problem because the consumer wi
 If you are not able to use ``null`` for whatever reason, use ``new \stdClass()`` instead.
 It will get correctly converted into ``{}`` in the JSON response on Nextcloud 28 and later.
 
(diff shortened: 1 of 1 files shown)


come-nc (approved):
`list<empty>` and `array<empty>` are the same thing, but okay.

nickvergessen:
> list<empty> and array<empty> are the same thing, but okay.

Only in PHP, it's problematic in JSON and therefore spills over to clients.
https://3v4l.org/42qO0 that is exactly where we are coming from

come-nc:
> > list and array are the same thing, but okay.
> 
> Only in PHP, it's problematic in JSON and therefore spills over to clients. https://3v4l.org/42qO0 that is exactly where we are coming from

I did not write list and array but `list<empty>` and `array<empty>`. Both are `[]`.

provokateurin:
It's a bit easier to handle it this way though, then the checks for `array<TYPE>` are simpler.

PR #12232, opened 2024-09-26
Fixed webhook equals operator from $eq to $e

Lumpiness:
In the webhook listeners [source code](https://github.com/nextcloud/server/blob/4febe107e65927d9be298e2b45869343b7073874/apps/webhook_listeners/lib/Service/PHPMongoQuery.php#L252), the `equals` function is called by the `$e` operator, but not by `$eq`


come-nc (approved):

come-nc:
@Lumpiness Could you fix DCO on the commit as explained in https://github.com/nextcloud/documentation/pull/12232/checks?check_run_id=30692244879 ?

PR #12233, opened 2024-09-26
show which modules are enabled by default

jameskimmel:
### ☑️ Resolves

* show which modules are enabled by default


provokateurin:
I think this will depend a lot on the distro, so we can't do this really

susnux:
I agree this is not consistent across Linux distributions. e.g. `-zlib` is not always enabled.

PR #12234, opened 2024-09-26
remove php-exif

jameskimmel:
### ☑️ Resolves

* not needed, since it is already enabled by default.

### DIFF
--- a/admin_manual/installation/php_configuration.rst
+++ b/admin_manual/installation/php_configuration.rst
@@ -59,7 +59,6 @@ Required for specific apps:
 Recommended for specific apps (*optional*):
 
 * PHP module gmp (for SFTP storage)
-* PHP module exif (for image rotation in pictures app)
 
 For enhanced server performance (*optional*) select one or more of the following
 caches:


tflidd:
I don't see that it is required by the default setup of Nextcloud:
https://docs.nextcloud.com/server/latest/admin_manual/installation/php_configuration.html

jameskimmel:
> I don't see that it is required by the default setup of Nextcloud: https://docs.nextcloud.com/server/latest/admin_manual/installation/php_configuration.html

It isn't. But also if you wanna make use of it, there is no need to install the module.
Another possibility would be to add (included with PHP) like we did for the fileinfo module in the same doc.

kesselb:

How PHP is packaged depends on your operating system. It got better over the years and is now very similar for most distributions.

I usually check https://pkgs.org/download/php-exif for such questions. Ubuntu, Debian, Fedora and others bundle it with common. OpenSuse and Alpine have an extra package.

As there's no harm by listening php-exif here and an admin can always use `php -m|grep exif` to check if the module is there I would rather not merge it.

jameskimmel:
That makes total sense. Sorry that I wasn't aware of how this is handled.
Thank you for taking the time to look at it!

PR #12260, opened 2024-10-02
Clarify username for postgresql's peer authentication method

mario-campos:
The documentation is not clear that the database user name should be the name of the PHP process when using PostgreSQL's peer authentication method. This was only hinted at in a subsequent paragraph, after I had run the PSQL commands with the user name `username`.

I also updated the example to use the common `www-data` user name, which will probably familiar to most users.

### ☑️ Resolves

* Fix #…


mario-campos:
> Replacing it with "www-data" makes it valid for e.g. Ubuntu but not for other distros using different user.

That's fair, but then why not replace it with something more useful and universal (e.g. `php-process-username`)?

PR #12263, opened 2024-10-03
enh(Admin/AI): update background job worker docs

kyteinsky:

### DIFF
--- a/admin_manual/ai/app_context_chat.rst
+++ b/admin_manual/ai/app_context_chat.rst
@@ -47,18 +47,32 @@ Installation
 
 0. Make sure the :ref:`Nextcloud Assistant app<ai-app-assistant>` is installed
 1. :ref:`Install AppAPI and setup a Deploy Demon<ai-app_api>`
-2. Install the *context_chat_backend* ExApp via the "External Apps" admin page in Nextcloud
+2. Install the *context_chat_backend* ExApp via the "External Apps" admin page in Nextcloud, or by executing
+
+.. code-block::
+
+   occ app_api:app:register context_chat_backend
+
 3. Install the *context_chat* app via the "Apps" page in Nextcloud, or by executing
 
 .. code-block::
 
    occ app:enable context_chat
 
-4. Optionally, run two instances of this occ command for faster processing of requests:
+4. Install a text generation backend like *llm2* (via the "External Apps" page in Nextcloud) or *integration_openai* (via the "Apps" page in Nextcloud), or by executing
 
 .. code-block::
 
-   occ background-job:worker 'OC\TaskProcessing\SynchronousBackgroundJob'
+   occ app_api:app:register llm2
+
+or
+
+.. code-block::
+
+   occ app:enable integration_openai
+
+
+5. Optionally but recommended, setup background workers for faster pickup of tasks. See :ref:`the relevant section in AI Overview<ai-overview_improve-ai-task-pickup-speed>` for more information.
 
 **Note**: Both apps need to be installed and both major version and minor version of the two apps must match for the functionality to work (ie. "v1.3.4" and "v1.3.1"; but not "v1.3.4" and "v2.1.6"; and not "v1.3.4" and "v1.4.5"). Keep this in mind when updating.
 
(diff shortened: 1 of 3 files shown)


kyteinsky:
Should we revert to using "task throughput" instead of "task pickup speed"? I thought the latter was simpler since throughput can be confused with AI inference speed.

admin_manual/ai/app_context_chat.rst
```diff
@@ -69,15 +83,15 @@ Context chat will automatically load user data into the Vector DB using backgrou
 
 .. code-block::
 
-   occ background-job:worker 'OCA\ContextChat\BackgroundJobs\StorageCrawlJob'
+   occ background-job:worker -t 60 'OCA\ContextChat\BackgroundJobs\StorageCrawlJob'
```
julien-nc, line 86: Not sure if `-t 60` should be added here because the occ command does not run in a loop.

admin_manual/ai/overview.rst
```diff
@@ -175,6 +175,80 @@ Apps can integrate their content with Context Chat to make it available for quer
...
+----------------------------
+
+Most AI tasks will be run as part of the background job system in Nextcloud which only runs jobs every 5 minutes by default.
+To pick up scheduled jobs faster you can set up background job workers that process AI tasks as soon as they are scheduled.
+
+Screen or tmux session
+^^^^^^^^^^^^^^^^^^^^^^
+
+Run the following occ command inside a screen or a tmux session, preferably 4 or more times for parallel processing of multiple requests by different or the same user (and as a requirement for some apps like context_chat).
+It would be best to run one command per screen session or per tmux pane to keep the logs visible and the worker easily restartable.
```
julien-nc, line 190:
```suggestion
It would be best to run one command per screen session or per tmux window/pane to keep the logs visible and the worker easily restartable.
```

admin_manual/ai/overview.rst
```diff
@@ -175,6 +175,80 @@ Apps can integrate their content with Context Chat to make it available for quer
...
+
+.. code-block::
+
+   set -e; while true; do sudo -u www-data occ background-job:worker -v -t 60 "OC\TaskProcessing\SynchronousBackgroundJob"; done
+
+You may want to adjust the number of workers and the timeout to your needs. The above command will run with a timeout of 60 seconds which means any changes to the settings or the code will be picked up after 60 seconds (worst case scenario). This timeout does not, in any way, affect the processing or the timeout of the AI tasks.
```
julien-nc, line 196: It's not obvious why the workers need to be restarted. Maybe one more sentence would make it clearer. Something like: `If the Php code or the Nextcloud settings values are changed while a worker is running, those changes won't be effective inside the runner. For that reason, the worker needs to be restarted regularly.`
Or even a bit more explicit and detailed. Wdyt?
This more detailed explanation would fit well in the intro (after "Most AI tasks will be run as part of the background job system in Nextcloud...").

admin_manual/ai/overview.rst
```diff
@@ -175,6 +175,80 @@ Apps can integrate their content with Context Chat to make it available for quer
...
+   [Install]
+   WantedBy=multi-user.target
+
+2. Create a shell script in ``/opt/nextcloud-ai-worker/taskprocessing.sh`` with the following content and make sure to make it executable:
+
+.. code-block::
+
+   #!/bin/sh
+   echo "Starting Nextcloud AI Worker $1"
+   sudo -u www-data php occ background-job:worker -t 60 'OC\TaskProcessing\SynchronousBackgroundJob'
```
julien-nc, line 223: Maybe add a line changing directory to the NC one (or use an absolute path).

julien-nc (changes requested):
Noice! A few inline comments.

julien-nc (approved):

admin_manual/ai/app_context_chat.rst
```diff
@@ -69,15 +83,15 @@ Context chat will automatically load user data into the Vector DB using backgrou
 
 .. code-block::
 
-   occ background-job:worker 'OCA\ContextChat\BackgroundJobs\StorageCrawlJob'
+   set -e; while true; do sudo -u www-data occ background-job:worker -v -t 60 "OCA\ContextChat\BackgroundJobs\StorageCrawlJob"; done
```
marcelklehr, line 86: Shouldn't this have two backslashes? :thinking:

kyteinsky: dunno, works fine as is

marcelklehr: alright, nevermind then :)

PR #12264, opened 2024-10-03
Clarify the description of REST vs OCS in accordance to sugestions discussed

christianlupus:
### ☑️ Resolves

* The statement about OCS to be legacy is removed
* A comparison with (plain) REST is made

### 🖼️ Screenshots

T.B.D.

### DIFF
--- a/developer_manual/basics/controllers.rst
+++ b/developer_manual/basics/controllers.rst
@@ -732,10 +732,15 @@ The following policy for instance allows images, audio and videos from other dom
 OCS
 ^^^
 
-.. note:: This is purely for compatibility reasons. If you are planning to offer an external API, go for a :doc:`../digging_deeper/rest_apis` instead.
+In order to simplify exchange of data between the Nextcloud backend and any client (be it the web frontend or whatever else), the OCS API has been introduced.
+Here, JSON and XML responders have been prepared and are installed without additional effort.
 
-In order to ease migration from OCS API routes to the App Framework, an additional controller and response have been added. To migrate your API you can use the **OCP\\AppFramework\\OCSController** base class and return your data in the form of a DataResponse in the following way:
+.. note::
+    The usage of OCS is closely related to the usage of :doc:`../digging_deeper/rest_apis`.
+    Unless you have a clear use-case, it is advised to use OCS over pure REST.
+    A more detailed description can be found in :ref:`ocs-vs-rest`.
 
+To use OCS in your API you can use the **OCP\\AppFramework\\OCSController** base class and return your data in the form of a **DataResponse** in the following way:
 
 .. code-block:: php
 
@@ -759,7 +764,7 @@ In order to ease migration from OCS API routes to the App Framework, an addition
 
 The format parameter works out of the box, no intervention is required.
 
-In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table of your app.
+In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table in ``appinf/routes.php`` of your app.
 Inside these are normal routes.
 
 .. code-block:: php
@@ -777,6 +782,10 @@ Inside these are normal routes.
    ];
 
 Now your method will be reachable via ``<server>/ocs/v2.php/apps/<APPNAME>/api/v1/shares``
+
+.. versionadded:: 29
+    You can use the attribute ``ApiRoute`` as described in :doc:`Routing <routing>` instead of the entry in ``appinfo/routed.php`` as an alternative.
+
 
 Handling errors
 ^^^^^^^^^^^^^^^
--- a/developer_manual/conf.py
+++ b/developer_manual/conf.py
@@ -183,7 +183,7 @@ latex_elements = {
 #'pointsize': '10pt',
 
 # Additional stuff for the LaTeX preamble.
-'preamble': '\extrafloats{100}\maxdeadcycles=500\DeclareUnicodeCharacter{274C}{\sffamily X}',
+'preamble': '\\extrafloats{100}\\maxdeadcycles=500\\DeclareUnicodeCharacter{274C}{\\sffamily X}',
 }
 
 # Grouping the document tree into LaTeX files. List of tuples
--- a/developer_manual/digging_deeper/rest_apis.rst
+++ b/developer_manual/digging_deeper/rest_apis.rst
@@ -6,7 +6,8 @@ REST APIs
 
 .. sectionauthor:: Bernhard Posselt <dev@bernhard-posselt.com>
 
-Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface. It is recommended though to inherit from ApiController and add **@CORS** annotations to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
+Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface.
+It is recommended though to inherit from ApiController and add **@CORS** annotations to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
 
 .. code-block:: php
 
@@ -44,7 +45,8 @@ CORS also needs a separate URL for the preflighted **OPTIONS** request that can 
     )
 
 
-Keep in mind that multiple apps will likely depend on the API interface once it is published and they will move at different speeds to react to changes implemented in the API. Therefore it is recommended to version the API in the URL to not break existing apps when backwards incompatible changes are introduced::
+Keep in mind that multiple apps will likely depend on the API interface once it is published and they will move at different speeds to react to changes implemented in the API.
+Therefore it is recommended to version the API in the URL to not break existing apps when backwards incompatible changes are introduced::
 
     /index.php/apps/myapp/api/1.0/resource
 
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
         }
 
     }
+
+.. _ocs-vs-rest:
+
+Relation of REST and OCS
+------------------------
+
+There is a close relationship between REST APIs and :ref:`OCS <ocscontroller>`.
+Both provide a way to transmit data between the backend of the app in the Nextcloud server and some frontend.
+
+The following combinations of attributes might be useful for various scenarios:
+
+#. Plain frontend route: No special attribute related to CSRF or CORS
+#. Plain Frontend with CRSF checks disabled: Add the ``#[NoCSRFRequired]`` attribute
+#. REST route with CORS enabled: Add the ``#[CORS]`` attribute
+#. OCS-based route
+
+There are different ways a clients might interact with your APIs.
+These ways depend on your API configuration (what you allow) and on which route the request is finally made.
+
+- *Access from web frontend* means the user is browses the Nextcloud web frontend with a browser.
+- *Access from an external app* indicates that the user is not using the normal browser (as logged in) but directly navigates a certain URL.
+  This can be in a browser tab or an external program (like an Android app or simply a curl command line).
+- *Access from external website* means that the user browses some third party web site and *magically* data from your app appears.
+  Technically, the other website would embed/load/use images, JSON data, or other resources from a URL pointing to the Nextcloud server.
+
+.. list-table:: Comparision of different API types
+    :header-rows: 1
+    :align: center
+
+    * - Description
+      - 1 (plain)
+      - 2 (no CSRF)
+      - 3 (CORS)
+      - 4 (OCS)
+    * - URL prefix (relative to server)
+      - ``/apps/<appid>/``
+      - ``/apps/<appid>/``
+      - ``/apps/<appid>/``
+      - ``/ocs/v2.php/apps/<appid>/``
+    * - Access from web frontend
+      - yes
+      - yes (CSRF risk)
+      - yes (CSRF risk)
+      - yes
+    * - Access from external app
+      - --- (CSRF protection blocks)
+      - yes
+      - yes
+      - yes
+    * - Access from external web page
+      - ---
+      - ---
+      - yes
+      - yes
+    * - Transmitted data type
+      - plain data
+      - plain data
+      - plain data
+      - encapsulated data
+
+As a rule of thumb one can conclude that OCS provides a good way to handle most use cases.
+The only exception to this is if you want to provide an API for external usage where you have to comply with an externally defined API scheme.
+Here, the encapsulation introduced in OCS might be in your way.


developer_manual/basics/controllers.rst
```diff
@@ -778,6 +783,10 @@ Inside these are normal routes.
 
 Now your method will be reachable via ``<server>/ocs/v2.php/apps/<APPNAME>/api/v1/shares``
 
+.. versionadded:: 29
+    You can use the attribute ``ApiRoute`` as described in :doc:`Routing <routing>` instead of the entry in ``appinfo/routed.php`` as an alternative.
```
provokateurin, line 787:
```suggestion
    You can use the attribute ``ApiRoute`` as described in :doc:`Routing <routing>` instead of the entry in ``appinfo/routes.php`` as an alternative.
```

developer_manual/conf.py
```diff
@@ -183,7 +183,7 @@
 #'pointsize': '10pt',
 
 # Additional stuff for the LaTeX preamble.
-'preamble': '\extrafloats{100}\maxdeadcycles=500\DeclareUnicodeCharacter{274C}{\sffamily X}',
+'preamble': '\\extrafloats{100}\\maxdeadcycles=500\\DeclareUnicodeCharacter{274C}{\\sffamily X}',
```
provokateurin, line 186: :eyes:

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -6,7 +6,8 @@ REST APIs
 
 .. sectionauthor:: Bernhard Posselt <dev@bernhard-posselt.com>
 
-Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface. It is recommended though to inherit from ApiController and add **@CORS** annotations to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
+Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface.
+It is recommended though to inherit from ApiController and add **@CORS** annotations to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
```
provokateurin, line 10: No relevant diff?

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -44,7 +45,8 @@ CORS also needs a separate URL for the preflighted **OPTIONS** request that can
     )
 
 
-Keep in mind that multiple apps will likely depend on the API interface once it is published and they will move at different speeds to react to changes implemented in the API. Therefore it is recommended to version the API in the URL to not break existing apps when backwards incompatible changes are introduced::
+Keep in mind that multiple apps will likely depend on the API interface once it is published and they will move at different speeds to react to changes implemented in the API.
+Therefore it is recommended to version the API in the URL to not break existing apps when backwards incompatible changes are introduced::
```
provokateurin, line 49: Same here

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+Relation of REST and OCS
+------------------------
+
+There is a close relationship between REST APIs and :ref:`OCS <ocscontroller>`.
+Both provide a way to transmit data between the backend of the app in the Nextcloud server and some frontend.
+
+The following combinations of attributes might be useful for various scenarios:
+
+#. Plain frontend route: No special attribute related to CSRF or CORS
+#. Plain Frontend with CRSF checks disabled: Add the ``#[NoCSRFRequired]`` attribute
```
provokateurin, line 96: CSRF checks should only be disabled when absolutely necessary, this line should have a big warning

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+#. REST route with CORS enabled: Add the ``#[CORS]`` attribute
+#. OCS-based route
+
+There are different ways a clients might interact with your APIs.
+These ways depend on your API configuration (what you allow) and on which route the request is finally made.
+
+- *Access from web frontend* means the user is browses the Nextcloud web frontend with a browser.
+- *Access from an external app* indicates that the user is not using the normal browser (as logged in) but directly navigates a certain URL.
+  This can be in a browser tab or an external program (like an Android app or simply a curl command line).
```
provokateurin, line 105: These two cases are different, as the CSRF checks want to prevent the former while the later should usually be allowed.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+- *Access from external website* means that the user browses some third party web site and *magically* data from your app appears.
+  Technically, the other website would embed/load/use images, JSON data, or other resources from a URL pointing to the Nextcloud server.
```
provokateurin, line 107: This use-case is relatively niche and I think the general advice is to avoid it (and CORS altogether) as it opens unintended security issues if you are not careful. I think this should have some warnings as well.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+      - 4 (OCS)
+    * - URL prefix (relative to server)
+      - ``/apps/<appid>/``
+      - ``/apps/<appid>/``
+      - ``/apps/<appid>/``
+      - ``/ocs/v2.php/apps/<appid>/``
+    * - Access from web frontend
+      - yes
+      - yes (CSRF risk)
+      - yes (CSRF risk)
```
provokateurin, line 126: Maybe explain that for CORS one needs to disable CSRF and thus automatically poses a security risk through CSRF attacks

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+      - yes
+    * - Access from external app
+      - --- (CSRF protection blocks)
+      - yes
+      - yes
+      - yes
```
provokateurin, line 132:
```suggestion
      - no
```
OCS is still protected against CSRF attacks using the OCS-APIRequest header or a CSRF token.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+    * - Access from external web page
+      - ---
+      - ---
+      - yes
+      - yes
```
provokateurin, line 137:
```suggestion
      - no
```
Not true for OCS, unless CORS is enabled

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+    * - Transmitted data type
+      - plain data
+      - plain data
+      - plain data
+      - encapsulated data
```
provokateurin, line 142:
```suggestion
      - plain data
      - plain data
      - plain data
      - encapsulated data (JSON or XML)
```

"plain data" is not very descriptive. In fact any response can be used for any of these methods, just OCS handles the DataResponses differently

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+
+As a rule of thumb one can conclude that OCS provides a good way to handle most use cases.
```
provokateurin, line 144:
```suggestion
As a rule of thumb one can conclude that OCS provides a good way to handle most use cases including sufficient security checks.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+The only exception to this is if you want to provide an API for external usage where you have to comply with an externally defined API scheme.
+Here, the encapsulation introduced in OCS might be in your way.
```
provokateurin, line 146:
```suggestion
Here, the encapsulation introduced in OCS and CSRF checks might be in your way.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+
+Relation of REST and OCS
+------------------------
+
+There is a close relationship between REST APIs and :ref:`OCS <ocscontroller>`.
+Both provide a way to transmit data between the backend of the app in the Nextcloud server and some frontend.
+
+The following combinations of attributes might be useful for various scenarios:
+
+#. Plain frontend route: No special attribute related to CSRF or CORS
```
provokateurin, line 95:
```suggestion
#. Plain frontend route: ``Controller`` class
```

developer_manual/digging_deeper/rest_apis.rst
provokateurin, line 96:
```suggestion
#. Plain Frontend with CRSF checks disabled: ``Controller`` class and ``#[NoCSRFRequired]`` attribute on the method
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+#. Plain Frontend with CRSF checks disabled: Add the ``#[NoCSRFRequired]`` attribute
+#. REST route with CORS enabled: Add the ``#[CORS]`` attribute
```
provokateurin, line 97:
```suggestion
#. REST route with CORS enabled: ``Controller`` class and ``#[CORS]`` attribute on the route
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,66 @@ To add an additional method or header or allow less headers, simply pass additio
...
+#. OCS-based route
```
provokateurin, line 98:
```suggestion
#. OCS-based route: ``OCSController`` class
```

provokateurin, line 98: There is another case with OCS+CSRF

provokateurin (changes requested):
Thanks a lot, this is already super helpful!
The RST syntax for the table really messed with my brain, I need to take a closer look at it in proper table shape.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: This is just styling to put every sentence on it's own line. I can drop that as well.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: These are some of *the few words* I wanted to add...

developer_manual/digging_deeper/rest_apis.rst
christianlupus: Isn't [this line](https://github.com/nextcloud/server/blob/94c529409813e03d632662283a5cd302ef8e9781/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L204) disabling the csrf checks for OCS when done with a bearer token?

developer_manual/digging_deeper/rest_apis.rst
christianlupus: What is former and latter here for you? Browser tab and curl?

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I will add some words here as well

developer_manual/digging_deeper/rest_apis.rst
christianlupus: Yes, good point.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I should be a bit more verbose on this, I see

developer_manual/digging_deeper/rest_apis.rst
provokateurin: Yes

developer_manual/digging_deeper/rest_apis.rst
provokateurin: Not really, it also checks that the custom header is set (https://github.com/nextcloud/server/blob/94c529409813e03d632662283a5cd302ef8e9781/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L235) which is a CSRF check.

christianlupus:
I tried to address the points discussed here. I decided that the CORS should be better put in a separate section as it seems orthogonal to the original question (what should I use Controller or OCSController?).

For simpler reading, I put screenshots of the current state here:

grafik

---

grafik

developer_manual/basics/controllers.rst
```diff
@@ -759,7 +764,7 @@ In order to ease migration from OCS API routes to the App Framework, an addition
 
 The format parameter works out of the box, no intervention is required.
```
provokateurin, line 765: I know you didn't change this, but this sentence is a bit unclear. It talks about the `?format=json` parameter, but doesn't really explain it. IMO using the format parameter is not good, the `Accept: application/json` header should be used instead as it is standardized.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,72 @@ To add an additional method or header or allow less headers, simply pass additio
...
+
+There is a close relationship between REST APIs and :ref:`OCS <ocscontroller>`.
+Both provide a way to transmit data between the backend of the app in the Nextcloud server and some frontend.
+
+The following combinations of attributes might be relevant for various scenarios:
+
+#. Plain frontend route: ``Controller`` class
+#. Plain Frontend with CRSF checks disabled: ``Controller`` class and ``#[NoCSRFRequired]`` attribute on the method
+#. OCS-based route: ``OCSController`` class
+#. OCS-based route with CSRF disabled: ``OCSController`` class and ``#[NoCSRFRequired]`` attribute on the method
```
provokateurin, line 98: I think this case doesn't exist, NoCSRFRequired should not have an effect on OCSController, but I will have to check that in the code to be sure.

provokateurin (changes requested):
Looks pretty good now!

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I was thinking back and forth on this as well. In the end, I decided to keep the case as there is a certain difference (at least from the implementation):

For methods with `NoCSRFRequired`, [this section](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L186-L207) is doing nothing.

For the case where the attribute is not given, the strict cookie check boils down to a check on [the `OCS-APIREQUEST` header](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Http/Request.php#L454-L457) or no valid session token.

With the header in place (as the OCS controller generally expects), the second part is also skipped: The CSRF check is [hardcoded to true](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Http/Request.php#L430-L432), thus [`isInvalidCSRFRequired`](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L226C19-L226C40) returns false.

When the header is not set, you enter the[ second if block](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L195-L207) and will check the inner if. The check `isinstance OCSController` is obviously true. The [`isValidOCSRequest`](https://github.com/nextcloud/server/blob/5bd1873511c214330822b6579a1f5374ed7a5dfe/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L234-L237) is somewhat redundant as it can only be reached with the header unset.  Alternatively, a bearer authorization header is sufficient to make the CSRF checker happy (even without OCS header and without `NoCSRFRequired` attribute).

I was adding this case as a reaction on your statement

> There is another case with OCS+CSRF

In my opinion the last case could be dropped (and was never planned in the first round) but I do not see clearly the motivation of it, to be honest. Did I misunderstand your statement above?

developer_manual/basics/controllers.rst
christianlupus: I tried to address this in the latest commit.
grafik

developer_manual/digging_deeper/rest_apis.rst
provokateurin: > In my opinion the last case could be dropped (and was never planned in the first round) but I do not see clearly the motivation of it, to be honest. Did I misunderstand your statement above?

I messed up, I meant OCS+CORS...
Currently the CORS cases are completely missing in the list.

christianlupus: Yeah, okay, then that makes more sense. I can read the cors posts and extend by the OCS+cors as well.

I will add a warning about cors in general similar to csrf of you do not mind.

christianlupus:
Updated text:

grafik

developer_manual/digging_deeper/rest_apis.rst
christianlupus: If you have any idea how to describe the case *Access from external website* more precisely or otherwise better, feel free to comment :wink:.

christianlupus:
@provokateurin are you still looking some things up, is there any action from my side required at the moment, or was there simply no time to review/prepare merging?

provokateurin:
No sorry, I was just busy with other things. I'll take another look in the coming days :+1:

developer_manual/basics/controllers.rst
```diff
@@ -757,9 +764,13 @@ In order to ease migration from OCS API routes to the App Framework, an addition
 
     }
 
-The format parameter works out of the box, no intervention is required.
+For ``OCSController`` classes and their methods, :ref:`responders <controller-responders>` can be registered as with any other ``Controller`` method.
+The ``OCSController`` class have however automatically two respo    nders pre-installed:
```
provokateurin, line 768:
```suggestion
The ``OCSController`` class have however automatically two responders pre-installed:
```

developer_manual/basics/controllers.rst
```diff
@@ -757,9 +764,13 @@ In order to ease migration from OCS API routes to the App Framework, an addition
...
+Both JSON (``application/json``) and XML (``text/xml``) are generated on-the-fly depending on the request by the browser/user.
+To select the output format, the format parameter or the ``Accept`` header of the request work out of the box, no intervention is required.
```
provokateurin, line 770:
```suggestion
To select the output format, the `?format=` query parameter or the ``Accept`` header of the request work out of the box, no intervention is required.
```

developer_manual/basics/controllers.rst
```diff
@@ -757,9 +764,13 @@ In order to ease migration from OCS API routes to the App Framework, an addition
...
+It is advised to prefer the header generally, as this is the more programmatic way.
 
-In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table of your app.
+In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table in ``appinf/routes.php`` of your app.
```
provokateurin, line 773:
```suggestion
In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table in ``appinfo/routes.php`` of your app.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+------------------------
+
+There is a close relationship between REST APIs and :ref:`OCS <ocscontroller>`.
+Both provide a way to transmit data between the backend of the app in the Nextcloud server and some frontend.
+
+The following combinations of attributes might be relevant for various scenarios:
+
+#. Plain frontend route: ``Controller`` class
+#. Plain frontend with CRSF checks disabled: ``Controller`` class and ``#[NoCSRFRequired]`` attribute on the method
+#. REST route with CORS enabled: ``Controller`` class and ``#[CORS]`` and ``#[NoCSRFRequired]`` attributes on the route
```
provokateurin, line 97:
```suggestion
#. Plain frontend route with CORS enabled: ``Controller`` class and ``#[CORS]`` and ``#[NoCSRFRequired]`` attributes on the route
```
I wouldn't make a differentiation between these

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+#. OCS-based route: ``OCSController`` class
+#. OCS-based route with CORS enabled: ``OCSController`` class and ``#[CORS]`` attribute on the method
```
provokateurin, line 99:
```suggestion
#. OCS route: ``OCSController`` class
#. OCS route with CORS enabled: ``OCSController`` class and ``#[CORS]`` attribute on the method
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+  The CSRF checker will typically fail, so enabling CORS enforces you to disable the CSRF checker as well.
+  Although the disabled CSRF checker in itself is a security issue to consider, adding CORS opens up this even more.
+  You should make sure, that you understand the implications completely when enabling CORS and do so only when there is a good use case.
+
+There are different ways a clients might interact with your APIs.
+These ways depend on your API configuration (what you allow) and on which route the request is finally made.
+
+- *Access from web frontend* means the user is browses the Nextcloud web frontend with a browser.
+- *Access from an external app* indicates that the user is not using the normal browser (as logged in) but directly navigates a certain URL.
+  This can be in a new browser tab or an external program (like an Android app or simply a curl command line).
```
provokateurin, line 116: I don't understand what these two cases have in common.
Security wise they are also a bit different, since you need CSRF protection to prevent redirect attacks via GET while this is completely irrelevant for non-browser use cases like curl.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+- *Access from external website* means that the user browses some third party web site and *magically* data from your app appears.
+  Technically, the other website would embed/load/use images, JSON data, or other resources from a URL pointing to the Nextcloud server.
```
provokateurin, line 118:
```suggestion
- *Access from external website* means that the user browses some third party web site and data from your Nextcloud server appears.
  The other website has to embed/load/use images, JSON data, or other resources from a URL pointing to the Nextcloud server, to be able to do this.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+In contrast, the OCS controller will encapsulate the data in an outer shell that provides some more (meta) information.
+For example a status code (similar to the HTTP status code) is transmitted at top level.
+The actual data is transmitted in the ``data`` property.
+
+As a rule of thumb one can conclude that OCS provides a good way to handle most use cases including sufficient security checks.
+The only exception to this is if you want to provide an API for external usage where you have to comply with an externally defined API scheme.
+Here, the encapsulation introduced in OCS and CSRF checks might be in your way.
+
+.. [#] Only if you have set ``#[NoCSRFRequired]``.
+       OCS controllers have other CSRF checks in place that might with CORS without disabling the CSRF checks completely.
```
provokateurin, line 173:
```suggestion
       OCS controllers have other CSRF checks in place that work with CORS without disabling the CSRF checks completely. 
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+For example a status code (similar to the HTTP status code) is transmitted at top level.
+The actual data is transmitted in the ``data`` property.
+
+As a rule of thumb one can conclude that OCS provides a good way to handle most use cases including sufficient security checks.
+The only exception to this is if you want to provide an API for external usage where you have to comply with an externally defined API scheme.
+Here, the encapsulation introduced in OCS and CSRF checks might be in your way.
+
+.. [#] Only if you have set ``#[NoCSRFRequired]``.
+       OCS controllers have other CSRF checks in place that might with CORS without disabling the CSRF checks completely. 
+       Using the ``OCS-APIREQUEST`` header is also a CSRF protection but is compatible with CORS.
```
provokateurin, line 174:
```suggestion
       Using the ``OCS-APIRequest`` header is a CSRF protection which is compatible with CORS.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,95 @@ To add an additional method or header or allow less headers, simply pass additio
...
+.. [#] The OCS controller needs the request header ``OCS-APIREQUEST`` to be set to ``true``.
```
provokateurin, line 175:
```suggestion
.. [#] The OCS controller needs the request header ``OCS-APIRequest`` to be set to ``true``.
```

developer_manual/conf.py
christianlupus: I just got a warning from sphinx/LaTeX when trying to build the PDF dev documentation locally. I wanted to get rid of any warnings, so this was part of the process.... I suspect this can be seen as resolved.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I think this is solved as part of the warning attached to this section. Feel free to comment and reopen this comment.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: They have in common that the CSRF token is not (guaranteed to be) known as part of the Cookie header and that the user is doing a dedicated action to navigate there (not by using a JS/AJAX request).
You are right these might be considered as separate cases. From user perspective (what can be done? not what is secure?), both navigate to an arbitrary URL and fetch the data. Examples:

- The user navigates to http://cloud.example.com/apps/cookbook/recipe/1234 the basic HTML template should be rendered in the browser and load the recipe 1234. If the actual call to the HTML template would be CSRF-protected any bookmarked link would break.
- An app wants to fetch data from the cookbook using a REST-based API http://cloud.example.com/apps/cookbook/api/v1/recipes/1234. Here, the CSRF checks might cause trouble as well prohibiting legitimate access to the endpoint.

I have not worked out the security side of this. What do you mean by redirect attacks via GET?

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I think, we are arguing from different viewing angles here. Yes, the custom header is some way to enforce CSRF protection. Either a token _or_ the custom header is a way to prevent access to the endpoint.

However, when thinking of external (native) apps, this would allow to access the endpoint just by sending the header without knowing the actual token. 

PS: Having reread the last comment, I see the point in splitting _direct access in browser_ (which you would not do on OCS directly) from the _native app_ approach.

christianlupus:
OK, unfortunately, I had the impression that I needed to restructure the controller chapter. This was a bit back and forth. Many of the changes are just rearrangements. Sorry for the mess. I tried to separate the data from the template requests.

image

As the changes are lengthy, I would have to copy&paste the complete section here. Thus, instead, I post the built documentation for download: [Developer manual.zip](https://github.com/user-attachments/files/18480454/Developer.manual.zip)

If it helps, I could postpone this PR, rebase the restructuring of thee controller class into a new PR, get that merged, and then come back to this PR here.

provokateurin:
Hm yes this is indeed getting a bit hard to review. A separate PR with only the restructuring would be very appreciated.

christianlupus:
Yeah, I see that. I opened #12549 mostly as a cherry-pick from this to first do the restructuring. Then, this can be worked on easier. Until #12549 is resolved, let's freeze this PR.

provokateurin:
Needs a rebase :)

mejo-:
Just a short note from someone who just started to migrate their app to OCS API with OpenAPI support: getting this PR merged would be a great improvement already. Maybe further improvements can be done in follow-up PRs later on to already profit from the clarifications done so far :blush:

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I reread this and we are focusing on the data responses. Thus, the browser bookmark issue is no longer an issue. Instead, a reference to the restructured controller section was added. Thus it boils down to non-browser access as far as I see.

christianlupus:
The main part is in REST vs OCS is this part:

image

---

The updated code in the security guidelines:

image

provokateurin:
I'm a bit busy at the moment, but will give this a proper review in the week after next. I expect that everything will be fine, since we already discussed most changes intensively.

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+The following combinations of attributes might be relevant for various scenarios:
+
+#. Plain frontend route: ``Controller`` class
+#. OCS route: ``OCSController`` class
+#. OCS route with CORS enabled: ``OCSController`` class and ``#[CORS]`` attribute on the method
+
+.. warning::
+  Adding the ``#[NoCRSFRequired]`` attribute imposes a security risk.
+  You should not add this to your controller methods unless you understand the implications and be sure that you absolutely need the attribute.
+  Typically, you can instead use the ``OCS-APIRequest`` header for data requests, instead.
```
provokateurin:
```suggestion
  Typically, you can use the ``OCS-APIRequest`` header for data requests instead, in order to satisfy the CSRF checks in your API requests.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+
+.. warning::
+  Adding the attribute ``#[CORS]`` alone is not sufficient to allow access using CORS with plain frontend routes.
+  Without further measures, the CSRF checker would fail.
+  So, enabling CORS for plain controllers is generally and highly discouraged.
+
+  You would have to disable the CSRF checker (one more security risk) or use the ``OCP-APIRequest`` header to successfully pass the checker.
```
provokateurin:
```suggestion
  You would have to disable the CSRF checks (one more security risk) or use the ``OCP-APIRequest`` header or send a CSRF token to successfully pass the checks.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+  The latter requires dedicated JS code on the importing page.
+
+There are different ways a clients might interact with your APIs.
+These ways depend on your API configuration (what you allow) and on which route the request is finally made.
+
+- *Access from web frontend* means the user is browses the Nextcloud web frontend with a browser.
```
provokateurin:
```suggestion
- *Access from web frontend* means the user is accessing the Nextcloud web frontend with a web browser.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+- *Access from non-browser* is if the user accesses the resource or page not using the default browser.
+  Thus can be e.g. a script using CURL or an Android app.
```
provokateurin:
```suggestion
- *Access from non-browser* is if the user accesses the resource or page using something that is not a web browser, like an Android app or a curl command.```

File: developer_manual/digging_deeper/rest_apis.rst [hunk 33]
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+      - yes
+      - yes
+      - yes
+    * - Access from non-browser
+      - partial [#]_
+      - yes
+      - yes
+    * - Access from external website
+      - ---
+      - ---
```
provokateurin:
```suggestion
      - no
      - no
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+      - yes
+    * - Encapsulated data
+      - no
+      - yes (JSON or XML)
+      - yes (JSON or XML)
+
+.. [#] The external app has to satisfy the CSRF checks.
+       That is, you need to have the ``OCS-APIRequest`` HTTP request header set to ``true``.
+       This is only possible for NC 30 onwards, older versions do not respect the header.
```
provokateurin:
```suggestion
       This is only possible for Nextcloud 30 onwards, older versions do not respect the header.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+
+Methods from ``Controller`` classes can return ``DataResponse`` objects similar to ``OCSController`` class methods.
+For methods of a ``Controller`` class, the data of this response is sent e.g. as JSON as you provide it.
+Basically, the output is very similar to what ``json_encode`` would do.
+In contrast, the OCS controller will encapsulate the data in an outer shell that provides some more (meta) information.
```
provokateurin:
```suggestion
In contrast, the ``OCSController`` will encapsulate the data in an outer shell that provides some more (meta) information.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+Here, the encapsulation introduced in OCS and CSRF checks might be in your way.
+
+
+Historical options
+~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 30
+  The information in this section are mainly for reference purposes. Do not use the approaches in new code.
+
+Before NC server 30 the plain ``Controller`` classes' methods did not respect the ``OCS-APIRequest`` header.
```
provokateurin:
```suggestion
Before Nextcloud 30 the plain ``Controller`` classes' methods did not respect the ``OCS-APIRequest`` header.
```

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -79,3 +81,147 @@ To add an additional method or header or allow less headers, simply pass additio
...
+      - yes
+      - yes (CSRF risk)
+      - yes (CSRF risk)
+    * - Access from non-browser
+      - ---
+      - yes
+      - yes
+    * - Access from external website
+      - ---
+      - ---
```
provokateurin:
```suggestion
      - no
      - yes
      - yes
    * - Access from external website
      - no
      - no
```

developer_manual/prologue/security.rst
```diff
@@ -227,7 +229,11 @@ To prevent CSRF in an app, be sure to call the following method at the top of al
   <?php
   OCP\JSON::callCheck();
 
-If you are using the App Framework, every controller method is automatically checked for CSRF unless you explicitly exclude it by setting the ``#[NoCSRFRequired]`` attribute or ``@NoCSRFRequired`` annotation before the controller method, see :doc:`../basics/controllers`
+If you are using the App Framework, every controller method is automatically checked for CSRF unless you explicitly exclude it by setting the ``#[NoCSRFRequired]`` attribute or ``@NoCSRFRequired`` annotation before the controller method, see :doc:`../basics/controllers`.
+
+Additionally, it is advised to carefully select the HTTP method used for requests.
+Requests of type ``GET`` should not alter data but just read existing data.
+As long as no other attack is involved, any non-``GET`` request requires at least user interaction (transmitting a form).
```
provokateurin: Not sure how relevant this is, since it's easy to submit a form via JS, so it does not really need any user interaction. The two sentences above are good though, maybe they should also mention HTML templates again?

developer_manual/prologue/security.rst
```diff
@@ -250,6 +256,28 @@ Always validate the URL before redirecting if the requested URL is on the same d
...
+However, to protect B and its property (the image), the browsers do not silently embed the image of B into the page of A.
+Instead, B is kindly asked by the browser if embedding is allowed (the so-called `preflight <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_).
+
+To do so, there is a first request made to the resource on B with the ``OPTIONS`` HTTP command/verb.
+The server only answers with the headers as specified and adds ``Access-Control-*`` headers.
+These define, what the browser is to be allowed to do.
+Only if the destination server B confirms cross site resource sharing is allowed, the browser access the resource.
+
+Basically, accessing foreign resources is not limited to embedding images.
+Using JavaScript, arbitrary XHR/Ajax requests can be directed at arbitrary other hosts.
```
provokateurin:
```suggestion
Using JavaScript, arbitrary XHR/Ajax requests can be directed at arbitrary other hosts, which might be used to call APIs that leak your data.
```

developer_manual/basics/controllers.rst
```diff
@@ -547,6 +562,11 @@ Because returning JSON is such a common task, there's even a shorter way to do t
 
 Why does this work? Because the dispatcher sees that the controller did not return a subclass of a Response and asks the controller to turn the value into a Response. That's where responders come in.
 
+.. deprecated:: 30
+
+    Usage of classical controllers for data transmission should be avoided. Use OCS instead.
```
nickvergessen:
```suggestion
    Usage of "index.php"-controllers for data transmission should be avoided. Use OCS instead.
```

developer_manual/conf.py
nickvergessen: Please extract this, so we can check if prod systems building the PDFs have the same issue?

developer_manual/digging_deeper/rest_apis.rst
```diff
@@ -6,7 +6,8 @@ REST APIs
 
 .. sectionauthor:: Bernhard Posselt <dev@bernhard-posselt.com>
 
-Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface. It is recommended though to inherit from ApiController and add ``#[CORS]`` attribute to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
+Offering a RESTful API is not different from creating a :doc:`route <../basics/routing>` and :doc:`controllers <../basics/controllers>` for the web interface.
+It is recommended though to inherit from ApiController and add **@CORS** annotations to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
```
nickvergessen:
```suggestion
It is recommended though to inherit from ApiController and add ``#[CORS]`` attribute to the methods so that `web applications will also be able to access the API <https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS>`_.
```

developer_manual/prologue/security.rst
```diff
@@ -250,6 +256,28 @@ Always validate the URL before redirecting if the requested URL is on the same d
   <?php
   header('Location: https://example.com'. $_GET['redirectURL']);
 
+
+CORS
+----
+
+`Cross-origin resource sharing (CORS) <https://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_ (see also on `MDN <https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS>`__) is a method implemented by browser to access resources from different domains at the same time.
```
nickvergessen, line 263:
```suggestion
`Cross-origin resource sharing (CORS) <https://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_ (see also on `MDN <https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS>`_) is a method implemented by browser to access resources from different domains at the same time.
```

Please add a screenshot if the URL renders with 2 underscores or one

developer_manual/prologue/security.rst
```diff
@@ -250,6 +256,28 @@ Always validate the URL before redirecting if the requested URL is on the same d
...
+Assume, there is a website published on host A.
+The URL would for example be https://A/path/to/index.html.
```
nickvergessen:
```suggestion
The URL would for example be ``https://A/path/to/index.html``.
```

developer_manual/prologue/security.rst
```diff
@@ -250,6 +256,28 @@ Always validate the URL before redirecting if the requested URL is on the same d
...
+If there is a _different_ host B that serves a resource (e.g. an image file) as https://B/assets/image.jpg, the index file on host A could simply link to the image on B.
```
nickvergessen:
```suggestion
If there is a _different_ host B that serves a resource (e.g. an image file) as ``https://B/assets/image.jpg``, the index file on host A could simply link to the image on B.
```

developer_manual/basics/controllers.rst
```diff
@@ -434,6 +438,9 @@ A ``OCP\\AppFramework\\Http\\Template\\SimpleMenuAction`` will be a link with an
 developers can implement their own types of menu renderings by adding a custom
 class implementing the ``OCP\\AppFramework\\Http\\Template\\IMenuAction`` interface.
 
+As the public template is also some HTML template, the same argumentation as for :ref:`regular templates<controller_template>` regarding the CSRF checks hold true:
+The usage of ``#[NoCSRFRequired]`` for public pages is considered acceptable and is actually needed to visit the page without an active account.
```
nickvergessen: This is not true as an isolated statement!?
Not every page for guests should have NoCSRFRequired.
It should at least be limited to such that return templates again, but I'm not sure that is correct. I think it is only necessary (and therefore should be limited to) pages that are entry points.
You can have multiple pages that return HTML templates but you reach the next page by submitting a form or something, then it should only be the first page that has it?
Basically it's:

    Only GET requests returning HTML that do not perform any data manipulation.

Or did I misunderstand something?

nickvergessen (changes requested):

developer_manual/basics/controllers.rst
provokateurin: > You can have multiple pages that return HTML templates but you reach the next page by submitting a form or something, then it should only be the first page that has it?

But the user could reload the page manually, so I think it's necessary to disable csrf for all templates.

nickvergessen: > But the user could reload the page manually, so I think it's necessary to disable csrf for all templates.

In which case they would see a "Form expired …" page from the browser. Otherwise this is exactly what CSRF is about. It should prevent being tricked into a page that is "deep linking" into a flow or to a page that is not a non-interactive frontpage

developer_manual/prologue/security.rst
christianlupus: With two underscores:
image

With one underscore:
image

I do not see any difference in the output to be honest. See also https://stackoverflow.com/a/27435441

developer_manual/digging_deeper/rest_apis.rst
christianlupus: So, OCS with CORS will block an external app but without the CORS attribute it will be accepted? sounds strange. Maybe we can fix this one in Berlin this week if you do not mind...

developer_manual/digging_deeper/rest_apis.rst
christianlupus: These have gone to the legacy section, so I guess this is settled then?

developer_manual/digging_deeper/rest_apis.rst
christianlupus: I think this is settled as well, correct?

developer_manual/digging_deeper/rest_apis.rst
christianlupus: Is this sufficient for you now? Otherwise we discuss in Berlin

developer_manual/digging_deeper/rest_apis.rst
christianlupus: This seems done, I resolve it.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: In legacy code and added there literally. Resolving this.

developer_manual/digging_deeper/rest_apis.rst
christianlupus: As we restricted to the _useful_ combinations, I think this case is not relavant anymore, am I correct?

provokateurin (approved):

PR #12283, opened 2024-10-14
feat(dev): Document return and param types changes for OCP\Files\Storage\IStorage inheritors in 31

provokateurin:
Documentation for:
https://github.com/nextcloud/server/pull/48219
https://github.com/nextcloud/server/pull/48494


come-nc (approved):
Can you write a rector for that? :smiling_imp:

provokateurin:
Hm yeah I guess that should be possible.
I don't know how many other implementations exist out there, so the effort might not be viable.

PR #12329, opened 2024-10-28
NCU unstable namespace

ArtificialOwl:
image

### DIFF
--- a/developer_manual/digging_deeper/api.rst
+++ b/developer_manual/digging_deeper/api.rst
@@ -7,3 +7,24 @@ PHP public API
 
 The public API is contained in the OCP namespace. See the `OCP API reference
 <https://nextcloud-server.netlify.app/>`_ for further details.
+
+
+PHP unstable API
+----------------
+
+To avoid releasing incomplete public API, it is possible to release a
+first version of the future API in the `NCU` namespace, following these rules:
+
+- Files are located in ``/lib/unstable/``
+- Code quality, comments, tests and psalm check are expected to be identical to the `OCP` namespace.
+- Classes must be tag as ``@experimental``, including the current version of Nextcloud.
+- Code from the `OCP` namespace must never mention anything coming from the `NCU` namespace. It can not require it as an argument, constant or return something from `NCU`.
+- An API can only live in this unstable namespace for one major release.
+- During this testing phase, the code and the API can be modified/restructured without limitation.
+- API within the testing namespace must have up-to-date documentation.
+- If accepted, the API will be copied to the `OCP` public namespace.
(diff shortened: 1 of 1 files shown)


developer_manual/digging_deeper/api.rst
```diff
@@ -7,3 +7,19 @@ PHP public API
...
 <https://nextcloud-server.netlify.app/>`_ for further details.
+
+
+PHP testing API
+---------------
+
+To avoid releasing uncompleted public API, it is possible to release a
+testing version of the future API in the NCT namespace, following those rules:
+
+- An API can only be nested in this testing namespace for 1 (one) major release,
```
provokateurin, line 18:
```suggestion
- An API can only live in this testing namespace for one major release,
```
Not sure what you meant with nested

provokateurin (approved):

provokateurin (approved):

ArtificialOwl:
> * OCP code must never mention NCT. It can not require it as an argument, constant or return something from NCT.
> 
>     * NCT will be added to the nextcloud-deps/OCP package for easier testing with psalm

added to documentation

developer_manual/digging_deeper/api.rst
```diff
@@ -7,3 +7,23 @@ PHP public API
 
 The public API is contained in the OCP namespace. See the `OCP API reference
 <https://nextcloud-server.netlify.app/>`_ for further details.
+
+
+PHP testing API
```
susnux, line 12: This sounds like this is about API for testing not API that is in testing state.
Maybe "Unstable PHP API"? As "unstable" is a more common term for this kind of API in general?

 (At least for me "testing" is more like we provide our `\Test` namespace)

nickvergessen: Was wondering similarly, also for the `lib/testing/` folder in Server. Other options could be Experimental or Preview (colliding with previews/thumbnails)

ArtificialOwl: Totally agree on those points, but from previous vocal discussions:

- `testing` is above `unstable` (at least from debian's point of view)
- 'Experimental' would means NCE to be the name of the namespace, which is an already used acronym,
- 'Preview' would also conflict,
- 'devel' can be confusing too, as it can be interpreted as API for development ...

unstable might the only valid term that does not conflict that much and also reflect the 'unstable'/'prone to changes'

ArtificialOwl: So, after discussion with @provokateurin, `testing` have been renamed to `unstable`.

Files location:  `/lib/unstable/`
namespace: `NCU\`

ArtificialOwl:
squashed, screenshot updated, merged.

PR #12339, opened 2024-11-05
chore: Update notes on public sharing key

nfebe:
At first, when the master encryption is enabled the master key is protected by a DEFAULT password generated automatically the docs does not make this clear.


admin_manual/configuration_files/encryption_details.rst
```diff
@@ -26,7 +26,7 @@ This is the default encryption mode in Nextcloud. With master key encryption ena
 Key type: public sharing key
 ----------------------------
 
-The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that can be generated by the server administrator. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
+The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that can be set by the server administrator. A default password is automatically generated once encryption is enabled. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
```
nfebe, line 29:
```suggestion
The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that can be set/changed by the server administrator. A default password is automatically generated once encryption is enabled. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
```

Is this true @artonge i.e When the default password is generated is can be changed by the administrator?

artonge: I don't think so. Better remove it then.

artonge: I can't find a way to do so by checking occ, the web UI settings, and also reading through the code. So I think not.

nfebe: So not possible, leave at the level of auto-generated then.

admin_manual/configuration_files/encryption_details.rst
```diff
@@ -26,7 +26,7 @@ This is the default encryption mode in Nextcloud. With master key encryption ena
 Key type: public sharing key
 ----------------------------
 
-The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that can be generated by the server administrator. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
+The public sharing key is used to secure files that have been publicly shared. The public sharing key is protected by a password that is automatically generated once encryption is enabled. The advantage of the public sharing key is that it is independent of the selected encryption mode so that Nextcloud is able to provide publicly shared files to outside parties.
```
artonge, line 29: Looking at the code, the sharing key is not protected by a password at all.

The master key however, is automatically generated. Paragraph just above.

admin_manual/configuration_files/encryption_details.rst
```diff
@@ -21,12 +21,12 @@ These conventions apply throughout this document:
 Key type: master key
 --------------------
 
-This is the default encryption mode in Nextcloud. With master key encryption enabled there is one central key that is used to secure the files handled by Nextcloud. The master key is protected by a password that can be generated by the server administrator. The advantage of the master key encryption is that the encryption is transparent to the users but has the disadvantage that the server administrator is able to decrypt user files without knowing any user password.
+This is the default encryption mode in Nextcloud. With master key encryption enabled there is one central key that is used to secure the files handled by Nextcloud. The master key is protected by a password that is automatically generated once encryption is enabled. The advantage of the master key encryption is that the encryption is transparent to the users but has the disadvantage that the server administrator is able to decrypt user files without knowing any user password.
```
artonge, line 24:
```suggestion
This is the default encryption mode in Nextcloud. With master key encryption enabled there is one central key that is used to secure the files handled by Nextcloud. The master key is protected by the instance `secret` that is generated at installation time. The advantage of the master key encryption is that the encryption is transparent to the users but has the disadvantage that the server administrator is able to decrypt user files without knowing any user password.
```

artonge (approved):

PR #12341, opened 2024-11-09
fix(developer): Rephrase some parts to avoid archaic wording

susnux:
### ☑️ Resolves

Avoid wording which could be considered offensive.


ChristophWurst (approved):

jancborchardt (commented):
Thank you! :)

PR #12371, opened 2024-11-16
Fix get task by id method

SystemKeeper:
Refs:

https://github.com/nextcloud/server/blob/67a02fa611c8a88dc500cff2c93033bb3db436b7/core/Controller/TaskProcessingApiController.php#L162-L178

https://github.com/nextcloud/server/blob/67a02fa611c8a88dc500cff2c93033bb3db436b7/core/Controller/TextProcessingApiController.php#L140-L155

https://github.com/nextcloud/server/blob/67a02fa611c8a88dc500cff2c93033bb3db436b7/core/Controller/TextToImageApiController.php#L115-L132


SystemKeeper:
/backport bc1c292b367f1a63a5afffedc8325031f81a1ab8 to stable29

SystemKeeper:
/backport bc1c292b367f1a63a5afffedc8325031f81a1ab8 to stable28

SystemKeeper:
Now that I look at it again, I think the note about ratelimit is also wrong here and only belongs to the schedule endpoint?

marcelklehr:
Yes, well spotted! The POST should be a GET, the ratelimit should be for the schedule endpoint. For TaskProcessing the status is a string nowadays, indeed. Initially it was an int.

marcelklehr (approved):

developer_manual/client_apis/OCS/ocs-taskprocessing-api.rst
```diff
@@ -10,4 +10,247 @@ The OCS Task processing API allows you to run task processing tasks, like prompt
 
 See the relevant `OpenAPI docs <https://docs.nextcloud.com/server/latest/developer_manual/_static/openapi.html#/operations/core-task_processing_api-schedule>`_.
 
-.. TODO ON RELEASE: Update version number above on release
```
marcelklehr, line 13: This should probably be kept, I think

marcelklehr (changes requested):

marcelklehr (approved):

SystemKeeper:
That PR differs now from the original idea, let me take another look.

SystemKeeper:
The part I wanted to update, was removed and linked to the OpenAPI docs in https://github.com/nextcloud/documentation/commit/50948b909f4d5944a9e094b161491e53a3a260dc. That's why the diff now looks weird. We can close this then, the OpenAPI docs should be fine.
Thanks for taking a look @rakekniven !

PR #12374, opened 2024-11-18
fix: add sentence about delay in cron execution.

yemkareems:
fix: add sentence about delay in cron execution.

### ☑️ Resolves

* Fix #…


schiessle:
@ArtificialOwl @yemkareems would be great to get it merged. Super small detail but I know people who would be happy about it :wink:

PR #12375, opened 2024-11-19
Update calendar.rst

jefbenet:
added 'step' #4 under import calendar noting the progress bar as this was not clear when I was using the import function as to what the status of the import was.

### ☑️ Resolves

* Fix #…


st3iny (approved):
LGTM, thanks.

PR #12390, opened 2024-11-22
fix(s3): document options for adjusting multipartcopy behavior

joshtrichards:
Added in nextcloud/server#41914

### ☑️ Resolves

* Fixes `pending documentation` label on nextcloud/server#41914


come-nc (approved):
But more documentation about this would be good. Related PR says "Multipart copy seems problematic with some backends so avoid it when possible.".
That should be reflected in the documentation, maybe with a list of which backends have problems or how to detect it?

juliusknorr (approved):

PR #12393, opened 2024-11-25
feat(dev): Add OCP\Files\Mount\IShareOwnerlessMount API for 31

provokateurin:
Documentation for https://github.com/nextcloud/server/pull/49073


nfebe (commented):
This would only show on the upgrade to 31 guide, thereafter this won't be generally easy to find on the documentation. I am not sure if there is already a section for `OCP\Files\Mount\IMountPoint` it can also be mentioned there.

The current note in the upgrade to 31 section, looks good to me though.

nfebe (approved):

PR #12397, opened 2024-11-26
Mark the Projects feature as deprecated 

julien-nc:
and replaced by related resources.

Is it correct that it was announced deprecated from NC 25?

### 🖼️ Screenshots

image

This could be backported down to stable25.

### DIFF
--- a/user_manual/files/projects.rst
+++ b/user_manual/files/projects.rst
@@ -2,12 +2,15 @@
 Projects
 ========
 
+.. deprecated:: 25
+    This feature was replaced by the shipped related resources app.
+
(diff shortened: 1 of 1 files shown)


user_manual/files/projects.rst
```diff
@@ -2,12 +2,15 @@
 Projects
 ========
 
+.. deprecated:: 25
+    This feature was replaced by the shipped `related resources <https://apps.nextcloud.com/apps/related_resources>`_ app.
```
nickvergessen, line 6: https://github.com/nextcloud/server/blob/master/core/shipped.json#L36

Shipped by now, so the link to a no longer updated appstore entry might be confusing as well.

julien-nc: Removed the link.

AndyScherzinger:
Yes, let's backport to v25 👍

nickvergessen:
Docs are only build on 28+ iirc

PR #12399, opened 2024-11-27
[stable29] Fix wrong german translation for 'sharing' in user manual

julien-nc:
There's a mismatch between the user manual and the actual navigation menu entry in the personal settings.
In the navigation in Nextcloud, "Sharing" is translated as "Teilen":

image
 
but in the doc it is translated as "Freigabe".

This has been fixed in Transifex so it will be synced with the master branch of the docs.

Is this the right way to fix the translation issue for the stable29 branch of the docs?

### 🖼️ Screenshots

image


nickvergessen:
Translations are synced from transifex.

nickvergessen:
Hmmm seems we only translate master

nickvergessen (approved):

nickvergessen:
Raised https://github.com/nextcloud/docker-ci/issues/688

PR #12402, opened 2024-11-27
Change example for setting correct permissions.

stelzch:
Using `find -exec chmod … {} \;`  executes one process of chmod for every file/directory, whereas `find -exec chmod … {} +` collects files into a large list and invocates chmod much less.


blizzz:
Question is whether we an run into a limit of arguments or length with `find -exec chmod … {} +`. The original version is safer, but I am not sure how risky it could be, and it may depend as well. 

I'd keep it as is. When you know what you are doing, you can go with the `+` sign anyways.

PR #12406, opened 2024-11-29
fix(client): Fix Out-of-office endpoint and response data

nickvergessen:
### ☑️ Resolves

* Endpoint was missing the `/now` and was duplicating the second documented endpoint
* Data was wrong, see https://github.com/nextcloud/server/blob/370a9d77ea0aadd736d42741623cf98729797d8b/apps/dav/lib/ResponseDefinitions.php#L22-L35

### 🖼️ Screenshots

grafik


Ivansss (approved):
LGTM 👍

PR #10972, opened 2023-08-09
Update to documentation required to get CalDAV working on ios 16.5.1

joealsubash:
### ☑️ Resolves

* Fix #…

PR #12416, opened 2024-12-04
Synchronizing with iOS: Clarify how to use TLS in CalDAV URLs

blipp:
### ☑️ Resolves

When following the [steps in the manual to set up a Nextcloud calender on iOS](https://cloud.systemli.org/remote.php/dav/principals/users/blipp/), it is unintuitive that you do not need to specify the protocol, and that it will actually not work if you specify the protocol, only if you then also specify the port.

Interestingly, Nextcloud gives you the link with the protocol when you use the “copy iOS/macOS CalDAV address” functionality in the web interface. Maybe this is a bug?

### 🖼️ Screenshots

> Please add a screenshot of your changed or added page(s).
> This helps reviewers to quickly see how the resulting
> lists, code blocks, headers and links look.

Unsure how to do this, I am just editing the page in GitHub, I do not have a testing setup.


blipp:
Related pull request: #10972

PR #12419, opened 2024-12-06
enh(Maintenance): update sync with clients after data recovery

kyteinsky:
All other steps are already good.

### 🖼️ Screenshots

image


ChristophWurst (approved):
Nice

ChristophWurst:
nit: `enh` is not a typical conventional commit by most conventions. Use `feat`

kyteinsky:
> nit: `enh` is not a typical conventional commit by most conventions. Use `feat`

cool

PR #12431, opened 2024-12-13
add(admin): add command to change the default page

### LINKED ISSUE #11889: Missing from docs: `defaultpage` (core setting that sets a default page for all users)
`occ config:app:set --value "/apps/files/extstoragemounts" core defaultpage`

Sets a default page for all users (works with more things than the `defaultapp` parameter)

It’s been in place for >10 years through a number of code changes in that area. I suppose we should document it.

The code handling this is [here](https://github.com/nextcloud/server/blob/94bc020d073011eec8e1f084e9521356574e42f7/lib/private/URLGenerator.php#L279-L299).

oscardimanno:
### ☑️ Resolves

* Fix #11889

### 🖼️ Screenshots

immagine

### DIFF
--- a/admin_manual/configuration_server/dashboard_configuration.rst
+++ b/admin_manual/configuration_server/dashboard_configuration.rst
@@ -28,3 +28,9 @@ The default layout of the dashboard for all users can be set with the following 
   occ config:app:set dashboard layout --value="files,activity,calendar"
 
 Changing the default layout will not affect existing users that already have a custom layout stored.
+
+
+It's possible to replace the default dashboard with a custom app with the following command::
+
+  occ config:app:set core defaultpage --value "/apps/files/extstoragemounts"
+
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/dashboard_configuration.rst
```diff
@@ -28,3 +28,9 @@ The default layout of the dashboard for all users can be set with the following
   occ config:app:set dashboard layout --value="files,activity,calendar"
 
 Changing the default layout will not affect existing users that already have a custom layout stored.
+
+
+It's possible to replace the default dashboard with a custom app with the following command::
```
susnux, line 33:
```suggestion
It is possible to replace the default app, which is the dashboard app, with a custom app with the following command:
```

susnux (approved):

PR #12480, opened 2025-01-06
fix(AI/Overview): keep systemd runners alive when restarting the AIO

derigs:
I added some configurations to the Service block of the AI worker. This changes allow the AIO Container keep the runnners alive even when restarting the AIO.

### ☑️ Resolves

* Fix AIO AI sheduler fail after restart.


derigs:
@kyteinsky here is the PR

kyteinsky (approved):

PR #12493, opened 2025-01-09
fix(AppAPI): Apps page instead of External Apps page

kyteinsky:
### ☑️ Resolves

* Fix #…


kyteinsky:
Did I miss any instance?

edward-ly (approved):
That's all I can find as well.

PR #12496, opened 2025-01-09
feat(mail): add message translation documentation

GVodyanov:
### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/issues/10386#issuecomment-2553103406

### 🖼️ Screenshots

User manual:
image

Admin manual:
image

### DIFF
--- a/admin_manual/groupware/mail.rst
+++ b/admin_manual/groupware/mail.rst
@@ -127,3 +127,10 @@ multiple days.
 Each sent email will be analyzed by an AI to check whether a reply is expected.
 
 The feature can be enabled through the global :ref:`mail_llm_processing` setting.
+
+Translation
+-----------
+
+.. versionadded:: 4.2
+
+The mail app can optionally provide translations for messages if the :ref:`translation API <machine_translation>` is enabled.
--- a/user_manual/groupware/mail.rst
+++ b/user_manual/groupware/mail.rst
@@ -329,6 +329,15 @@ When you open a message in the Mail app, it proposes AI-generated replies. By si
 
 .. note:: Supported languages depend on the used large language model
 
+Mail translation
+~~~~~~~~~~~~~~~~
+
+.. versionadded:: 4.2
+
+You are able to translate messages to your configured languages similarly to Talk.
+
+.. note:: Please note that translation features have to be enabled on the server
+
 Thread summary
 --------------
 
--- a/admin_manual/ai/app_assistant.rst
+++ b/admin_manual/ai/app_assistant.rst
@@ -42,6 +42,8 @@ The AI admin settings will show all types of Assistant Tasks that are implemente
 
 **Note**: When using our on-premise AI apps, make sure you have a GPU with enough VRAM that fits all the features you need. For each app documented here you will find its hardware requirements.
 
+.. _machine_translation:
+
 Machine translation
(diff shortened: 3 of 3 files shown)


admin_manual/groupware/mail.rst
```diff
@@ -127,3 +127,8 @@ multiple days.
 Each sent email will be analyzed by an AI to check whether a reply is expected.
 
 The feature can be enabled through the global :ref:`mail_llm_processing` setting.
+
+Translation
+-----------
+
```
ChristophWurst, line 133:
```suggestion

.. versionadded:: 4.2

```

user_manual/groupware/mail.rst
```diff
@@ -320,6 +320,13 @@ When you open a message in the Mail app, it proposes AI-generated replies. By si
 
 .. note:: Supported languages depend on the used large language model
 
+Mail translation
+~~~~~~~~~~~~~~~~
+
```
ChristophWurst, line 325: here too

ChristophWurst (approved):
Looks good. Let's add version info too

ChristophWurst (changes requested):
Preventing premature merge

ChristophWurst:
Mail 4.2 will also be available for Nextcloud 30, so let's backport

admin_manual/groupware/mail.rst
```diff
@@ -127,3 +127,10 @@ multiple days.
...
+.. versionadded:: 4.2
+
+The mail app can optionally provide translations for messages if the `translation API <https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/translation.html>`_ is enabled.
```
ChristophWurst, line 136: I think https://docs.nextcloud.com/server/latest/admin_manual/ai/app_assistant.html#machine-translation would be a better fit. Then we don't leave the admin manual and don't confuse admins with developer stuff ;-)

admin_manual/groupware/mail.rst
```diff
@@ -127,3 +127,10 @@ multiple days.
 Each sent email will be analyzed by an AI to check whether a reply is expected.
 
 The feature can be enabled through the global :ref:`mail_llm_processing` setting.
+
+Translation
+-----------
+
+.. versionadded:: 4.2
+
+The mail app can optionally provide translations for messages if the `translation API <https://docs.nextcloud.com/server/latest/admin_manual/ai/app_assistant.html#machine-translation>`_ is enabled.
```
ChristophWurst, line 136: Apologies for not being clearer but please don't use an absolute link. Use a reference for the hyperlink. We not only generate an online documentation, but also a PDF. The PDF would otherwise open the external link, when it could link inside of the document.
The second problem of the absolute link is also that it will always go to the *latest* version. At some point the current master will become stable30. Then you open the docs for 30 and end up at 31 or even newer.
Lastly, with an anchor and internal reference, sphinx will actually ensure that you are linking to something that exists. We don't run into the problem of dangling/dead links.

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html

GVodyanov: Thanks a lot for the detailed explanation <3

ChristophWurst (approved):
:shipit:

PR #12503, opened 2025-01-11
chore: Update MySQL 4-byte support chapter for current defaults

joshtrichards:
### ☑️ Resolves

`innodb_file_per_table` has been the default for awhile now. We can probably drop it entirely, but for now let's update the instructions to just give the reader a way to confirm.

### DIFF
--- a/admin_manual/configuration_database/mysql_4byte_support.rst
+++ b/admin_manual/configuration_database/mysql_4byte_support.rst
@@ -6,25 +6,17 @@ Enabling MySQL 4-byte support
 
     Be sure to backup your database before performing this database upgrade.
 
-In order to use Emojis (textbased smilies) on your Nextcloud server with a MySQL database, the
-installation needs to be tweaked a bit.
+To be able to handle 4-byte characters with MySQL/MariaDB (like emojis in filenames or comments) 
+it is recommended to enable 4-byte support in MySQL. If you're not sure if you need this, you 
+should probably enable it. This is also the appropriate thing to do if you encounter errors such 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_database/mysql_4byte_support.rst
```diff
@@ -6,25 +6,17 @@ Enabling MySQL 4-byte support
...
-In order to use Emojis (textbased smilies) on your Nextcloud server with a MySQL database, the
-installation needs to be tweaked a bit.
+To be able to handle 4-byte characters with MySQL/MariaDB (like emojis in filenames or comments) 
+it is recommended to enable 4-byte support in MySQL. If you're not sure if you need this, you 
+should probably enable it. This is also the appropriate thing to do if you encounter errors such 
+as ``Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes``.
 
 .. warning::
 
-    This manual only covers MySQL 8 or newer and MariaDB 10.2 or newer.
```
nickvergessen, line 14: I would prefer to keep this for a little longer :see_no_evil: 
Explicit named versions don't hurt

susnux: But then we should adjust the MariaDB version. Maybe just add a reference to system requirements

skjnldsv: Done https://github.com/nextcloud/documentation/commit/51eaa50b7416c5c56bb918a16dd0f0a99eef931f

PR #12544, opened 2025-01-28
Remove User/Whats-New page, because it is not maintained anyway?!

Rello:
The "whats new" just reserves a slot in the primary nav - but it does not provide any useful content.
updated 4 years ago - and also too expensive to keep maintained.

Remove?


susnux (approved):
I think we promote this in firstrunwizard nowadays and on the blog

PR #12549, opened 2025-01-29
Restructure significant part of the controller documentation

christianlupus:
This mainly restructures the controller section.

### ☑️ Resolves

* Preparation for #12264: Restructuring 

#### Routing

Screenshot 2025-01-29 at 14-03-44 Routing — Nextcloud latest Developer Manual latest documentation

#### Controllers

Screenshot 2025-01-29 at 14-01-50 Controllers — Nextcloud latest Developer Manual latest documentation

### DIFF
--- a/developer_manual/basics/controllers.rst
+++ b/developer_manual/basics/controllers.rst
@@ -335,7 +335,6 @@ Cookies can be set or modified directly on the response class:
         }
    }
 
-
 Responses
 ---------
 
@@ -1093,9 +1092,6 @@ The following policy for instance allows images, audio and videos from other dom
         }
 
     }
-
-
----
 
 .. [1] Even though the header name ``OCS-APIRequest`` hints purely at OCS controllers, with NC 30 classic controller methods respect this header as well.
        Until NC 30, classical controller methods did not respect the header.
--- a/developer_manual/basics/controllers.rst
+++ b/developer_manual/basics/controllers.rst
@@ -686,6 +686,34 @@ A file download can be triggered by returning a DownloadResponse:
 Creating custom responses
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
+If no premade Response fits the needed use case, it is possible to extend the Response base class and custom Response. The only thing that needs to be implemented is the **render** method which returns the result as string.
+
+Creating a custom XMLResponse class could look like this:
+
+.. code-block:: php
+
+    <?php
+    namespace OCA\MyApp\Http;
+
+    use OCP\AppFramework\Http\Response;
+
+    class XMLResponse extends Response {
+
+        private array $xml;
+
+        public function __construct(array $xml) {
+            $this->addHeader('Content-Type', 'application/xml');
+            $this->xml = $xml;
+        }
+
+        public function render(): string {
+            $root = new SimpleXMLElement('<root/>');
+            array_walk_recursive($this->xml, array ($root, 'addChild'));
+            return $xml->asXML();
+        }
+
+    }
+
 Streamed and lazily rendered responses
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -815,37 +843,6 @@ The following policy for instance allows images, audio and videos from other dom
 
 
 ---
-
-Creating custom responses
-^^^^^^^^^^^^^^^^^^^^^^^^^
-
-If no premade Response fits the needed use case, it is possible to extend the Response base class and custom Response. The only thing that needs to be implemented is the **render** method which returns the result as string.
-
-Creating a custom XMLResponse class could look like this:
-
-.. code-block:: php
-
-    <?php
-    namespace OCA\MyApp\Http;
-
-    use OCP\AppFramework\Http\Response;
-
-    class XMLResponse extends Response {
-
-        private array $xml;
-
-        public function __construct(array $xml) {
-            $this->addHeader('Content-Type', 'application/xml');
-            $this->xml = $xml;
-        }
-
-        public function render(): string {
-            $root = new SimpleXMLElement('<root/>');
-            array_walk_recursive($this->xml, array ($root, 'addChild'));
-            return $xml->asXML();
-        }
-
-    }
 
 Streamed and lazily rendered responses
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/developer_manual/basics/controllers.rst
+++ b/developer_manual/basics/controllers.rst
@@ -387,6 +387,43 @@ A :doc:`template <front-end/templates>` can be rendered by returning a TemplateR
 Public page templates
 ^^^^^^^^^^^^^^^^^^^^^
 
+For public pages, that are rendered to users who are not logged in to the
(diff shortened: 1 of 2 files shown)


developer_manual/basics/controllers.rst
```diff
@@ -335,11 +335,173 @@ Cookies can be set or modified directly on the response class:
...
+            ]);
+        }
+
+    }
+
+For ``OCSController`` classes and their methods, :ref:`responders <controller-responders>` can be registered as with any other ``Controller`` method.
+The ``OCSController`` class have however automatically two responders pre-installed:
+Both JSON (``application/json``) and XML (``text/xml``) are generated on-the-fly depending on the request by the browser/user.
+To select the output format, the ``?format=`` query parameter or the ``Accept`` header of the request work out of the box, no intervention is required.
+It is advised to prefer the header generally, as this is the more programmatic way.
```
provokateurin, line 481:
```suggestion
It is advised to prefer the header generally, as this is the more standardized way.
```

The `?format=` is pretty bad IMO, since it does not adhere to HTTP standards while the `Accept` header is exactly made for this purpose. I'm not even sure we should explain here that such a query parameter ever existed.

developer_manual/basics/controllers.rst
```diff
@@ -475,9 +475,14 @@ To migrate your API you can use the **OCP\\AppFramework\\OCSController** base cl
...
 
-The format parameter works out of the box, no intervention is required.
+For ``OCSController`` classes and their methods, :ref:`responders <controller-responders>` can be registered as with any other ``Controller`` method.
+The ``OCSController`` class have however automatically two responders pre-installed:
+Both JSON (``application/json``) and XML (``text/xml``) are generated on-the-fly depending on the request by the browser/user.
+To select the output format, the ``?format=`` query parameter or the ``Accept`` header of the request work out of the box, no intervention is required.
+It is advised to prefer the header generally, as this is the more programmatic way.
 
-In order to make routing work for OCS routes you need to add a separate ‘ocs’ entry to the routing table of your app. Inside these are normal routes.
+In order to make routing work for OCS routes you need to add :ref:`a separate 'ocs' entry<routes_ocs>` to the routing table in ``appinfo/routes.php`` of your app.
```
provokateurin, line 483: This should also mention the attributes you can use since a few versions and link to https://github.com/nextcloud/documentation/blob/de2b54eea51e510b55ff1e1d7d22f5f168eea5f3/developer_manual/basics/routing.rst#routing as an explanation of how to use both.

developer_manual/basics/controllers.rst
```diff
@@ -825,6 +825,77 @@ To turn off checks the following *Attributes* can be added before the controller
...
 
+In the following some examples of configurations are given.
+
+Showing an HTML page by the user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A typical app needs an ``index.html`` page to show all content within.
+This page should be visible by all users in the instance.
+Therefore, you need to loosen the restriction from admins only (``#[NoAdminRequired]``).
+Additionally, as the user might not have a CSRF checker cookie set yet, the CSRF checks should be disabled (which is fine as this is a template response).
```
provokateurin, line 835: This is true for all pages that the user visits directly via the browser, not only template responses.
I'm not sure how to word it, such that it's not understood as "just disable it everywhere when something doesn't work".

developer_manual/basics/controllers.rst
```diff
@@ -751,88 +877,23 @@ In order to ease migration from OCS API routes to the App Framework, an addition
...
 
     }
 
-The format parameter works out of the box, no intervention is required.
-
-In order to make routing work for OCS routes you need to add a separate 'ocs' entry to the routing table of your app.
-Inside these are normal routes.
-
-.. code-block:: php
+The ``#[NoAdminRequired]`` is needed here as normal users should be able to access the data in fact.
```
provokateurin, line 886:
```suggestion
The ``#[NoAdminRequired]`` is needed here as normal users should be able to access the data.
```

developer_manual/basics/routing.rst
```diff
@@ -36,6 +34,45 @@ The route array contains the following parts:
...
+
+OCS routes
+----------
+
+Registering of OCS routes requires to use a :ref:`corresponding controller<ocscontroller>`.
+Additionally, the route must be configured to be an OCS route in the router.
+To do so, you use the ``ocs`` key in the ``routes.php`` file instead of the key ``routes``.
+The rest of the structure is the same.
+
+You can of course combine non-OCS with OCS routes.
```
provokateurin, line 47:
```suggestion
You can of course have both index.php and OCS routes.
```
Combining doesn't really make sense in this context.

developer_manual/basics/controllers.rst
provokateurin: Ok I saw you changed the routing explanation in a later commit.
I would still like to have attributes mentioned more clearly though.

provokateurin (changes requested):
This was so much better to review now! :heart:

I'm not sure about https://github.com/nextcloud/documentation/pull/12549/commits/0935fa8ff5764acc025f4b83d01b5259e4adfe0e. Right now the introduction of each chapter will be part of the previous chapter.
In a real book this might make sense, but here people jump around between sections using the TOC or direct links, so they will easily miss the introduction. I would suggest to just make the introductions the first paragraph of their respective chapters.

developer_manual/basics/controllers.rst
christianlupus: I'd keep it for now. n this big PR, we have enough stuff to think about. The deletion of the `?format=` parameter can be postponed (as it is currently part of the docs) into a separate PR. Do you agree?

developer_manual/basics/controllers.rst
christianlupus: How about something like this?

```suggestion
To make routing work for OCS, the route must be registered in the core.
This can be made in two ways:
You can add an attribute `#[ApiRoute]` to the controller method.
Alternatively, you can add :ref:`a separate 'ocs' entry<routes_ocs>` to the routing table in ``appinfo/routes.php`` of your app.
```

developer_manual/basics/controllers.rst
christianlupus: These are just some (three) examples. We could, however, add a sentence that for the sake of the example, the typical case of a template response is used. At least in NC all HTML pages are somewhat templates AFAIK unless you use lazy rendering or such. Am I wrong here?

developer_manual/basics/controllers.rst
christianlupus: The sugestion is resolved, just keeping this open about the `?format=` question

developer_manual/basics/controllers.rst
provokateurin: `This can be done in two ways` but other than that sounds good.

developer_manual/basics/controllers.rst
provokateurin: Right, maybe I was just thinking too much. You can serve custom HTML, but indeed almost everything is template based.

christianlupus:
About the introduction texts in 0935fa: They are in fact in the parent chapters. So, the serve as an introduction on why the subchapter are grouped the way they are right now. We can easily drop this commit it is serves no purpose.

provokateurin (approved):
Nothing to comment on, so merging :rocket:

PR #12558, opened 2025-01-31
fix: Add marker to positions that need adjusting on branch off

nickvergessen:


blizzz:
Not sure this is the right place. Nobody will see it on branch-off, will anyone?

susnux (approved):
```diff
- fix: Add marker to positions that need adjusting on branch off #12558 
+ chore: Add marker to positions that need adjusting on branch off #12558 
```

blizzz:
> The marker is on all other links from #12557 Only missing on new bits. I check all links with `/latest/` every now and then, but there are way too many as all other docs (talk, appstore, vue, mysql, …) also link to latest and should not be replaced

sounds like a task for grep¹ and sed.

¹any flavour of it

PR #12561, opened 2025-01-31
feat: file conversion docs

elzody:
Related to https://github.com/nextcloud/server/issues/49603


elzody:
I don't think the current failures related to the Codespell test are related at all. I never touched those pages, and it looks like it's trying to spell check strings which should not be spell checked.

juliusknorr (approved):

PR #12562, opened 2025-02-01
fix(admin): Update required sqlite to 3.24+

joshtrichards:
### ☑️ Resolves

3.24+ is required for UPSERT

Used by Server v30 in nextcloud/server#45655

Noticed because someone tried to install v30 on Debian 9 ;-)

Refs:

- https://www.sqlite.org/lang_upsert.html
- https://www.sqlite.org/lang_conflict.html

P.S. Might even make sense to bump to 3.35, but this gets us closer.


nickvergessen:
Should also implement a check on the system side and installer

nickvergessen (approved):

Altahrim (approved):

PR #12596, opened 2025-02-06
harden_server.rst: larger HSTS max-age value

xalt7x:
Nowadays, the common recommendation is to set HTTP Strict Transport Security max-age value to at least 1 year.
It's also min. acceptable value for preload lists.   
Please see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preload

### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/installation/harden_server.rst
+++ b/admin_manual/installation/harden_server.rst
@@ -140,7 +140,7 @@ VirtualHost file::
  <VirtualHost *:443>
    ServerName cloud.nextcloud.com
      <IfModule mod_headers.c>
-       Header always set Strict-Transport-Security "max-age=15552000; includeSubDomains"
+       Header always set Strict-Transport-Security "max-age=31536000; includeSubDomains"
      </IfModule>
   </VirtualHost>
 


nickvergessen (approved):
Should then also adjust in server:
https://github.com/nextcloud/server/blob/9836e9b16484582d309c8437ab46d82e34956941/apps/settings/lib/SetupChecks/SecurityHeaders.php#L102-L112

joshtrichards (commented):
I'm not adverse to changing it to 1 year in the docs, but not sure about adjusting our setup check as I commented over there.

Note: Needs to be updated in the docs that cover Nginx too at same time.

xalt7x:
@joshtrichards ,
Well, this value is recommended by many companies, including [Google](https://developer.chrome.com/docs/lighthouse/best-practices/has-hsts) and [Microsoft](https://learn.microsoft.com/en-us/exchange/plan-and-deploy/post-installation-tasks/security-best-practices/configure-http-strict-transport-security-in-exchange-server?view=exchserver-2019). Though it really generally comes alongside with recommendation to use `preload`.   
If this value has no use (and even more so) for server, then I guess I shouldn't change it in documentation as it will create inconsistency...

PR #12726, opened 2025-02-18
Removing database READ-COMMITTED obsolete indications

ThibautPlg:
Hello,

As per this PR https://github.com/nextcloud/server/issues/17947 and an internal NC helpdesk ticket (number 9653557), Nextcloud sets the transaction isolation level automatically on every connection, thus removing the need to configure it on the database engine. 
This PR removes two mentions of "READ COMMITTED" in the latest doc. This can also probably be merged in older versions of the documentation.

Best regards,

### DIFF
--- a/admin_manual/configuration_database/linux_database_configuration.rst
+++ b/admin_manual/configuration_database/linux_database_configuration.rst
@@ -26,15 +26,6 @@ Requirements
 
 .. _db-transaction-label:
 
-Database "READ COMMITTED" transaction isolation level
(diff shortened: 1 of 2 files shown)


nickvergessen:
I think it still makes sense to have it mentioned, so admins can be aware which isolation is used and maybe it's not possible to set it on a connection level or something in a future version of a db or something?

ThibautPlg:
I see and agree. 
However it should not be a strict **requirement**. It feels weird to be listed as a mandatory thing when in the end it's handled by the code whether we set it or not. 
Maybe a better phrasing instead of a full removal then?

PR #12801, opened 2025-02-28
feat: add schema for CalendarObjectCreatedEvent webhook

edward-ly:
### ☑️ Resolves

* Follow-up to https://github.com/nextcloud/server/pull/51082

### 🖼️ Screenshots

Screenshot 2025-02-28 at 11-37-20 Webhook Listeners — Nextcloud latest Administration Manual latest documentation


ChristophWurst (approved):

st3iny (commented):
I just realized that we are leaking Sabre classes in this public (OCP) API, for example, in the supported component set member. I know it's a bit late now but we should keep that in mind for the future.

edward-ly:
Do the Sabre classes contain important information that may be useful for webhooks? If not, maybe it's OK to filter out some unneeded properties.

PR #12887, opened 2025-03-18
build: Copy merged OpenAPI spec from server sources

provokateurin:
Possible since https://github.com/nextcloud/server/pull/51674


provokateurin:
With https://github.com/nextcloud/server/pull/51674 there is no build step required anymore, as we can just copy the merged spec from the server checkout directly.

sorbaugh (approved):

PR #12906, opened 2025-03-19
Warning about Share URL change after moving a calendar via occ

b3nks:
Warning about Share URL change after moving a calendar.

### ☑️ Resolves

This should be known before running the command as clients (propably many) need to be adjusted manually.


b3nks:
@rakekniven Hi, when does this version go online?

b3nks:
Ah! It went into the box :). Perfect.

PR #12983, opened 2025-04-02
docs(groupware): Rework occ groupware commands

kesselb:
### ☑️ Resolves

* Improved readability of the section
* Documented list and delete calendar subscriptions https://github.com/nextcloud/server/pull/51136

### 🖼️ Screenshots
| B | A |
| --- | --- |
| Screenshot 2025-04-02 at 18-11-20 Using the occ command — Nextcloud latest Administration Manual latest documentation | Screenshot 2025-04-03 at 14-02-18 Using the occ command — Nextcloud latest Administration Manual latest documentation Screenshot 2025-04-03 at 14-02-37 Using the occ command — Nextcloud latest Administration Manual latest documentation |

### DIFF
--- a/admin_manual/occ_command.rst
+++ b/admin_manual/occ_command.rst
@@ -522,49 +522,74 @@ before. If you want to be notified in that case, set the
 Dav commands
 ------------
 
-A set of commands to create and manage addressbooks and calendars::
+Manage addressbooks and calendars::
 
  dav
-  dav:create-addressbook                 Create a dav addressbook
-  dav:list-addressbooks                  List all addressbooks of a user
-  dav:create-calendar                    Create a dav calendar
-  dav:create-subscription                Create a dav calendar subscription
-  dav:delete-calendar                    Delete a dav calendar
-  dav:fix-missing-caldav-changes         Insert missing calendarchanges rows for existing events
-  dav:list-calendars                     List all calendars of a user
-  dav:move-calendar                      Move a calendar from an user to another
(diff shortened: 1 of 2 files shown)


admin_manual/occ_command.rst
```diff
@@ -577,46 +599,66 @@ This example will delete the birthday calendar of user molly::
...
 
-``dav::move-calendar [name] [sourceuid] [destinationuid]`` allows the admin
-to move a calendar named ``name`` from a user ``sourceuid`` to the user
-``destinationuid``. You can use the force option `-f` to enforce the move if there
-are conflicts with existing shares. The system will also generate a new unique
-calendar name in case there is a conflict over the destination user.
+ sudo -u www-data php occ dav:move-calendar personal dennis sabine
 
-Note that this will change existing share URLs.
+**Misc:**
```
ChristophWurst, line 613: is this too deep to use a heading? https://sublime-and-sphinx-guide.readthedocs.io/en/latest/topics.html#use-headings

ChristophWurst: headings allow deep links, so I'd prefer them over bold text

kesselb: We have one level of nesting left, so we can use " for the headings. That was also my first version yesterday, but I didn't like that those headings were added to the navigation, so I went with strong instead. After giving it a second thought today, I don't think it's too bad to have them in the navigation, and it's also quicker to jump to the right item. Updated the pr accordingly.

ChristophWurst (approved):

PR #13006, opened 2025-04-04
fix: update macOS requirement for desktop clients

nilsding:
Qt6 only supports macOS 12 and newer.  The download websites mention 12+ already.

Also include Alma 8 (RHEL8 / Rocky 8) as a supported distro for our AppImages, as this is the base we use for building the client...

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -105,7 +105,7 @@ We strongly recommend using the latest version of your operating system to get t
 of our clients.
 
 * **Windows** 10+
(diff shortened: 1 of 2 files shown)


Rello:
we should patch the update server then also I guess?

https://github.com/nextcloud/client_updater_server/blob/master/src/Response.php#L130

nilsding:
for the updater server I would let them update until 3.15.x as the last supported version on macOS 11 if possible (that one still used Qt 6.7, which still supported macOS 11 ...)

mgallien (approved):

PR #13045, opened 2025-04-14
fix(admin): Update retention documentation

nickvergessen:
Bildschirmfoto vom 2025-04-14 13-06-40


admin_manual/file_workflows/retention.rst
```diff
@@ -19,6 +19,18 @@ The rule from the example will delete all files tagged with ``Temporary file`` 1
...
+--------
+
+There are 2 options available that can be used to decide when to delete a file:
+
+- **Creation:** Time the file was created on the Nextcloud Server or uploaded to it.
+- **Last modification:** Time when the file was last modified.
+
+Since it often resulted in confusion in the past, the upload time is also considered
+in the modification case. Otherwise users would copy a file into the Nextcloud Sync client directory
+and the file would be deleted during the next background job execution.
```
ShGKme, line 32: This still sounds a bit confusing to me. How it is "considered", why the file would be deleted.

What about something like:

> With the **Last modification** option, the uploading also counts as a modification.
> Thus uploading a long-modified file doesn't retire quickly after the upload.

ShGKme (approved):

ShGKme (approved):

ShGKme:
Should we note when we do and don't have the creation date?

PR #13052, opened 2025-04-17
feat: Make it clearer that master key is used by default

come-nc:
And which section are not pertinent with master key.

TODO:

- [x] Re-order sections so that it makes more sense and user keys specific stuff is at the bottow (less important)
- [x] Move other stuff related to user keys like LDAP section to subsections of user keys
- [x] Mind the labels (not break them)
- [x] Add screenshots

### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

Copie d'écran_20250422_093637
Copie d'écran_20250422_093837


provokateurin (approved):

susnux (approved):
reading seems fine

PR #13071, opened 2025-04-22
Fix spellcheck job on master

come-nc:
This reverts commit 31bea5ae823844bb84b2d002d084b520638b8f91 which broke spellcheck, and adds the same change but without the space so that spellcheck correctly works and checks rst files but not js ones.


come-nc:
Copie d'écran_20250422_173857
It works again :tada:

PR #13073, opened 2025-04-22
update recommended ubuntu LTS version from 22.04 to 24.04 ?

alexanderdd:
I hope this is correct? In the portal we say 24.04 is recommended https://portal.nextcloud.com/article/Installation/Installation---Supported-platforms

### DIFF
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -14,8 +14,8 @@ For best performance, stability and functionality we have documented some recomm
 +------------------+-----------------------------------------------------------------------+
 | Platform         | Options                                                               |
 +==================+=======================================================================+
-| Operating System | - Ubuntu 24.04 LTS                                                    |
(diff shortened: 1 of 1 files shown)


admin_manual/installation/system_requirements.rst
```diff
@@ -14,8 +14,8 @@ For best performance, stability and functionality we have documented some recomm
 +------------------+-----------------------------------------------------------------------+
 | Platform         | Options                                                               |
 +==================+=======================================================================+
-| Operating System | - Ubuntu 24.04 LTS                                                    |
-| (64-bit)         | - **Ubuntu 22.04 LTS** (recommended)                                  |
+| Operating System | - **Ubuntu 24.04 LTS** (recommended)                                                    |
+| (64-bit)         | - Ubuntu 22.04 LTS                                  |
```
susnux, line 18: RST needs the table to be aligned otherwise it does not compile.
you need to adjust the spaces to match with the other rows.

susnux (approved):

PR #13090, opened 2025-04-28
fix(Admin/AppAPI): remove leftover todo message

kyteinsky:
### ☑️ Resolves

* Fix #…


oleksandr-nc (approved):
I thought you left it there on purpose to do it later and not forget about it, but ok..

kyteinsky:
yeah but my internal notes shouldn't be visible to the end reader hehe

PR #13107, opened 2025-05-04
fix(docs): format of occ command

rakekniven:
To pass CI


szaimen:
Thanks @rakekniven but this needs to be fixed here: https://github.com/nextcloud/server/pull/52584. Feel free to approve that one :)

PR #13119, opened 2025-05-12
fix: clarify that group admins cannot add existing users to their group

skjnldsv:
It wasn't clear enough in my opinion.
Group admins can CREATE & DELETE.
But they cannot see all the existing users and add them to their groups.


skjnldsv:
> Who can do this if the Group Admin can't?
> The global administrator?

Yes

PR #13126, opened 2025-05-15
Add hint to pretty URLs: edit config htaccess.ignorefrontcontroller if htaccess SetEnv does not work

rr-it:
### ☑️ Resolves

* Fix https://github.com/nextcloud/documentation/issues/13125

### 🖼️ Screenshots

Please double check the syntax of this PR - especially the internal linking syntax!

### DIFF
--- a/admin_manual/installation/source_installation.rst
+++ b/admin_manual/installation/source_installation.rst
@@ -179,6 +179,10 @@ your .htaccess file::
 
 After each update, these changes are automatically applied to the ``.htaccess``-file.
 
+.. note:: In case the automatically added ``.htaccess`` configuration `SetEnv front_controller_active true` does not work for your environment:
+   Edit ``config/config.php`` and add ``'htaccess.IgnoreFrontController' => true``.
+   See :doc:`../configuration_server/config_sample_php_parameters` for a detailed description.
+
 .. _enabling_ssl_label:
 
 Enabling SSL


rr-it:
@rakekniven So maybe add an anchor to the target document - like:
`.. _htaccess-ignorefrontcontroller:`

And then use a reference in this document
`` :ref:`some description<htaccess-ignorefrontcontroller>` ``

admin_manual/configuration_server/config_sample_php_parameters.rst
```diff
@@ -1270,6 +1270,8 @@ following conditions are met Nextcloud URLs won't contain `index.php`:
 
 Defaults to ``''`` (empty string)
 
+.. _config-php-sample-htaccess-ignorefrontcontroller:
```
kesselb, line 1273: I'm afraid you cannot add a label here because the section is generated, and it will be overwritten.

admin_manual/installation/source_installation.rst
```diff
@@ -179,6 +179,10 @@ your .htaccess file::
 
 After each update, these changes are automatically applied to the ``.htaccess``-file.
 
+.. note:: In case the automatically added ``.htaccess`` configuration `SetEnv front_controller_active true` does not work for your environment:
+   Edit ``config/config.php`` and add ``'htaccess.IgnoreFrontController' => true``.
+   See :ref:`Configuration Parameters: htaccess.IgnoreFrontController<config-php-sample-htaccess-ignorefrontcontroller>` for a detailed description
```
kesselb, line 184:
```suggestion
   See :doc:`../configuration_server/config_sample_php_parameters` for a detailed description.
```

That will work, but not directly jump to the right item. Should be good enough for now.

kesselb (changes requested):

kesselb (approved):

PR #13135, opened 2025-05-18
Fix call on null errors in cache docs code blocks

JonathanTreffler:
### ☑️ Resolves

* Fix: Some code blocks referenced an undefined $this->cache variable instead of the defined $cache

### 🖼️ Screenshots

before:
image

after:
image


ChristophWurst (approved):

ChristophWurst:
Thank you!

PR #13169, opened 2025-05-30
fix: favicon on docs homepage

joshtrichards:


joshtrichards:
It works when adjusting as per this PR in the browser inspector. The favicon immediately appears.

joshtrichards:
It's live today and seems to be working :)

PR #13176, opened 2025-06-02
Add warning about context_chat in the Flows section

julien-nc:
Add warning about context_chat not respecting files_accesscontrol rules.

grafik

### DIFF
--- a/admin_manual/file_workflows/access_control.rst
+++ b/admin_manual/file_workflows/access_control.rst
@@ -8,6 +8,11 @@ rules. If all rules of a group hold true, the group matches the request and
 access is being denied. The rules criteria range from IP address, to user
 groups, collaborative tags and :ref:`some more <available-rules-label>`.
 
+.. note:: In case you are using the :ref:`Context Chat App <ai-app-context_chat>`,
+    please keep in mind, that it is not affected by the File Access Control rules
+    and will respond with indexed information, even when the file is not accessible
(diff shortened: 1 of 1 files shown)


admin_manual/file_workflows/access_control.rst
```diff
@@ -8,6 +8,9 @@ rules. If all rules of a group hold true, the group matches the request and
 access is being denied. The rules criteria range from IP address, to user
 groups, collaborative tags and :ref:`some more <available-rules-label>`.
 
+.. note:: Keep in mind that Context Chat is not affected by the File Access Control rules
+    and will index files for which access is denied by a rule.
```
nickvergessen:
```suggestion
.. note:: In case you are using the :ref:`Context Chat App <ai-app-context_chat>`,
    please keep in mind, that it is not affected by the File Access Control rules
    and will respond with indexed information, even when the file is not accessible
    by the user due to access control rules.
```

Didn't try the label, but it should link to https://github.com/nextcloud/documentation/blob/master/admin_manual/ai/app_context_chat.rst?plain=1#L5

julien-nc: I tried the label and it works.

julien-nc: Change suggestion applied.

nickvergessen (approved):

PR #13210, opened 2025-06-10
fix(ocs): Minor improvements to the OCS OpenAPI documentation

mejo-:


developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -58,7 +58,7 @@ Best practices
 Note that you can find a step-by-step tutorial after this section.
 You can also read the tutorial before reading the best practices.
 
-PREFER to expose your APIs using OCS
+Prefer to expose your APIs using OCS
```
provokateurin: These uppercase words were intentional to be more visible. I would like to stick to that if you have no problems with that.

developer_manual/client_apis/OCS/ocs-openapi.rst
```diff
@@ -544,10 +544,14 @@ It is best to start with helper methods that are used multiple times like the ``
         ...
     }
 
+The return type for `DataResponse` required three arguments: status code, data and headers. See the constructor of `DataResponse` for details.
```
provokateurin: The other inheritors of Response should be explained as well. Can you maybe create a table for that?

developer_manual/client_apis/OCS/ocs-openapi.rst
mejo-: I see. I found them irritating, but that's probably a matter of preferences :wink: Reverted this change.

developer_manual/client_apis/OCS/ocs-openapi.rst
mejo-: I added a table now with some of them and a note where to find that information. Maybe that's better than listing all of them extensively, also because others might be added later on. What do you think?

provokateurin (approved):

PR #13212, opened 2025-06-11
Update source_installation.rst - fix heading hierarchy for snap section

alexanderdd:


alexanderdd:
> The number of characters of "-" and "^" must be the same as that of the heading.

Thank you, didn't know that. I didnt see a quick way to preview rendered output - is there one?

For now, I will adjust the number of ^ (later this week), but wouldn't it make sense to switch to #, ##, ### heading hierarchy? Or is ---- and ^^^^ the only way?

alexanderdd:
@rakekniven length of ^^^^ should be fixed now

PR #13213, opened 2025-06-12
chore: delete unused `user_manual_index.html`

skjnldsv:
Seemed mostly empty


user_manual_index.html
```diff
@@ -1,10 +0,0 @@
-<!DOCTYPE html>
-<html lang="en">
-	<head>
-		<meta charset="utf-8">
-		<title>Nextcloud Documentation</title>
-		<meta http-equiv="refresh" content="0; url=en" />
```
nickvergessen, line 6: its doing an old school redirect from 
https://docs.nextcloud.com/server/latest/user_manual/
to 
https://docs.nextcloud.com/server/latest/user_manual/en/

are we sure it's not the active one?

skjnldsv: Your right, it's being copied to the user_manual root dir.
Its location in this repo root is a bit odd, but I'll adjust :+1:

PR #13238, opened 2025-06-17
feat(devmanual): add section about system config key conventions

ChristophWurst:
### ☑️ Resolves

* Fix topic discussed via chat

### 🖼️ Screenshots

image

### DIFF
--- a/developer_manual/basics/storage/configuration.rst
+++ b/developer_manual/basics/storage/configuration.rst
@@ -73,6 +73,22 @@ System values are saved in the :file:`config/config.php` and allow the app to mo
 
 .. note:: It's also possible to use ``getSystemValueBool``, ``getSystemValueString``, ``getSystemValueInt`` to get type hinted return values.
 
+Naming conventions
+~~~~~~~~~~~~~~~~~~
+
+For consistency there are conventions for config keys:
+
+* System config keys should only contain lower case letters, numbers, `_` and `-`. This ensures that they can be used as environment variables.
+* Keys can be scoped to subsystems like ``<subsystem>_<key>``. This makes it easier to group related configuration.
+
+Here are some examples:
+
+1. ``files_external_allow_create_new_local``
+2. ``filesystem_cache_readonly``
+3. ``log_rotate_size``
+4. ``mail_smtpname``
+5. ``session_lifetime``
+
 App values
 ----------
 


developer_manual/basics/storage/configuration.rst
```diff
@@ -73,6 +73,22 @@ System values are saved in the :file:`config/config.php` and allow the app to mo
 
 .. note:: It's also possible to use ``getSystemValueBool``, ``getSystemValueString``, ``getSystemValueInt`` to get type hinted return values.
 
+Naming conventions
+~~~~~~~~~~~~~~~~~~
+
+For consistency there are conventions for config keys:
+
+* System config keys should only contain lower case letters, numbers, `_` and `-`. This ensures that they can be used as environment variables.
```
nickvergessen:
```suggestion
* System config keys should only contain lower case letters, numbers and `_`. This ensures that they can be used as environment variables.
```

If we make rules, lets use a single separator all the time

susnux: The dash does also not work.
Only use letters numbers and underscore

szaimen: Yeah, hyphens are not supported by normal env variables iirc

szaimen: Indeed

ChristophWurst: Makes sense! I wasn't aware of that

developer_manual/basics/storage/configuration.rst
```diff
@@ -73,6 +73,22 @@ System values are saved in the :file:`config/config.php` and allow the app to mo
 
 .. note:: It's also possible to use ``getSystemValueBool``, ``getSystemValueString``, ``getSystemValueInt`` to get type hinted return values.
 
+Naming conventions
+~~~~~~~~~~~~~~~~~~
+
+For consistency there are conventions for config keys:
+
+* System config keys should only contain lower case letters, numbers and `_`. This ensures that they can be used as environment variables.
```
nickvergessen:
```suggestion
* System config keys should only contain lower case letters, numbers and ``_``. This ensures that they can be used as environment variables.
```
Did you mean to make it code?

szaimen:
> Only thing one could argue would be to use `__` instead of `_` to separate subsystem and key, to not confuse with separating words. E.g

Yes, thought the same but this would mean even more changes to config keys, so I wouls argue it is better to leave it like it is currently with one separator...

kesselb:
> Only thing one could argue would be to use __ instead of _ to separate subsystem and key, to not confuse with separating words

That sounds like a good idea.

ChristophWurst:
> Additionally, I have not worked with any other systems so far that would use double underscore, so it seems a bit unusual to do.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-9.0#naming-of-environment-variables 🌝

susnux:
> Yes, thought the same but this would mean even more changes to config keys, so I wouls argue it is better to leave it like it is currently with one separator...

We would not have to change them if they work. Kind of legacy dept.

My biggest problem with the environment config is that it does not allow scoped configuration, where we use arrays.
For example:
 ```php
'redis.cluster' => [
	'seeds' => [
		'localhost:7000',
		'localhost:7001',
	],
	'timeout' => 0.0,
],
```

**Comment by nickvergessen** [comment]
Well for me the main question is:
Should all values be env-override-able or not? I don't think the answer is a clear yes, but I also lack the idea when/why an instance would use an environment overwrite (is it temporary?) vs touching the config file? Something like the cache, DB connection, etc. sound like a "no" to me. I can see how one would try to do it with the DB password and other secrets, so they are not stored in the filesystem, etc. but then we can also simply check if we can "flatten" out those specific values, so they can be specifically overwritten instead of having to translate the full array to ENV?

**Comment by nickvergessen** [comment]
```
NC_REDIS_CLUSTER__SEEDS__0=localhost:7000
NC_REDIS_CLUSTER__SEEDS__1=localhost:7001
NC_REDIS_CLUSTER__TIMEOUT=0.0
```

The bigger question with arrays is how replacing works. Let's say file has:
```php
        "log.condition": {
            "apps": {
                "1": "admin_audit",
                "50": "calendar-appointments",
                "51": "appointments",
                "52": "calendar",
                "200": "spreed-bfp",
                "501": "imaginary"
            }
        },
```

would:
```env
NC_log_condition__apps__0=perf_debug
```
result in:
```php
        "log.condition": {
            "apps": {
                "0": "perf_debug"
            }
        },
```
or:
```php
        "log.condition": {
            "apps": {
                "0": "perf_debug",
                "1": "admin_audit",
                "50": "calendar-appointments",
                "51": "appointments",
                "52": "calendar",
                "200": "spreed-bfp",
                "501": "imaginary"
            }
        },
```

ChristophWurst:
Probably the second one, because of this:

```
php > $config = ["1" => "foo", "2" => "bar"];
php > $config[0] = "replaced";
php > print_r($config);
Array
(
    [1] => foo
    [2] => bar
    [0] => replaced
)
```

nickvergessen:
Build is failing:
>  /home/runner/actions-runner/_work/documentation/documentation/admin_manual/configuration_server/config_sample_php_parameters.rst:4006: ERROR: Unexpected indentation. [docutils]
/home/runner/actions-runner/_work/documentation/documentation/admin_manual/configuration_server/config_sample_php_parameters.rst:4007: WARNING: Block quote ends without a blank line; unexpected unindent. [docutils]

ChristophWurst:
To which extent? Some of the ideas won't work right now and need changes in server

PR #13387, opened 2025-07-07
rename title from "Encryption configuration" to "Server-Side Encrypti…

alexanderdd:
…on configuration"

we also have e2ee, which is not covered here, so I feel the title needs to be more specific

### DIFF
--- a/admin_manual/configuration_files/encryption_configuration.rst
+++ b/admin_manual/configuration_files/encryption_configuration.rst
@@ -1,6 +1,6 @@
-========================
-Encryption configuration
-========================
+====================================
+Server-Side Encryption configuration
+====================================
 
 The primary purpose of the Nextcloud server-side encryption is to protect users' 
 files on remote storage, such as Dropbox and Google Drive, and to do it easily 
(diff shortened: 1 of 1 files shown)


alexanderdd:
Yes IMO we should also rename the titles of the other pages, and all file names.

Do you agree that this makes sense?

alexanderdd:
@rakekniven I don't know how to include those changes here in this PR, it was easier/faster for me to create separate PRs:
https://github.com/nextcloud/documentation/pull/13504
https://github.com/nextcloud/documentation/pull/13505

Can you approve those PRs, and this PR here?

AFAIK the Menu entries are generated from the titles of the documents (not the *.rst file names), so those should be updated automatically to "Server-Side encryption".

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -1,6 +1,6 @@
-========================
-Encryption configuration
-========================
+====================================
+Server-Side Encryption configuration
+====================================
```
rakekniven, line 3: Server-Side Encryption details -> Server-side encryption details

rakekniven (changes requested):

rakekniven (approved):

PR #13404, opened 2025-07-11
feat(user_manual): Make translation versions more visible 

### LINKED ISSUE #8310: Make translations more visible for user manual
I just discovered by accident, that the user manual is translated in other languages.

grafik

Who would guess that translations are hidden behind there? It's probably even more important than being able to switch between versions.

nickvergessen:
…dropdown

### ☑️ Resolves

* Fix #8310 
* Added the language on the handle that opens the dropdown
### 🖼️ Screenshots Before

Closed | Opened
---|---
grafik | grafik

### 🖼️ Screenshots After

Closed | Opened
---|---
Bildschirmfoto vom 2025-07-11 13-48-55 | grafik

### DIFF
--- a/admin_manual/_templates/versions.html
+++ b/admin_manual/_templates/versions.html
@@ -2,15 +2,22 @@
 {# Add rst-badge after rst-versions for small badge style. #}
   <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="versions">
     <span class="rst-current-version" data-toggle="rst-current-version">
-      <span class="fa fa-book"> Read the Docs</span>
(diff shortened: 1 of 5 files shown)


skjnldsv (approved):

nickvergessen:
grafik

Should we add some generic flags then similar to ^
Or only the 🌐 but it's blue like all the link text and therefor barely visible in all cases?
I think it would make sense to have something with color to drag a bit of attention to a language selection

nickvergessen:
Done, let me know what you think

PR #13421, opened 2025-07-15
Document how to build translated documentations

### LINKED ISSUE #8280: How to build translated versions?
Thanks a lot for this awesome documentation for Nextcloud. By following the documentation in README.rst I was able to successfully build both the HTML and the PDF version locally.

However, I wanted to build a PDF of the german version to provide it in a translated version of the Nextcloud skeleton folder. I can see that the repository contains translated strings in user_manual/locale, but I couldn't find any documentation on how to build a translated version of the manual from this.

Maybe some instructions for building translated versions could also be added to the README.rst? Or is that documented somewhere else? Then we could maybe point to that external documentation? Thanks a lot in advance!

nickvergessen:
### ☑️ Resolves

* Fix #8280

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -143,6 +143,17 @@ Using venv
 5. Now you can use ``make ...`` to build all the stuff - for example ``make html`` to build the HTML flavor of all manuals
    The build assets will be put into the individual documentation subdirectories like ``developer_manual/_build/html/com``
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
Looks good but didn't test

PR #13086, opened 2025-04-27
explain v31 db row format warning

meonkeys:
### ☑️ Resolves

There is no associated issue, just this PR.

### 🖼️ Screenshots

Screenshot from 2025-04-27 14-28-52

PR #13449, opened 2025-07-25
fix(admin_manual/release_notes): correct ROW FORMAT migration in 31

ShGKme:
### ☑️ Resolves

* For: https://github.com/nextcloud/documentation/pull/13086
* Ref: https://github.com/nextcloud/server/pull/48547/
* The `occ db:convert-mysql-charset` command changes the **charset**, not the `ROW_FORMAT`

### 🖼️ Screenshots

image


susnux:
> The `occ db:convert-mysql-charset` command changes the **charset**, not the `ROW_FORMAT`

It does both (if the table is in wrong charset): 
https://github.com/nextcloud/server/blob/38768d1f8d8a0697c71580fa5821984bb10db7fe/lib/private/Repair/Collation.php#L62

susnux (approved):

kesselb (approved):

PR #13470, opened 2025-07-30
chore(i18n): Add missing dot (again)

rakekniven:
Reported at Transifex.


hamza221:
Do you know how far should we backport ? 
if not I can check

PR #13493, opened 2025-08-07
docs(user-manual): add a note about translation detection

hamza221:
### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/issues/11068

### 🖼️ Screenshots
image

### DIFF
--- a/user_manual/groupware/mail.rst
+++ b/user_manual/groupware/mail.rst
@@ -357,6 +357,8 @@ You are able to translate messages to your configured languages similarly to Tal
 
 .. note:: Please note that translation features have to be enabled on the server
 
+.. note:: Since version 5.3, if LLM is enabled by admin, translations will be suggested
+
 Thread summary
 --------------
 


hamza221:
I'm not sure why it's arabic, I thin it picked the first language alphabetically while building

user_manual/groupware/mail.rst
```diff
@@ -357,6 +357,8 @@ You are able to translate messages to your configured languages similarly to Tal
 
 .. note:: Please note that translation features have to be enabled on the server
 
+.. note:: Since version 5.3, if LLM is enabled by admin, translations will be suggested
```
kesselb, line 360: There's also `.. versionchanged:: 5.3 blub blub` for such cases.

Here's an example https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html#ocp-preview-beforepreviewfetchedevent. 

Not requesting it as a change, because it depends on the use case I guess.

hamza221: I'm aware of that, but I think in this case it makes sense to add it to the note as we're advertising it as a sub-feature at least documentation wise, and the proper feature (translation) has a different version that's already mentioned above .

PR #13511, opened 2025-08-14
fix command (replace "=" with " ")

alexanderdd:
reported by a customer, confirmed by Björn


schiessle (approved):
I can confirm that I also use it in my local scripts always without the "="

PR #13529, opened 2025-08-21
chore(AI/ContextChat): update inital indexing docs

kyteinsky:
### ☑️ Resolves

* Fix #…

### DIFF
--- a/admin_manual/ai/app_context_chat.rst
+++ b/admin_manual/ai/app_context_chat.rst
@@ -81,13 +81,13 @@ Auto-indexing
 
 The indexing jobs are set up to run during the Nextcloud instance's maintenance window (typically during the night) only. If you have not set a maintenance window, indexing will run 24/7.
 
-You can set up a background job worker explicitly for Context Chat to avoid slowing down normal background job operation on larger instances.
+| You can set up a separate cron job to run every 30 minutes for Context Chat to avoid slowing down normal background job operation on larger instances.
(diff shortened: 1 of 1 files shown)


admin_manual/ai/app_context_chat.rst
```diff
@@ -81,13 +81,13 @@ Auto-indexing
 
 The indexing jobs are set up to run during the Nextcloud instance's maintenance window (typically during the night) only. If you have not set a maintenance window, indexing will run 24/7.
 
-You can set up a background job worker explicitly for Context Chat to avoid slowing down normal background job operation on larger instances.
+| You can set up a separate cron job to run every 30 minutes for Context Chat to avoid slowing down normal background job operation on larger instances.
+| The following command can bypass the maintenance window so it can either be set to run during the day even with a maintenance window set, or it can be set to run during the weekends 24/7 to speed up the indexing process.
```
oleksandr-nc, line 85: will it be good idea to include at least one example how to do that? or this is not needed?

kyteinsky: not sure if that belongs here though, admins might know how to setup a cron job or can look it up. It would be simiar to how the `cron.php` is already run on their system, just at 30 minute intervals.
maybe we can point to this section for possible ways to do it?
https://docs.nextcloud.com/server/latest/admin_manual/ai/overview.html#improve-ai-task-pickup-speed

PR #13647, opened 2025-09-10
feat: Automated EoL warning banner for unsupported releases

### LINKED ISSUE #9634: Add canonical tags in pages linking to 'stable'; and add a warning for users.
* [ ] I'd love to add a link to the 'stable' variant of each page on all our pages, to point Google and other search engines there. https://moz.com/learn/seo/canonicalization

* [ ] I'd also like to see a warning for users, in particular for no longer supported versions, that this version of Nc is no longer supported. We have this for 12, see https://docs.nextcloud.com/server/12/admin_manual/ - but it seems it's not there for newer docs..

joshtrichards:
* Adds a warning banner for outdated documentation
* Visible regardless of landing page
* Applies to all types (Admin + Dev + User).
* Automated (uses existing `version_start` variable to determine currently last supported release)
* Does not require theme changes (the RTD theme doesn't support announcements)

### ☑️ Resolves

* Fix #9634

Should be backported as far back as possible (well to v13+) so that all published (ancient) documentation versions pick this up[^1].

[^1]: (may also need to trigger new deploys for those; not sure if they still get picked up automatically).

### 🖼️ Screenshots

image

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -92,8 +92,8 @@ if (version.isdigit() and version < version_start):
     rst_prolog = """.. danger::
         **OUTDATED DOCUMENTATION**
     
-            *You are viewing documentation for a retired version of Nextcloud.
-            Do not follow these instructions for current releases.*
+        *You are viewing documentation for a retired version of Nextcloud.
+        Do not follow these instructions for current releases.*
 
         **To ensure you have the most reliable and up-to-date guidance,
         please visit the** `Nextcloud Documentation homepage


conf.py
```diff
@@ -87,6 +87,19 @@ def generateVersionsDocs(current_docs):
 edit_on_github_project = 'nextcloud/documentation'
 edit_on_github_branch = 'master'
 
+# Automatically add EoL warning banner to docs for unsupported releases
+if (version.isdigit() and version < version_start):
+    rst_prolog = """.. danger::
+        **OUTDATED DOCUMENTATION**
+    
+            *You are viewing documentation for a retired version of Nextcloud software.
```
kesselb:
```suggestion
            *You are viewing documentation for a retired version of Nextcloud.
```

conf.py
```diff
@@ -87,6 +87,19 @@ def generateVersionsDocs(current_docs):
...
+            Do not follow these instructions for current releases.*
+
+        **To ensure you have the most reliable and up-to-date guidance,
+        please visit the** `official Nextcloud Documentation library
```
kesselb:
```suggestion
        please visit the** `Nextcloud Documentation
```

joshtrichards:
```suggestion
        please visit the** `Nextcloud Documentation homepage
```

joshtrichards:
> I find the styling a bit wild. Have you done it like that purposely to catch the readers attention? 

Yeah pretty much. That and the options are somewhat limited with the way this is being done (unless we think it's worth the effort to really polish up much further).

> I would prefer to not indent the "retired" paragraph.

Fair. No strong feelings either way. I'll adjust.

joshtrichards:
TLDR: Older branches don't know they're old. :-P

I overlooked that the "currently supported" version range variables are within each branch's respective `conf.py`. We don't bump them in our older branches so they are always wrong (outside of `master`). They're not useful as-is for querying for authoritative release status. 

Two thoughts:

* I could update those variables in all old branches; but that then becomes yet another manual process (since they'd be out of state at every EoL/stable bump cycle)
* Or I could change how we populate those variable values by automating it. :) The updater_server now provides per major EoL/stable data via its web API. I'd backport that change too, but only once and then we'd never have to do it again.
  - Bonus: Automates "supported version" range tracking in `master` too (yet another manual task that someone always needs to do at every major release cycle). Incidentally this could also eventually be extended to automating/syncing some other currently manual tasks: `/latest` and `/stable` URL symbolic link adjustments and even generating the current docs homepage with all the appropriate versions.

PR #13666, opened 2025-09-13
Update note on Server URL format

theking2:
The server url string might need a query string.


user_manual/groupware/sync_thunderbird.rst
```diff
@@ -36,6 +36,8 @@ Calendars
 #. Type your **Username** and **Location** (Server URL), then click on **Find Calendars**.
 #. Choose which calendars you want to add and click **Subscribe**
 
+> **_NOTE:_**  The Server URL does not required a scheme (`https://`) but might need a query string like `/remote.php/dav`
```
susnux: Thats not the query part of the URL but the path. The query part is everything behind the `?`.
I think this should rather point out that you should copy the URL from the calendar app (there are "settings" -> "primary DAV address" which is the URL you need here)

theking2: thanks, yeah you are right about the first part. the second part `settings / primary DAV address` does not exit on my nextcloud based system and calendar app.

susnux: On the calendar app, on the bottom left:

Screenshot 2025-09-13 at 18-18-52

Then scroll down and there it is:

user_manual/groupware/sync_thunderbird.rst
```diff
@@ -36,6 +36,9 @@ Calendars
 #. Type your **Username** and **Location** (Server URL), then click on **Find Calendars**.
 #. Choose which calendars you want to add and click **Subscribe**
 
+
+.. note:: The Server URL does not require a scheme (`https://`) but might need a path like `/remote.php/dav`
```
joshtrichards, line 40: Do you have any warnings about `/.well-known` URLs in your setup checks (in Nextcloud that is under *Administration settings->Overview*)?

AFAIK recent Thunderbird versions query the RFC `/.well-known/{caldav, carddav}` URLs automatically.

skjnldsv:
@theking2 thanks for your PR.
Have you been able to check the two pending comments by @susnux and @joshtrichards ? :blush:

rakekniven:
Hello @theking2 ,

you recently merged branch master into your branch.

What would be even more important for PR is that you reply to the comments from @joshtrichards and @joshtrichards.

rakekniven:
Will close it next week if no reply comes in.

user_manual/groupware/sync_thunderbird.rst
kingma-sbw: Excellent: That would fit right in the documentation

user_manual/groupware/sync_thunderbird.rst
theking2: No warnings here

user_manual/groupware/sync_thunderbird.rst
theking2: This probably is the royal way to add Nextcloud caldav/calcard support to Thunderbird. But it is nowhere mentioned in the [documentation](https://docs.nextcloud.com/server/latest/user_manual/en/groupware/sync_thunderbird.html)

theking2: Actually this is _not_ a technical issue but a _documentation_ issue. The documentation flies over how to establish the URL to use. It merely states "2. In the next window, type your Username and Location (Server URL)." but "Server URL" is understandibly interpreted as `https://cloud.example.com/`

theking2:
I added a note to @joshtrichards . The questions are technical nature and not documentation. Two things imo

PR #13683, opened 2025-09-18
Remove wasm MIME type from nginx-subdir.conf.sample

kinafu:
On my nginx/1.29.1 the wasm MIME-type is included in /etc/nginx/mime.types

### ☑️ Resolves

the error 
```
web-1    | 2025/09/18 19:32:07 [warn] 9#9: duplicate extension "wasm", content type: "application/wasm", previous content type: "application/wasm" in /etc/nginx/nginx.conf:16
web-1    | nginx: [warn] duplicate extension "wasm", content type: "application/wasm", previous content type: "application/wasm" in /etc/nginx/nginx.conf:16
```


admin_manual/installation/nginx-subdir.conf.sample
```diff
@@ -52,7 +52,6 @@ server {
     include mime.types;
     types {
         text/javascript mjs;
```
joshtrichards, line 57:
```suggestion
        text/javascript mjs;
        # uncomment below for Nginx <= 1.21.0 
        #application/wasm wasm; 
```

joshtrichards: Since we have to support a wide variation of Nginx versions (including EoL ones that are still supported within non-EoS distributions), albeit less so as time passes. 

Refs:

https://github.com/nginx/nginx/commit/130a3ec5010227ca93498a1eb3a182062daeb349

https://sources.debian.org/src/nginx/

joshtrichards (commented):
Thanks for the nudge. It doesn't hurt anything since dups are just a warning, but I agree it should be tidied up.

skjnldsv (approved):

skjnldsv:
@joshtrichards what do you think, we can merge it now?
We're talking about an nginx release from 5 years ago :)

EDIT: I see bullseye is still on 1.18, while the fix is 1.21. I think your comment approach is best. I added it

admin_manual/installation/nginx-subdir.conf.sample
```diff
@@ -55,7 +55,6 @@ server {
...
```
skjnldsv, line 57:
```suggestion
        text/javascript mjs;
        # uncomment below for Nginx <= 1.21.0 
        # application/wasm wasm;
```

PR #13694, opened 2025-09-22
feat(AI): add config details in translate2 admin docs

kyteinsky:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

image


marcelklehr:
I think we shouldn't document this. If admins ever should need to change any value in config.json, we should make a UI for it.

PR #13716, opened 2025-09-26
feat: Add removal of OC.SystemTags JS API

CarlSchwan:
### 🖼️ Screenshots

image

For https://github.com/nextcloud/server/pull/55320


nickvergessen (approved):
I think it also makes sense to add a deprecation in the code and announce it in the 32 docs

PR #13749, opened 2025-10-02
chore: Update iOS minimum required versions.

Ivansss:


nickvergessen:
Red CI fixed in https://github.com/nextcloud/documentation/pull/13750

PR #13754, opened 2025-10-05
refactor(admin): Rewrite Server-Side Encryption chapter

joshtrichards:
Updated the server-side encryption documentation.

- Added context regarding "encryption" to help reader understand SSE in the context of other encryption solutions (both within Nextcloud and provided by their OS/etc)
- Improved clarity throughout
- Reorganized and streamlined
- Additional reference links
- Fixed a few outdated comments/notes
- Expanded/clarified key management modes and encryption methods a bit

### ☑️ Resolves

Follow-up items:
- [ ] Review https://github.com/nextcloud/documentation/labels/feature%3A%20encryption%20%28server-side%29 tagged issues

### 🖼️ Screenshots

(works better if you open the screenshot image in a new tab to view it full-size)
image

### DIFF
--- a/admin_manual/configuration_files/encryption_configuration.rst
+++ b/admin_manual/configuration_files/encryption_configuration.rst
@@ -1,358 +1,330 @@
-====================================
-Server-side encryption configuration
-====================================
+======================
+Server-side Encryption
+======================
 
-The primary purpose of the Nextcloud server-side encryption is to protect users' 
-files on remote storage, such as Dropbox and Google Drive, and to do it easily 
-and seamlessly from within Nextcloud.
+Overview
+--------
 
-Server-side encryption separates encryption of local and remote storage. 
-This allows you to encrypt remote storage, such as Dropbox and 
-Google, without having to also encrypt your home storage on your Nextcloud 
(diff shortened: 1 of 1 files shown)


Toorero:
It would be cool if you could mention that `occ:fix-encrypted-version` will (most likely) lead to data loss on the corrupted files since it will "just" roll-back to the most recent version that has a valid signature *discarding* more recent versions with signature errors.

alexanderdd (commented):
Thank you for working on this. Sorry, I don't have time to look at this at the moment.

schiessle:
@joshtrichards thanks for working on it. One suggestion from my side. We should mention that user keys have some draw backs which can easily irritate users. 

A example I use all the time: Think about a group share, the moment you share a file with a group Nextcloud will take the public keys of all group members and encrypt the file-key with it. With large groups this can have a (huge) performance impact. If a admin adds later additional users to the group they will see the file shared with the group but they will not be able to decrypt it because they where not part of the group when the file-key was encrypted. This can lead to a lot of confusion and 1st level support request like "i can't open a file shared with me", "nextcloud is broken", etc.

PR #13765, opened 2025-10-07
Create generator-generic-ossf-slsa3-publish.yml

mauricio375:
### ☑️ Resolves

* Fix #…


kesselb:
Hi @mauricio375, would you mind explaining what that is about?

susnux:
Please reopen if still actual and use case is clarified

PR #13789, opened 2025-10-11
fix: drop hard-coded Activity app config

joshtrichards:
### ☑️ Resolves

Dropping the hard-coded bits, to swap with the ones being added to the config sample itself in the ``server``  repo. See nextcloud/server#55706 for context. Not dependent on that PR directly so merge at-will.

### DIFF
--- a/admin_manual/configuration_server/activity_configuration.rst
+++ b/admin_manual/configuration_server/activity_configuration.rst
@@ -29,8 +29,7 @@ Furthermore it is recommended to configure the background job ``Webcron`` or
 ``Cron`` as described in :doc:`background_jobs_configuration`.
 
 There is also a configuration option ``activity_expire_days`` available in your
-``config.php`` (See :ref:`label-activity-app-config`) which allows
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/config_sample_php_parameters.rst
```diff
@@ -4223,27 +4223,6 @@ Defaults to ``true``
...
-	'activity_expire_days' => 365,
-
-Every day a cron job is ran, which deletes all activities for all users
-which are older then the number of days that is set for ``activity_expire_days``
-
-::
-
-	'activity_use_cached_mountpoints' => false,
-
-Before enabling this, read the warning in :ref:`label-activities-groupfolders`
```
provokateurin, line 4245: The warning should probably be removed as well, as it's now in the config.sample.php?

joshtrichards: Perhaps. Maybe revisit whenever the Activity chapter gets reworked at some point down the road?

PR #13795, opened 2025-10-13
fix: add note about change in behaviour of the job

yemkareems:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

image
Newly added paragraph

### DIFF
--- a/admin_manual/configuration_files/file_versioning.rst
+++ b/admin_manual/configuration_files/file_versioning.rst
@@ -24,7 +24,7 @@ oldest file versions until it meets the disk space limit again.
 Nextcloud manages file versions using a combination of on-save pruning and scheduled cleanup. This ensures that versions are retained while respecting storage quotas.
 
 During Version Creation
-----------------------
+-----------------------
 
 Nextcloud automatically creates new file versions whenever a file is modified, allowing users to restore previous states when needed. After each new version is stored, the system automatically checks storage limits and retention rules. Versions are filtered according to the above pattern to keep representative versions and remove redundant ones. If the user’s quota is exceeded, auto-expiry is triggered.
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_files/file_versioning.rst
```diff
@@ -54,4 +54,22 @@ Deactivate background job: ``occ config:app:set --value=no files_versions backgr
...
+The ``versions_retention_obligation`` setting behaves slightly differently
+depending on whether version expiration is triggered by an *explicit job run*
+(for example, via ``occ versions:expire`` or a scheduled background job) or by
+*automatic expiry*.
+
+Explicit job-based expiration is *strict* and adheres closely to the configured
+``versions_retention_obligation``.
+
+Auto-expiry follows the same retention principles but may apply them more flexibly, allowing minor deviations when
+necessary to maintain adequate free storage space.
```
artonge: This needs to be more explicit on what kind of deviations. This is deterministic, so we should be able to describe it.

yemkareems: Added the paragraph above note and additional options corrected as per config

admin_manual/configuration_files/file_versioning.rst
```diff
@@ -21,6 +21,20 @@ The Versions app never uses more than 50% of the user's currently available
...
```
artonge: I don't think that is true, or if it is, then we need to be explicit on what are the deviations. Admins need to know what's happening with the files.

yemkareems: Is it fine to change it to "Auto-expiry follows the same retention principles as standard version retention. However, when the available storage space becomes critically low, Nextcloud may delete older versions earlier than their configured retention period to free up space."

PR #13804, opened 2025-10-15
feat: Mark IQueryBuilder::execute as removed

CarlSchwan:
Ref https://github.com/nextcloud/server/pull/55720

image


ChristophWurst:
Given the impact, I think this deserves more detail. Explain how to migrate with short examples. Explain the differences in error handling.

PR #13810, opened 2025-10-21
Add note about davfs2 group creation

Koruudo:
I just went through setting up davfs2 on Arch and it did not create the davfs2 group for me. I also had to edit the config to make it use the group, once I created it, because it was defaulting to network. I was getting the following error when trying to mount it:

user <username> must be member of group network

Please feel free to adjust my comment if need be. :)

### 🖼️ Screenshots

Screenshot_20251021_172219

### DIFF
--- a/user_manual/files/access_webdav.rst
+++ b/user_manual/files/access_webdav.rst
@@ -146,6 +146,9 @@ automatically every time you log in to your Linux computer.
 
     usermod -aG davfs2 <username>
 
+.. note::
+	**If the davfs2 group doesn't exist after installing the package, you may need to create it yourself and, possibly, adjust the davfs config file to use the group after you've created it.
+
 #. Then create a ``nextcloud`` directory in your home directory for the
    mount point, and ``.davfs2/`` for your personal configuration file::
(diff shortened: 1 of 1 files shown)


user_manual/files/access_webdav.rst
```diff
@@ -142,6 +142,9 @@ automatically every time you log in to your Linux computer.
 
     usermod -aG davfs2 <username>
 
+.. note::
+	**If the davfs2 group doesn't exist after installing the package, you may need to create it yourself and, possibly, adjust the davfs config file to use the group after you've created it.
```
tflidd: A note box and bold is perhaps a bit much.

skjnldsv:
```suggestion
	If the davfs2 group doesn't exist after installing the package, you may need to create it yourself and, possibly, adjust the davfs config file to use the group after you've created it.
```

skjnldsv (approved):

skjnldsv:
Took long enough! Sorry and thank you @Koruudo !!

PR #13831, opened 2025-11-02
Improved info on Send email button

edwin-hendriks:
Improved explanation on errors that will happen if you have not configured a personal email address.

### ☑️ Resolves

* Prevent people from making the same error

### 🖼️ Screenshots

n.a.


skjnldsv:
I agree with tflidd's comment :)
I think this means the server UX should be improved instead

PR #13850, opened 2025-11-10
Add Accept header to curl example in ocs-api-overview.rst

sunetfreitag:
The return type of ocs calls can easily be changed by adding an Accept header to the request. Since some ocs calls return xml as their default, while others return json, it might be beneficial to the reader to state this early in the documentation.

### DIFF
--- a/developer_manual/client_apis/OCS/ocs-api-overview.rst
+++ b/developer_manual/client_apis/OCS/ocs-api-overview.rst
@@ -37,6 +37,13 @@ For example: you can perform a :code:`GET` request to get information about a us
 
     curl -u username:password -X GET 'https://cloud.example.com/ocs/v1.php/...' -H "OCS-APIRequest: true"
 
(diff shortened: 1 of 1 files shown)


tflidd (commented):
Yes, might be helpful if people prefer json. Github marks a typo, can you fix this?

developer_manual/client_apis/OCS/ocs-api-overview.rst
```diff
@@ -37,6 +37,13 @@ For example: you can perform a :code:`GET` request to get information about a us
 
     curl -u username:password -X GET 'https://cloud.example.com/ocs/v1.php/...' -H "OCS-APIRequest: true"
 
+You can change the response type of your request by adding an Accept header. For example, if you prefer JSON respones:
```
skjnldsv:
```suggestion
You can change the response type of your request by adding an Accept header. For example, if you prefer JSON responses:
```

skjnldsv (approved):

PR #13853, opened 2025-11-11
fix: correct/improve webhook listener event descriptions and examples

joshtrichards:
### ☑️ Resolves

* Fix nextcloud/server#56371
* Fix https://github.com/nextcloud/server/pull/53648#pullrequestreview-2953626347

### Summary

- **Clarified missing `id` fields:**
  - Added notes on when the `id` is missing from node event payloads, and recommended using the corresponding `Before*` event if the `id` is required.
- **Clearer event grouping and applies-to lists:**
  - Grouped Node and Calendar Object events by payload format and specified which events use each type.
- **Added extensibility notice:**
  - Clarified that installed apps can register additional webhook-compatible events, so available events may vary.
- **Context for optional app events:**
  - Noted that Forms and Tables events apply only if those apps are installed.
- **Miscellaneous, Formatting and Sphinx improvements:**
  - Made the JSON output prettier
  - Dropped duplicate paragraph
  - Improved blank line usage and literal formatting for better documentation rendering and consistency.
- Mention `occ` command

Still some follow-up items but this is enough for this particular PR. Ideas for future PR(s):

- OCS
  - registration
  - auth
  - ephemeral tokens
- AppAPI / ExApp integration/support
- MapperEvent & IWebHookCompatibleEvent
- `curl` / OCS API request/response example

### 🖼️ Screenshots

image

### DIFF
--- a/admin_manual/webhook_listeners/index.rst
+++ b/admin_manual/webhook_listeners/index.rst
@@ -47,13 +47,46 @@ You can use the OCS API to add webhooks for specific events. See:
 Note: When authenticating with the OCS API to register webhooks, the account you
 use must have administrator rights or delegated administrator rights.
 
+Listing registered webhooks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can list all currently registered webhook listeners from the command line:
+
+.. code-block:: bash
+
+   occ webhook_listeners:list
+
+By default this prints a table. Use the ``--output`` option to change the format.
+Each row contains the full configuration of a registered webhook
+
 Filters
 ~~~~~~~
 
-When registering a webhook listener, you can specify a filter parameter. The value of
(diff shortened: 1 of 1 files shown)


admin_manual/webhook_listeners/index.rst
```diff
@@ -172,571 +165,302 @@ It is recommended to restart this worker at least once a day to make sure code c
...
+
+.. code-block:: json
+
+   {
+     "event": "NodeCreatedEvent",
+     "node": {
+       "id": 437,
+       "path": "/admin/files/test-webhook.txt"
+     }
+   }
```
marcelklehr, line 278: Don't these have `time` and `user` variables as well?

joshtrichards: Indeed. Oops. Fixed all the payloads. Thanks and apologies for the delay.

PR #13885, opened 2025-11-24
feat(dev): Add docs for UserFacingProcessingException

marcelklehr:
### ☑️ Resolves

* Docs for https://github.com/nextcloud/server/pull/55713

### 🖼️ Screenshots
image

image

### DIFF
--- a/developer_manual/digging_deeper/task_processing.rst
+++ b/developer_manual/digging_deeper/task_processing.rst
@@ -242,6 +242,11 @@ The task class objects have the following methods available:
  * ``getWebhookUri()`` This returns the webhook URI that will be notified when the task execution has ended
  * ``getWebhookMethod()`` This returns the HTTP method that will be used for the webhook when the task execution has ended
 
+
+.. versionadded:: 33.0.0
+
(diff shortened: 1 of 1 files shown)


developer_manual/digging_deeper/task_processing.rst
```diff
@@ -242,6 +242,11 @@ The task class objects have the following methods available:
  * ``getWebhookUri()`` This returns the webhook URI that will be notified when the task execution has ended
  * ``getWebhookMethod()`` This returns the HTTP method that will be used for the webhook when the task execution has ended
 
+
+.. versionadded:: 33.0.0
+
+ * ``getUserFacingErrorMessage()`` This returns any error message that is meant to be displayed to the user, even if a task has failed, this is not guaranteed to be set.
```
kyteinsky, line 248: maybe it can be clarified here what is returned when it's not set

marcelklehr: We also don't do that for the other methods :thinking:

kyteinsky: it might be better to do for all of them but the function signatures/docs can be seen in the IDE too

kyteinsky (approved):

PR #13893, opened 2025-11-26
fix(developer): `nextcloud/max-version` is required since Nextcloud 11

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud/documentation/issues/13892


joshtrichards:
I wonder if we should simply remove much of this duplicated content.

The [app store's own documentation](https://nextcloudappstore.readthedocs.io/en/latest/developer.html#app-metadata), which we briefly link to at the top of this page, already covers most of this material. The documentation needs to be fixed there as well; however, perhaps we should drop most of this duplicated content here instead, leaving the app store doc page to serve as the sole source of truth.

come-nc (approved):

PR #13895, opened 2025-11-28
docs(admin): clarify LDAP user cleanup and manual deletion process

### LINKED ISSUE #13868: Clarification for LDAP user cleanup
### Before You Submit

- [x] I have checked for existing issues and pull requests related to my suggestion.
- [x] I have posted on the [Nextcloud Community Help Forum](https://help.nextcloud.com) if my suggestion is a troubleshooting scenario or I'm unsure about the solution.

### Type of Suggestion

Suggesting an improvement

### Relevant Documentation Links

https://docs.nextcloud.com/server/stable/admin_manual/configuration_user/user_auth_ldap_cleanup.html

### Summary

I would like clarification on what it means when a user is marked as deleted 

### Details

After reading the documentation of the LDAP user cleanup, I'm unsure what exactly the LDAP cleanup is doing when it detects that a user no longer exists in the LDAP. I'll try to give my understanding of the process

1. A user is found to be unavailable on LDAP
2. The user is marked as `deleted` in the `oc_preferences` table. The user now shows up with `occ ldap:show-remnants`
3. The cleanup deletes the items listed on the documentation page (group assignments, user preferences, home folder, `oc_storages` entry.

What remains of the user on the Nextcloud instance? The user account still exists and has to be manually deleted, but what other data of the user is still there while it is in its "marked deleted" state?

I think the page would benefit from a rewrite to clarifiy these things.

### Additional Notes

_No response_

joshtrichards:
### ☑️ Resolves

* Fix #13868

* Clarified the behavior of LDAP user cleanup when users are no longer present in LDAP.
* Explained what it means for a user to be "marked as deleted"
* Reorganized existing content (and slightly edited it to facilitate the reorganization) for overall clarity

### DIFF
--- a/admin_manual/configuration_user/user_auth_ldap_cleanup.rst
+++ b/admin_manual/configuration_user/user_auth_ldap_cleanup.rst
@@ -70,7 +70,7 @@ Following flags can be specified additionally:
 * ``--json``: instead of a table, the output is json-encoded. **This makes it easy to process the data programmatically if desired**.
 
 Manually Deleting User Accounts
-------------------------------
+-------------------------------
 
 After reviewing the users that have been marked for deletion, you can manually remove an account and all its data using:
 
--- a/admin_manual/configuration_user/user_auth_ldap_cleanup.rst
+++ b/admin_manual/configuration_user/user_auth_ldap_cleanup.rst
@@ -2,20 +2,23 @@
 LDAP user cleanup
 =================
 
-LDAP User Cleanup is a new feature in the ``LDAP user and group backend`` 
-application. LDAP User Cleanup is a background process that automatically 
-searches the Nextcloud LDAP mappings table, and verifies if the LDAP users are 
-still available. Any users that are not available are marked as ``deleted`` in 
-the ``oc_preferences`` database table. Then you can run a command to display 
-this table, displaying only the users marked as ``deleted``, and then you have 
-the option of removing their data from your Nextcloud data directory.
+Overview
+--------
 
-These items are removed upon cleanup:
+LDAP User Cleanup is a feature in the Nextcloud LDAP application. LDAP User Cleanup periodically and automatically runs in the background, checking the Nextcloud LDAP user mappings and verifying if mapped users are still available in your LDAP directory. Any accounts that are no longer found in LDAP are **marked for deletion** within Nextcloud—this prevents login for those users but does not immediately remove their data.
 
-* Local Nextcloud group assignments
-* User preferences (DB table ``oc_preferences``)
-* User's Nextcloud home folder
-* User's corresponding entry in ``oc_storages``
+.. note::
+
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_user/user_auth_ldap_cleanup.rst
```diff
@@ -2,20 +2,23 @@
...
-These items are removed upon cleanup:
+LDAP User Cleanup is a feature in the Nextcloud LDAP application. LDAP User Cleanup periodically and automatically runs in the background, checking the Nextcloud LDAP user mappings and verifying if mapped users are still available in your LDAP directory. Any accounts that are no longer found in LDAP are **marked for deletion** within Nextcloud—this prevents login for those users but does not immediately remove their data.
 
-* Local Nextcloud group assignments
-* User preferences (DB table ``oc_preferences``)
-* User's Nextcloud home folder
-* User's corresponding entry in ``oc_storages``
+.. note::
+
+   LDAP User Cleanup does *not* immediately delete user accounts or data. When users are no longer found in LDAP, their accounts are **marked for deletion** within Nextcloud. At this stage, all account data—including files, folders, preferences, and group memberships—remains in place. The user is simply prevented from logging in.
```
come-nc, line 12: Marking for deletion does not prevent login per-se.
Users cannot login anymore because they do not match the login filter anymore.

When using another login method there have been issues in the past where the users could login. It’s usually either a bug or misconfiguration though.

Users which are marked as deleted will also disappear from user listing in Nextcloud. I think that’s important to note. In the UI it looks like the user is gone.

admin_manual/configuration_user/user_auth_ldap_cleanup.rst
```diff
@@ -28,31 +31,28 @@ There are two prerequisites for LDAP User Cleanup to operate:
...
-configured with ``ldapUserCleanupInterval``. For example, if you have 200 LDAP 
+configure with ``ldapUserCleanupInterval``. 
+
+For example, if you have 200 LDAP 
 users and your ``ldapUserCleanupInterval`` is 20 minutes, the process will 
 examine the first 50 users, then 20 minutes later the next 50 users, and 20 
 minutes later the next 50, and so on.
 
-The amount of users to check can be set to a custom value via occ command. The
+The amount of users to check can be set to a custom value via OCC. The
```
come-nc, line 41: Are we using this uppercase version in the rest of the documentation :thinking: 
It looks odd to me, original was clearer.

come-nc (commented):
I’m a bit concerned by this version insisting a lot on LDAP user cleanup preventing users from login, it may induce people to think that deleted users can still login until the background job detects them.
This is not the case, as soon as a user is removed from LDAP or do not match the filter anymore, login is impossible.
Also, any user that is found to be missing from the LDAP in the course of any request will also be flagged as deleted. So the background job is not the only way for users to be detected as missing, the software is proactive.

This page should also refer to the option to see remnants as disabled documented in https://github.com/nextcloud/documentation/pull/11741

skjnldsv:
@come-nc could you maybe suggest another approach ?
Could you take over ?

else maybe @joshtrichards want to resume or adjust this PR ? :)

TheJanzap:
I hope someone can clean up this PR soon, because this version is much clearer than the current version. I had to deal with LDAP cleanup again and had to come back to this PR to better understand the process.

PR #13897, opened 2025-11-28
docs(user-manual): add a note about opting out of classification

hamza221:
### ☑️ Resolves

* ref https://github.com/nextcloud/mail/issues/11716

### 🖼️ Screenshots
image


user_manual/groupware/mail.rst
```diff
@@ -73,6 +73,8 @@ Messages will automatically be marked as important based on which messages you i
 
 .. figure:: images/priority-inbox.png
 
+.. note:: You can opt out of classifying messages as important in account settings, or while adding a new mail account.
```
ChristophWurst: This can also be a regular paragraph. Something along "The automatic classification is optional. You can opt-out when setting up an account. The classification can also be turned on and off in the account settings at any time.

PR #13910, opened 2025-12-05
Update migrating_owncloud.rst - oC 10.16

AndyScherzinger:
change to oC 10.16

### DIFF
--- a/admin_manual/maintenance/migrating_owncloud.rst
+++ b/admin_manual/maintenance/migrating_owncloud.rst
@@ -18,6 +18,7 @@ See the table below for a version map, where migrating is easily possible:
 | 10.13.x           | 25.0.13                       |
 | 10.14.x           | 25.0.13                       |
 | 10.15.x           | 25.0.13                       |
+| 10.16.x           | 25.0.13                       |
 +-------------------+-------------------------------+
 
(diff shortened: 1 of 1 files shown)


admin_manual/maintenance/migrating_owncloud.rst
```diff
@@ -18,6 +18,7 @@ See the table below for a version map, where migrating is easily possible:
 | 10.13.x           | 25.0.13                       |
 | 10.14.x           | 25.0.13                       |
 | 10.15.x           | 25.0.13                       |
+| 10.16.x           | 25.0.13                       |
```
szaimen, line 21: Target nc version needs to be updated?

AndyScherzinger: Unsure, needs feedback from @binford6000 - but if you folks agree than fine by me, but than it probably also needs extra text for the php versions.

binford6000: > Unsure, needs feedback from @binford6000 - but if you folks agree than fine by me, but than it probably also needs extra text for the php versions.

No, we should leave the target NC version as is. Main idea was just to not block upgrading directly to latest NC due to version string. Admin documentation may need something like "If you are familiar with deploying php (link to nc<->php version) then you can directly upgrade to the latest NC version.

PR #13935, opened 2025-12-12
feat(objectstore): add bucket created event

leftybournes:
Document BucketCreatedEvent found in https://github.com/nextcloud/server/pull/56565

### 🖼️ Screenshots

bucket_created_event

### DIFF
--- a/developer_manual/basics/_available_events_ocp.rst
+++ b/developer_manual/basics/_available_events_ocp.rst
@@ -527,6 +527,11 @@ Event fired after the filesystem has been torn down
 
 .. versionadded:: 28
 
+``OCP\Files\ObjectStore\Events\BucketCreatedEvent``
+***************************************************
+
+.. versionadded:: 33
+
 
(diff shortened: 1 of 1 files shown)


developer_manual/basics/_available_events_ocp.rst
```diff
@@ -527,6 +527,11 @@ Event fired after the filesystem has been torn down
 
 .. versionadded:: 28
 
+``OCP\Files\ObjectStore\Events\BucketCreatedEvent``
+***************************************************
+
+.. versionadded:: 33
```
joshtrichards, line 533:
```suggestion
.. versionadded:: 28
```

This is fine. I guess this was a feature, but also backported. It's technically in current maintenance releases going back to v29, but originally for v33. Either version is probably fine given that.

susnux: I think 33 is ok as its only available in maintenance releases otherwise and you can not rely on it being available in other major versions.

PR #13946, opened 2025-12-18
feat(upgrade-guide-33): Add change related to snowflake ids

CarlSchwan:
### 🖼️ Screenshots

image


developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_33.rst
```diff
@@ -80,6 +80,17 @@ Default user agent for outgoing requests changed
 
 Starting with this release, the default user agent for requests done by the instance was changed from ``Nextcloud Server Crawler`` to ``Nextcloud-Server-Crawler/X.Y.Z``, where ``X.Y.Z`` is the current server version.
 
+Snowflake IDS
+^^^^^^^^^^^^^
+
+The following tables are now using snowflake ids:
```
artonge, line 86: And also now contain a large int, which need to be casted to string to work with 32bit, right? And which is also too large for JSON number, so need to be a string too.

CarlSchwan: added this info just below

PR #14000, opened 2026-01-19
nextcloud/app_api#727: Clarify optional status

IBBoard:
### ☑️ Resolves

* Improve documentation to go along with nextcloud/app_api#727 and make it clear that the AppAPI is entirely optional. This is most useful if further changes are made for that ticket to use the supplied URL and link users to the docs

### 🖼️ Screenshots

(Not sure how we're supposed to provide screenshots when GitHub doesn't render RST files!)

### DIFF
--- a/admin_manual/exapps_management/AppAPIAndExternalApps.rst
+++ b/admin_manual/exapps_management/AppAPIAndExternalApps.rst
@@ -10,6 +10,9 @@ an ecosystem for **ExApps** (short for "External Apps") was introduced, allowing
 
 Most of our :doc:`Artificial Intelligence <../ai/index>` (AI) apps are developed as ExApps and thus may require some preparation of your Nextcloud instance before you can install them.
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
> It is nice to clarify, but from a performance and security point of view, wouldn't it make more sense to disable it by default and just enable it when it is really needed?

Cc @DaphneMuller, for your team :)

DaphneMuller:
@oleksandr-nc can you have a look at this

oleksandr-nc:
> It is nice to clarify, but from a performance and security point of view, wouldn't it make more sense to disable it by default and just enable it when it is really needed?

Without ExApps installed, this doesn't impact performance or security. And ExApps have a much smaller impact on performance than regular PHP applications.


> @oleksandr-nc can you have a look at this

PR looks good 👍

oleksandr-nc (approved):

PR #14064, opened 2026-02-11
feat(upgrade-34): Add SELECT FOR UPDATE doc

CarlSchwan:
### ☑️ Resolves

* Fix #…


ChristophWurst:
Shouldn't this go onto a general page about the query builder? The upgrade docs idea is to help developers update their app. This is only helpful if they had a raw query before and can now use the query builder. Otherwise this is irrelevant.

ChristophWurst:
Sure, it absolutely can be. But if you are new to Nextcloud development or just read the manual the features should be documented too. It shouldn't be necessary to read all the upgrade docs, no?

PR #14182, opened 2026-03-31
Emphasize account types (CalDAV/CardDAV) for iOS sync in groupware/sync_ios.rst

Helianthropy:
Adds bold to “CalDAV” and "CardDAV" in listed steps to draw attention to which account type to select and adds notes to each respective section to highlight a potential error a user might have made if they selected the wrong account type, not realizing/noticing that they are different.

###
Having struggled with why I couldn’t get calendar syncing to work on iOS versus macOS, I discovered my mistake. Proposing this change to help someone else quickly resolve (or avoid) the issue.

### DIFF
--- a/user_manual/groupware/sync_ios.rst
+++ b/user_manual/groupware/sync_ios.rst
@@ -11,7 +11,7 @@ Calendar
 #. Select Calendar Accounts.
 #. Select Add Account.
 #. Select Other as account type.
-#. Select Add CalDAV account.
+#. Select Add **CalDAV** account.
 #. For server, type the domain name of your server i.e. ``example.com``.
 #. Enter your user name and password.
 #. Select Next.
@@ -30,6 +30,8 @@ Your calendar will now be visible in the Calendar application.
 .. note:: Beginning with iOS 12 an SSL encryption is necessary. Therefore do **not** disable **SSL**
   (For this reason a certificate is required at your domain, https://letsencrypt.org/ will do).
 
+.. note:: If you select **CardDAV** only contact syncing will be made available.
+
 
 Contacts
 --------
@@ -40,7 +42,7 @@ Contacts
 #. Select Contacts Accounts.
 #. Select Add Account.
 #. Select Other as account type.
-#. Select Add CardDAV account.
+#. Select Add **CardDAV** account.
 #. For server, type the domain name of your server and path, i.e., ``example.com/remote.php/dav/principals/users/username/`` (replace **example.com** and **username**).
 #. Enter your user name and password.
 #. Select Next.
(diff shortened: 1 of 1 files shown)


user_manual/groupware/sync_ios.rst
```diff
@@ -50,6 +52,8 @@ You should now find your contacts in the address book of your iPhone.
 .. note:: Beginning with iOS 12 an SSL encryption is necessary. Therefore do **not** disable **SSL**
   (For this reason a certificate is required at your domain, https://letsencrypt.org/ will do).
 
+.. note:: If you select **CalDAV** only calendar syncing will be made available.
+
```
miaulalala, line 56: shouldn't this be the other way around? Or am I missing the purpose of this note?

miaulalala (changes requested):

user_manual/groupware/sync_ios.rst
Helianthropy: @miaulalala, The purpose was to point out potential errors the user may have made. Referring to my specific case, I was wondering why attempting to add a calendar only left me with contact syncing (in this case, it was because I must have previously selected CardDAV and never tried the add step again—merely attempting to refresh my settings in hopes that the calendar would appear). This specifically points out a potential error (and how it manifests) that may not be obvious to notice once it’s made without this foreknowledge. I replicated the same perspective for folks wanting contact syncing who may have only added a CalDAV account.

Retracing my steps when first looking to this documentation for support, after reviewing the steps, I thought everything was looking fine, then checked notes for other hints of an error. The notes did not help me in that moment and it was only after I later noticed the Account Settings title read “CardDAV” that I realized my error.

**I think the other way around would serve a similar purpose in providing a hint to the user trying to figure this out (one account doesn’t support all features), but I feel like pointing out an exact problem might register a bit better for the troubleshooting user who may otherwise feel they *did* follow the instructions correctly**—but being so close to this issue, I might be a bit biased.

skjnldsv: Thanks for the pr @Helianthropy 

As a quick side note, using ai for replying to comments is a bit counter productive as we get overly complicated replies instead of your own words :)
Don't hesitate to just write it as your own, it makes it easier for us to understand. 

Have a nice day ☀️

skjnldsv (approved):

user_manual/groupware/sync_ios.rst
Helianthropy: @skjnldsv, I’ll endeavor to be less verbose lest my own words get mistaken as coming from an LLM. 🙃

skjnldsv (approved):

PR #14180, opened 2026-03-30
feat: split developer uprade docs

ChristophWurst:
* critical changes = must change
* deprecations = should change
* new in this release = nice to have

Goal: make critical changes as concise as possible. Don't overwhelm maintainers with deprecations and new APIs.

Considerations: advertise Rector. Sort critical changes from most likely to affect a maintainer to most niche features.

### 🖼️ Screenshots

| Before | After |
|--------|--------|
| Screenshot 2026-03-30 at 15-36-59 Upgrade to Nextcloud 33 — Nextcloud latest Developer Manual latest documentation | Screenshot 2026-03-30 at 15-37-19 Upgrade to Nextcloud 33 — Nextcloud latest Developer Manual latest documentation |
| - | Screenshot 2026-03-30 at 15-38-22 Deprecations — Nextcloud latest Developer Manual latest documentation |
| - | Screenshot 2026-03-30 at 15-38-09 New in this release — Nextcloud latest Developer Manual latest documentation |

PR #14183, opened 2026-03-31
feat(developermanual): move new and deprecated APIs to dedicated pages

ChristophWurst:
This is the cleaned-up version of https://github.com/nextcloud/documentation/pull/14180.

I did not move the upgrade doc(s) yet as I'm a bit unsure how to do it without link breakage and how to only list the newest one and hide the old ones.

### 🖼️ Screenshots

| Header | Header |
|--------|--------|
| Index | Screenshot 2026-03-31 at 16-46-24 Release notes — Nextcloud latest Developer Manual latest documentation |
| New | Screenshot 2026-03-31 at 16-46-32 New in this release — Nextcloud latest Developer Manual latest documentation |
| Deprecated | Screenshot 2026-03-31 at 16-46-39 Deprecations — Nextcloud latest Developer Manual latest documentation |
| Upgrade guide | Screenshot 2026-03-31 at 16-47-12 Upgrade to Nextcloud 34 — Nextcloud latest Developer Manual latest documentation |

### DIFF
--- a/developer_manual/release_notes/deprecations.rst
+++ b/developer_manual/release_notes/deprecations.rst
@@ -1,0 +1,18 @@
+.. _deprecated-apis:
+
+============
+Deprecations
+============
+
+In order to improve our platform we are phasing out some APIs. Deprecated APIs are not removed before ???,
+unless the breakage can not be avoided.
+
+We highly recommend using the `Nextcloud Rector <https://packagist.org/packages/nextcloud/rector>`_
+rules, which can fix some API changes automatically.
+
+..
+    Add one section for each group of deprecations (e.g. group files api changes, authentication changes).
+    Also deprecate the feature at their dedicated documentation page.
+    After branch-off the contents below will be cleared.
+
+
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_34.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_34.rst
@@ -1,24 +1,15 @@
+.. _upgrade_to_34:
+
 =======================
 Upgrade to Nextcloud 34
 =======================
 
+..
+    Add one section for each change.
+    Only list changes absolutely necessary to keep an app running. Use the dedicated deprecation and new features pages for optional changes and announcements.
+
 Front-end changes
 -----------------
-
-Added APIs
-^^^^^^^^^^
-
-- TBD
-
-Changed APIs
-^^^^^^^^^^^^
-
-- TBD
-
-Deprecated APIs
-^^^^^^^^^^^^^^^
-
-- TBD
 
 Removed APIs
 ^^^^^^^^^^^^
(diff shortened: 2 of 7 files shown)


developer_manual/release_notes/deprecations.rst
```diff
@@ -0,0 +1,18 @@
+.. _deprecated-apis:
+
+============
+Deprecations
+============
+
+In order to improve our platform we are phasing out some APIs. Deprecated APIs are not removed before ???,
```
nickvergessen: ❓❓❓ lol

developer_manual/release_notes/deprecations.rst
```diff
@@ -0,0 +1,18 @@
+.. _deprecated-apis:
+
+============
+Deprecations
```
nickvergessen, line 4: Should also be a file per major for better linking?

developer_manual/release_notes/deprecations.rst
ChristophWurst: Right. That brings up the old discussion about what we promise as deprecation period. IIRC it was three years. Should I write it down or skip?

developer_manual/release_notes/deprecations.rst
ChristophWurst: I don't like it. This will get messy quickly. It also makes it impossible to have a stable link as every major release will have the version embedded in the URL.

ChristophWurst: We could collect all current deprecations on this page. First we list the new ones, then there's the rest. When a deprecation is finally removed, it's also removed from the page.

It's equally messy, but there is one stable URL for the page.

nickvergessen: > it's also removed from the page.

Could have a second page then for those?

ChristophWurst: That would grow endlessly

developer_manual/release_notes/deprecations.rst
ChristophWurst: @blizzz @nickvergessen do we state the 3y period?

skjnldsv: I think we should

nickvergessen: yea

developer_manual/release_notes/deprecations.rst
nickvergessen: Hmm, well then let's continue like this and see how it feels in the future

developer_manual/release_notes/deprecations.rst
nickvergessen: So fix this and we merge?

nickvergessen (approved):

developer_manual/release_notes/deprecations.rst
ChristophWurst: Somehow managed to fix, lose the fix while rebase, and merge 🤦 Fix incoming

PR #14184, opened 2026-04-01
style(docs): add subtle shadow to screenshots

Antreesy:
### ☑️ Resolves

* Fix blending of light-theme screenshots with the background
* Not visible on dark-theme docs, but it's a step forward nevertheless

### 🖼️ Screenshots

Before | After
--- | ---
image | image
image | image


jancborchardt (commented):
Looks good! :) Don’t we use variables here as well tho?
It’s fine to also use some bigger border-radius for images here, could be our `border-radius-large: 10px;`

skjnldsv (approved):

PR #14336, opened 2026-04-16
docs: fix stale and inaccurate occ command documentation

miaulalala:
- Rename "Run occ as your HTTP user" to "Running occ" with explanation of why the HTTP user is needed
- Rename "Run commands in maintenance mode" to "Limitations in maintenance mode" to accurately reflect the section content
- Fix environment variables section: both examples were identical, now shows inline variable and export as separate approaches
- Update background jobs: cron is the primary command, list all three modes in the command listing
- Remove db:generate-change-script (removed from server in https://github.com/nextcloud/server/pull/5772)
- Add note that antivirus commands require the files_antivirus app

Screenshots and PR description AI-Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### 🖼️ Screenshots
screencapture-file-home-anna-Documents-occ-diff-html-2026-04-16-13_49_51
screencapture-file-home-anna-Documents-occ-diff-html-2026-04-16-13_49_29

### DIFF
--- a/admin_manual/occ_command.rst
+++ b/admin_manual/occ_command.rst
@@ -52,8 +52,11 @@ occ command Directory
 
 .. _http_user_label:
 
-Run occ as your HTTP user
--------------------------
+Running occ
+-----------
+
+You must run ``occ`` as your HTTP user so that the file ownership and permissions
+on your Nextcloud data directory stay consistent with the web server.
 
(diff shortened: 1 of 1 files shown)


szaimen (approved):
LGTM, thanks! :)

PR #14574, opened 2026-04-24
feat(devmanual): clean up the new release notes

ChristophWurst:
### ☑️ Resolves

Follow-up to https://github.com/nextcloud/documentation/pull/14183 and https://github.com/nextcloud/documentation/pull/14565

This brings the *Release notes* into the final structure I had in mind with https://github.com/nextcloud/documentation/pull/14180

* There is one very prominent entrypoint for changes in this release
* There is one dedicated page for *critical changes*. 
  * These *have to be addressed*. It's cleared after branch-off because afterwards it is no longer relevant.
* There is one dedicated page for *new APIS*.
  * These are nice-to-have for devs, nothing terrible happens if that page is not read.
  * The APIs will have their main documentation on a persistent page. This is just an overview here.
  * It's cleared after branch-off because afterwards it is no longer relevant.
* There is one collecting page for *deprecations*.
  * These are nice-to-have for devs, nothing terrible happens if that page is not read.
  * They stay there as long as they are relevant, so until they are removed.

The old upgrade guides stay for now and will phase out. Maybe we hand-pick the deprecations and move them to the new file. I'm too lazy for that right now.
There is a redirect rule to send anyone from the 34 page to the new pages.

### 🖼️ Screenshots

| Entry point | Critical | New | Deprecated | Legacy |
|--------|--------|--------|--------|--------|
| Screenshot 2026-04-24 at 13-47-10 Release notes — Nextcloud latest Developer Manual latest documentation | Screenshot 2026-04-24 at 13-47-19 Critical changes — Nextcloud latest Developer Manual latest documentation | Screenshot 2026-04-24 at 13-47-29 New in this release — Nextcloud latest Developer Manual latest documentation | Screenshot 2026-04-24 at 13-47-36 Deprecations — Nextcloud latest Developer Manual latest documentation | Screenshot 2026-04-24 at 13-47-53 App upgrade guide — Nextcloud latest Developer Manual latest documentation |

### DIFF
--- a/developer_manual/conf.py
+++ b/developer_manual/conf.py
@@ -177,4 +177,6 @@ redirects = {
     "digging_deeper/changelog": "../app_publishing_maintenance/app_upgrade_guide/index.html",
     # Removed 2025-04
     "basics/front-end/l10n": "../translations.html",
+    # Moved 2026-04 - Upgrade to 34 moved to release notes
+    "app_publishing_maintenance/app_upgrade_guide/upgrade_to_34": "../../release_notes/critical_changes.html",
 }
--- a/developer_manual/app_publishing_maintenance/app_upgrade_guide/index.rst
+++ b/developer_manual/app_publishing_maintenance/app_upgrade_guide/index.rst
@@ -6,12 +6,15 @@ App upgrade guide
 
 Once you've created and published the first version of your app, you will want to keep it up to date with the latest Nextcloud features.
 
+.. note::
+
+   For the latest release notes, see the :doc:`Release Notes </release_notes/index>`.
+
 These sub pages will cover the most important changes in Nextcloud, as well as some guides on how to upgrade existing apps.
 
 .. toctree::
    :maxdepth: 1
 
-   upgrade_to_34.rst
    upgrade_to_33.rst
    upgrade_to_32.rst
    upgrade_to_31.rst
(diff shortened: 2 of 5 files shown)


developer_manual/conf.py
```diff
@@ -177,4 +177,6 @@
     "digging_deeper/changelog": "../app_publishing_maintenance/app_upgrade_guide/index.html",
     # Removed 2025-04
     "basics/front-end/l10n": "../translations.html",
+    # Moved 2026-04 - Upgrade to 34 moved to release notes
+    "app_publishing_maintenance/app_upgrade_guide/upgrade_to_34": "../../release_notes/critical_changes.html",
```
nickvergessen, line 181: should not forget to remove this after branch off? :P

ChristophWurst: It wouldn't do any harm, right? But yes can be removed at some point

ChristophWurst: The single purpose of this redirect is for someone who has already visited https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_34.html to get sent to the new pages instead of a 404.

skjnldsv (approved):
Sure! 👍

susnux:
Nice but I think this is a bit confusing to have another section split across the documentation.
Why not move all old notes to this new section as well like:

* Release notes
  * Critical changes
  * New in this version
  * Deprecations
  * Previous notes

We can still make the old links work with the redirects we already have in place for SEO.

ChristophWurst:
Sounds good to move them to this place and add redirects :+1:

PR #11420, opened 2024-01-03
Update external_storage_configuration_gui - cmd to run file checker i…

bhugh:
…n Docker Container

A different command is needed for the cron job, when NextCloud is running in a Docker container.  A short paragraph is added to explain this and show the needed command.

### ☑️ Resolves

* Fix #…

PR #14674, opened 2026-04-29
fix(occ-check): exclude docker-based occ commands from syntax check

skjnldsv:
This allow OCC commands pointing to docker AIO to pass https://github.com/nextcloud/documentation/pull/11420


skjnldsv:
Directly pushed in https://github.com/nextcloud/documentation/pull/11420

PR #14791, opened 2026-05-07
docs: Clarify recovery key documentation for master key vs. per-user key modes

### LINKED ISSUE #1340: Recovery key documentation outdated since NC 13?
https://help.nextcloud.com/t/recovery-key-field-missing/50590

Please clarify and correct the documentation if that is not correct any more.

skjnldsv:
## Summary

This PR fixes issue #1340 by clarifying the recovery key documentation, which has been confusing users since Nextcloud 13 changed the default encryption mode.

### Problem

Users reported that recovery key configuration options don't appear in the default Nextcloud installation, suggesting the documentation was outdated. This confusion was reported in:
- GitHub issue #1340: "Recovery key documentation outdated since NC 13?"
- Help forum: https://help.nextcloud.com/t/recovery-key-field-missing/50590
- Server issue nextcloud/server#8283

### Root Cause

Since Nextcloud 13, the default encryption mode uses **master keys** (system-wide encryption), but **recovery keys only exist in per-user key mode** (legacy, opt-in). The documentation didn't clearly state this limitation.

### Solution

Updated `admin_manual/configuration_files/encryption_configuration.rst` to:

1. **Update Key Management Modes section** to explicitly state:
   - Master Key (default): No recovery key options available
   - User Keys: Recovery keys are optional and only available in this mode

2. **Add reference label** for cross-linking from troubleshooting section

3. **Add caution box** in "Enabling file recovery keys" subsection explaining:
   - Recovery keys only work in per-user key mode
   - Default master key mode doesn't expose recovery key UI
   - How to switch modes if needed (`occ encryption:disable-master-key`)

4. **Add troubleshooting entry** directly addressing the issue:
   - "Why don't I see the recovery key option in the Encryption settings?"
   - Explains the master key default behavior
   - References server issue nextcloud/server#8283 for technical context
   - Clarifies that master key mode is recommended for most deployments

### Changes Made

- Enhanced Key Management Modes comparison with explicit recovery key information
- Added prominent warning box before recovery key configuration instructions
- Added new troubleshooting subsection with answer to the exact question users were asking
- Added reference to server issue nextcloud/server#8283 explaining the design decision

### Verification

✅ Documentation now clearly states recovery keys only exist in per-user key mode
✅ Troubleshooting section directly addresses GitHub issue #1340
✅ Commands to switch modes are clearly referenced
✅ Master key recommendation is clear with reasoning

### Related Issues

Fixes #1340
Related to nextcloud/server#8283

### Testing

The changes are documentation-only and don't affect functionality. The RST syntax has been verified and should build correctly.

### DIFF
--- a/admin_manual/configuration_files/encryption_configuration.rst
+++ b/admin_manual/configuration_files/encryption_configuration.rst
@@ -123,17 +123,15 @@ Key Management Modes
 
 - All files are encrypted with a central server-controlled key.
 - Admins can decrypt any user’s files.
-- Offers better performance and compatibility with more login/authentication modes.
-- **Recovery keys are not available in master key mode.** If a user forgets their password, admins can reset it and files remain accessible (protected by the master key, not the user password).
-- Recommended for most deployments due to improved performance and compatibility.
+- **Recovery keys are not available in master key mode.** Files remain accessible if a user forgets their password, as they are encrypted by the master key, not the user password.
+- Recommended for most deployments.
 
 **User Keys:**
 
 - Each user’s files are encrypted with a password-protected key.
 - Admins cannot (readily) decrypt files without the user's password, unless a recovery key is defined.
-- **Recovery keys are available in user key mode** as an optional safeguard: if a user forgets their password, admins can use the recovery key to reset access.
 - If a user forgets their password and no recovery key exists, their files are lost.
-- This mode requires more resources and does not work with all authentication methods (e.g., app passwords, single sign-on).
+- This mode does not work with all authentication methods (e.g., app passwords, single sign-on) and is only recommended for compatibility with older setups.
 
 **How to choose:**
 
@@ -313,18 +311,17 @@ User Keys: Sharing & Recovery
 
 **Enabling file recovery keys:**
 
-.. caution::
-   Recovery keys are **only available in per-user key mode**, not in the default master key mode.
-   If you do not see recovery key options in your Admin Encryption settings, your instance is using
-   master key mode (the default and recommended mode). To use recovery keys, you must first switch
-   to per-user key mode by running ``occ encryption:disable-master-key`` on a fresh installation
-   (before any files are encrypted).
+Recovery keys are only available in per-user key mode (not the default master key mode).
 
 - If you lose your Nextcloud password, you lose access to your encrypted files.
 - If a user loses their password, their files are unrecoverable unless a recovery key is enabled (per-user key mode only).
 - To enable recovery (in per-user key mode), go to Encryption in Admin page and set a recovery key password.
 - Users must enable password recovery in their Personal settings for the Recovery Key to work.
 - For users who have enabled password recovery, admins can reset passwords and recover files using the Recovery Key.
+
+.. warning::
+   The recovery process can be slow and resource-intensive, especially for instances with large amounts of encrypted data.
+   Test recovery procedures before relying on them in production.
 
 .. figure:: images/encryption10.png
 .. figure:: images/encryption7.png
@@ -343,24 +340,18 @@ Troubleshooting
 Why don't I see the recovery key option in the Encryption settings?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Recovery keys are **only available in per-user key mode**. Since Nextcloud 13, the default
-encryption mode uses **master keys** (system-wide encryption), which offer better performance
-and compatibility. Master key mode does not expose recovery key options in the Admin settings
-because recovery keys are not needed—admins can reset user passwords and files remain accessible.
+Recovery keys are only available in per-user key mode. Since Nextcloud 13, the default
+encryption mode uses master keys (system-wide encryption). Master key mode does not expose
+recovery key options in the Admin settings because recovery keys are not needed—admins can
+reset user passwords and files remain accessible.
 
-If you need recovery key functionality, you must switch to per-user key mode before encrypting
-any files::
+If you are using master key mode (the default and recommended mode), you do not need recovery
+keys. Recovery keys are only relevant for per-user key setups, which are maintained for
+compatibility with older deployments.
 
-   occ encryption:disable-master-key
-
-This command only works on fresh installations without existing encrypted data.
-
-For most use cases, **master key mode is recommended**. Recovery keys add complexity and are only
-needed in edge cases where per-user key encryption is required. See :ref:`Key Management Modes
-<encryption_configuration_key_management_modes>` for guidance on choosing the right mode.
-
-See also `GitHub Issue #8283 <https://github.com/nextcloud/server/issues/8283>`_ for technical
-context on this design decision.
+See :ref:`Key Management Modes <encryption_configuration_key_management_modes>` for guidance
+on the differences between master key and per-user key modes, and `GitHub Issue #8283
+<https://github.com/nextcloud/server/issues/8283>`_ for technical context on this design decision.
 
 Invalid private key for encryption app
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/admin_manual/configuration_files/encryption_configuration.rst
+++ b/admin_manual/configuration_files/encryption_configuration.rst
@@ -114,6 +114,8 @@ Before You Enable Encryption
 2. Back up your instance configuration and all encryption keys in a safe location before proceeding.
 3. Decide which key management mode suits your needs (see below).
 
+.. _encryption_configuration_key_management_modes:
+
 Key Management Modes
 --------------------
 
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -122,13 +124,16 @@ Key Management Modes
 - All files are encrypted with a central server-controlled key.
 - Admins can decrypt any user’s files.
 - Offers better performance and compatibility with more login/authentication modes.
-- Best for organizations that need to recover files if users forget their passwords.
+- **Recovery keys are not available in master key mode.** If a user forgets their password, admins can reset it and files remain accessible (protected by the master key, not the user password).
+- Recommended for most deployments due to improved performance and compatibility.
```
artonge: Is it really more performant?
And compatibility with what?

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -122,13 +124,16 @@ Key Management Modes
...
 
 **User Keys:**
 
 - Each user’s files are encrypted with a password-protected key.
 - Admins cannot (readily) decrypt files without the user's password, unless a recovery key is defined.
+- **Recovery keys are available in user key mode** as an optional safeguard: if a user forgets their password, admins can use the recovery key to reset access.
 - If a user forgets their password and no recovery key exists, their files are lost.
+- This mode requires more resources and does not work with all authentication methods (e.g., app passwords, single sign-on).
```
artonge: To be more precise, but I think it is negligible, so not sure if we should even mention it.

```suggestion
- This mode requires more disk space and does not work with all authentication methods (e.g., app passwords, single sign-on).
```

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -122,13 +124,16 @@ Key Management Modes
 - All files are encrypted with a central server-controlled key.
 - Admins can decrypt any user’s files.
 - Offers better performance and compatibility with more login/authentication modes.
-- Best for organizations that need to recover files if users forget their passwords.
+- **Recovery keys are not available in master key mode.** If a user forgets their password, admins can reset it and files remain accessible (protected by the master key, not the user password).
```
artonge:
```suggestion
- **Recovery keys are not available in master key mode.** If a user forgets their password, admins can reset it and files remain accessible as they are encrypted by the master key, not the user password.
```

admin_manual/configuration_files/encryption_configuration.rst
artonge: The argument is rather that it is fail proof if the user forgets its password. Another is that user key encryption does not make sense in the threat model of server side encryption, it adds unnecessary complexity, for zero added protection.

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -308,9 +313,16 @@ User Keys: Sharing & Recovery
 
 **Enabling file recovery keys:**
 
+.. caution::
+   Recovery keys are **only available in per-user key mode**, not in the default master key mode.
+   If you do not see recovery key options in your Admin Encryption settings, your instance is using
+   master key mode (the default and recommended mode). To use recovery keys, you must first switch
+   to per-user key mode by running ``occ encryption:disable-master-key`` on a fresh installation
+   (before any files are encrypted).
```
artonge: I don't think anybody will want to switch to user key only to use recovery keys.

```suggestion
   master key mode (the default and recommended mode).
```

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -308,9 +313,16 @@ User Keys: Sharing & Recovery
 
 **Enabling file recovery keys:**
 
+.. caution::
+   Recovery keys are **only available in per-user key mode**, not in the default master key mode.
```
artonge: This caution looks like a duplicate of the big paragraph below. Do we really need it?

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -328,6 +340,28 @@ LDAP and External User Backends
 Troubleshooting
 ---------------
 
+Why don't I see the recovery key option in the Encryption settings?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Recovery keys are **only available in per-user key mode**. Since Nextcloud 13, the default
+encryption mode uses **master keys** (system-wide encryption), which offer better performance
+and compatibility. Master key mode does not expose recovery key options in the Admin settings
```
artonge:
```suggestion
encryption mode uses **master keys** (system-wide encryption). Master key mode does not expose recovery key options in the Admin settings
```

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -328,6 +340,28 @@ LDAP and External User Backends
...
+because recovery keys are not needed—admins can reset user passwords and files remain accessible.
+
+If you need recovery key functionality, you must switch to per-user key mode before encrypting
+any files::
```
artonge: Again, not body need that feature unless they are using user key. I would delete.

artonge (commented):
Very verbose I think, I would only add a warning near the recovery key paragraph, but this is adding a lot of noise IMO. And also a lot of confusion as the AI generate part are truth-like but to truth.

admin_manual/configuration_files/encryption_configuration.rst
skjnldsv: Well, one master is always gonna be more performant than user-encryption, no ?
That's what I assumed when reading other threads.
- File sharing doesn't require re-encrypting files for each user for example
- Group changes don't trigger encryption overhead etc etc

skjnldsv: But I agree, this is not clear enough. Also, if you think the gain is minimal, then, we can drop that paragraph. But I thought this would be important as user encrypotion looks much more messy than one global master encryptino approach, no ?

admin_manual/configuration_files/encryption_configuration.rst
skjnldsv: AH, this refers to my other comment, I can drop mentions of this, sure ! 👍

admin_manual/configuration_files/encryption_configuration.rst
skjnldsv: It's here because it caused confusion. You can see on the foruml and the linked tickets that many users did not understa,d you could only have this on user keys.
I have no clue why people would want this, but if they do, they should know how to ?

admin_manual/configuration_files/encryption_configuration.rst
schiessle: > Group changes don't trigger encryption overhead etc etc

Just to add, it is not only overhead but if you share a file with a group and later add someone to the group they will see the file but can not open it because encryption for all recipients only happens when the file was shared and at this point in time the user was not in the group. This results in really bad UX.

admin_manual/configuration_files/encryption_configuration.rst
schiessle: I agree with @artonge , I would remove the last sentence "To use recovery keys, you must first switch
   to per-user key mode by running ``occ encryption:disable-master-key`` on a fresh installation
   (before any files are encrypted)." 

You can anyway only do it right at the beginning before any files are encrypted.

admin_manual/configuration_files/encryption_configuration.rst
schiessle: I think we should make clear that recovery key is not available in the default mode "master key".

But I get the point from @artonge, there is no need for the recovery key by itself. Because you will never have a need for it in the default mode. Their might be a need for per-user-keys (although I think the downsides are larger than all theoretical advantages) and than it is good to have the recovery key option. But I would try to avoid to "generate" a need for the recovery key and therefore make people think it is a good idea to enabled per-user-keys. It is in all aspects a bad idea:

- less performant
- bad UX (as described above with group shares,...)
- limitations (like no SSO)
- recovery-key solved a problem you don't have if you stick to the default

schiessle: also keep in mind that with the recovery process many things can go wrong or it can take very long if people have a lot of files.

The document should position the default (master key) as the state of the art for everyone who enabled server-side encryption today and position the per-user-key as a "compatibility option" for people who have this kind of setup for historical reasons.

skjnldsv: Thanks for the input @schiessle !
Really appreciated!

admin_manual/configuration_files/encryption_configuration.rst
artonge: Indeed, I forgot about the sharing aspect, so we can keep the performance then :+1:

skjnldsv:
Updated @artonge and @schiessle

artonge (approved):

admin_manual/configuration_files/encryption_configuration.rst
```diff
@@ -328,6 +337,22 @@ LDAP and External User Backends
 Troubleshooting
 ---------------
 
+Why don't I see the recovery key option in the Encryption settings?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
skjnldsv:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

come-nc (approved):

Issue #14036, opened 2026-01-29
Documentation Enhancement Suggestions


sproggit:
### Before You Submit

- [ ] I have checked for existing issues and pull requests related to my suggestion.
- [ ] I have posted on the [Nextcloud Community Help Forum](https://help.nextcloud.com) if my suggestion is a troubleshooting scenario or I'm unsure about the solution.

### Type of Suggestion

Suggesting an improvement

### Relevant Documentation Links

https://docs.nextcloud.com/server/stable/admin_manual/installation/php_configuration.html

### Summary

Minor Suggestins for PHP and Redis in the Documentation

### Details

I've just completed a migration of my local instance of 32.0.5 and hit a couple of issues - and wanted to share suggested enhancements for the documentation. For context, I've been performing the "bare metal" installation on a native Raspberry Pi running the latest Trixie OS version. 

1. Outdated List of Required PHP Modules [and other dependencies]. 
The list at the above URL is now increasingly out of date as more of the optional modules seem to be bundled by default with the core PHP package set. I've just deployed on PHP 8.4 and several of the identified modules cannot be found because they seem to be bundled [I'm using Debian packages via Raspberry Pi OS]. 

Similarly, when I ran "apachectl -M" on both my "old" and "new" hosts, I quickly spotted that the "new" host did not have "proxy_module (shared)" and "proxy_fcgi_module (shared)". I've raised this as a question with the Raspberry Pi folks, but anyone else performing a similar migration on the Pi platform will hit the same issue. It makes sense to capture the right response to stuff like this - or at least make clear if a gap is critical or not. My sense is that the community would gladly share given the means to do so. 

2. Major Issue - Redis
[...]

PR #14792, opened 2026-05-07
Add guidance about optional services in migration guide

skjnldsv:
## Summary
Adds explicit guidance about optional services in the migration guide to prevent migration failures.

## Problem
When migrating Nextcloud to a new server, users often encounter errors like "Redis server went away" because they fail to configure optional services (Redis, Memcached, etc.) on the new machine before completing the migration. The current migration guide makes no mention of these dependencies.

## Solution
Added an "important" notice in Step 1 of the migration guide that instructs users to:
1. Check the original system's `config/config.php` for optional services
2. Identify commonly configured services (Redis, Memcached, external storage, etc.)
3. Install and configure these same services on the new machine BEFORE copying Nextcloud files

## Changes
- Updated [admin_manual/maintenance/migrating.rst](admin_manual/maintenance/migrating.rst)
- Added explicit list of services to check for (Redis, Memcached, external storage, LDAP, mail, search backends)
- Emphasized the critical timing: services must be configured before startup

## Relates to
#14036

### DIFF
--- a/admin_manual/maintenance/migrating.rst
+++ b/admin_manual/maintenance/migrating.rst
@@ -37,6 +37,20 @@ the new location. It is also assumed that the authentication method
     your original machine was installed recently just copying that base
     configuration is a safe best practice.
 
+    .. important:: Before beginning the migration, **check the ``config/config.php`` file
+       on your ORIGINAL system** to identify any optional services that are configured,
+       such as:
+
+       * Redis or Memcached (for caching/sessions)
+       * External object storage (S3, etc.)
+       * LDAP
+       * Mail server settings
+       * Full-text search backends (Elasticsearch, etc.)
+
+       **You must also install and configure these same services on the NEW machine
+       before copying the Nextcloud files.** Failure to do so may cause errors such
+       as "Redis server went away" or connection failures during Nextcloud startup.
+
 
 #.  On the original machine then stop Nextcloud. First activate the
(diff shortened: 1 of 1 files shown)


miaulalala (approved):
🍓  nice

sproggit:
as the author of the Enhancement Suggestion that prompted the creation of this Issue, I think this captures the challenge well. I'm not sure if it helps to clarify further, but in the hope that it does... I think there are a couple of elements here, only one of which can be addressed by NextCloud... 

The method I chose to take to port a running instance to a new server was to create a new server environment and then perform a vanilla installation of NextCloud on the target host, using a new domain name and IP address. I then used that vanilla environment to debug dependencies and oddities before following the guide in terms of porting the site content to the new host. What I found was that the "vanilla site" worked just fine, but the the ported version didn't. 

I appreciate this might be difficult to do at the NextCloud platform level [because some dependencies might be introduced by optional Apps] but is there any way that NextCloud is or can be made aware of those dependencies, such that an administrator could migrate a "dependency list" out of their existing environment and use that to sanity check the destination?

skjnldsv:
@sproggit thanks for the input :)
So, your suggestion of having those listed at the NC level is great. But we don't really have a migration settings page on nc. We do have the user migration app, but this is really more about user data than it is about infrastructure.

I think having it in the docs should be good enough :+1:

PR #14852, opened 2026-05-14
fix(docs): list config.php keys to review when migrating servers

### LINKED ISSUE #12919: Copy config.php for "Migrating to a different server"
In documentation for "Migrating to a different server" in step 4 (and in referenced "Backup" and "Restoring backup") it is described to copy config files / folder. 
Im wondering about that because of some entries in config.php: 
1. datadirectory: Also in step 4 it is said that "Depending on the original installation method and the OS the files are located in different locations". But location of data files is stored in config.php and with copying it to new installation I would overwrite location setting. So either I have to copy config.php, so use same location as before (if just entry in config.php is containing it) and then also copy data to that folder.
2. dbpassword: New database has different password than old. So with overwriting config.php acces should not work anymore, or?
3. version: I did migration to upgrade NCPI, as direct update did not work in that case. So target has different version and at least displayed version would be wrong.
4. Other entries? So far I did migration without copying config.php, nextcloud came up but I did not dare to really switch to new one. Maybe other entries will cause issues.

skjnldsv:
### ☑️ Resolves

* Fix #12919

### What changed and why

The migration guide told users to copy `config/config.php` to the new server and only said *"if you change any paths, adapt them in config.php"*. That is not enough — several keys are server-specific and will silently be wrong after a copy, causing startup errors or data corruption:

- `datadirectory` — path should not change (only on specific very rare cases)
- `dbhost` / `dbname` / `dbuser` / `dbpassword` — new DB instance = new credentials
- `trusted_domains` — must include the new hostname
- `overwrite.cli.url` — must reflect the new URL
- `memcache.*` / `redis` connection params — cache host may differ
- `logfile` — log path may differ

Added an `important` admonition in step 4 (the file-copy step) that explicitly lists each of these keys and explains the consequence of leaving them stale.

### 🖼️ Screenshots

image
### ✅ Checklist

- [ ] I have run `codespell` or similar and addressed any spelling issues

### DIFF
--- a/admin_manual/maintenance/migrating.rst
+++ b/admin_manual/maintenance/migrating.rst
@@ -79,6 +79,45 @@ the new location. It is also assumed that the authentication method
       Changing the location of the data directory might cause a corruption of relations
       in the database and is not supported.
 
+    .. important::
+      After copying ``config/config.php`` to the new machine, **review and update every
+      server-specific value** before starting Nextcloud. At minimum check:
+
+      * ``datadirectory`` — keep this path identical to the original server wherever
+        possible. Changing the data directory path requires a database update and
+        is strongly discouraged; see :ref:`troubleshooting_data_directory`
+        if you have no choice.
+      * ``dbhost``, ``dbname``, ``dbuser``, ``dbpassword`` — update if the new server
+        uses a different database host or credentials.
+      * ``trusted_domains`` — add or replace the new server's hostname or IP address.
+      * ``overwrite.cli.url``, ``overwritehost``, ``overwriteprotocol``,
+        ``overwritewebroot`` — update to reflect the new server's URL and any
+        reverse-proxy setup.
+      * ``memcache.local``, ``memcache.distributed``, ``memcache.locking`` and the
+        associated connection parameters — update if the cache/session backend address
+        changed on the new machine. Depending on the backend in use, check ``redis``
+        or ``redis.cluster`` (for Redis / Redis Cluster) or ``memcached_servers``
+        (for Memcached).
+      * ``objectstore`` — if external object storage (S3, Swift, etc.) is configured,
+        verify that the endpoint, bucket, and credentials are still valid and reachable
+        from the new server.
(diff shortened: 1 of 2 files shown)


nickvergessen:
Changing data directory is not supported and should be avoided at any cost.
In theory a change of oc_storage entry is also needed and things like that

skjnldsv:
> Changing data directory is not supported and should be avoided at any cost. In theory a change of oc_storage entry is also needed and things like that

We indeed have a dedicated section about this. https://docs.nextcloud.com/server/stable/admin_manual/issues/general_troubleshooting.html#troubleshooting-data-directory

skjnldsv:
Done @nickvergessen

skjnldsv:
Updated screenshot too

admin_manual/maintenance/migrating.rst
```diff
@@ -79,6 +79,41 @@ the new location. It is also assumed that the authentication method
...
+      * ``objectstore`` — if external object storage (S3, Swift, etc.) is configured,
+        verify that the endpoint, bucket, and credentials are still valid and reachable
+        from the new server.
+      * ``mail_smtphost``, ``mail_smtpport`` — update if the SMTP relay differs on
+        the new server.
+      * ``logfile`` — update if the log path differs on the new server.
+      * ``tempdirectory`` — if set to a custom path, make sure that path exists and
+        is writable on the new server.
+
+      The ``secret`` and ``instanceid`` values are generated once at install time and
```
susnux, line 114: `serverid` must  also not change!

skjnldsv: Oh, didn't know that one !

skjnldsv: Well, it's not one I generated by default. And its documentation states it shouldn't change already. Should I really add it here you think? 
😊

susnux: Well it must not change otherwise there can be problems with snowflakes but hopefully they do not manually edit it :sweat_smile:

susnux (approved):

PR #14855, opened 2026-05-14
Clarify autocomplete share type values

mzl2233:
This clarifies the autocomplete API documentation by listing the valid `shareTypes` values next to the guest-account example, including `8` for guest accounts and the default user value. It also fixes the surrounding wording so readers can tell which query parameter to use and what happens when it is omitted.


skjnldsv:
Hey! Thanks for the PR.
Could this be a table instead please? :)

I think that would look much better

skjnldsv (approved):

skjnldsv:
Done, converted to a table directly on the branch.

PR #14937, opened 2026-05-20
docs(admin): document external storage change detection and limitations

### LINKED ISSUE #7849: Explanation why using external storage outside of Nextcloud is unreliable and hints to make it work
### Todos
- [ ] refine draft and maybe add an example with a real tree to illustrate
- [ ] add section in troubleshooting "why is Nextcloud not detecting changes" that links here
- [ ] add link to the section about rescan frequency that links here

Goal: the goal is to be able to point people at this explanation without needing to add anything. It needs to be understandable right away and straightforward.

### Draft

When mounting an external storage in Nextcloud but making changes in said external storage outside of Nextcloud, Nextcloud will not know about these changes right away because they are not indexed.

This means that desktop clients also will not sync the new changes because they are not aware of them.

1) By default with the option "Check once every access", whenever a user accessed the file tree in the web UI or mobile (or a PROPFIND is made on a deep folder), Nextcloud will rescan the tree level that got accessed and update its index (aka filecache) with the newly found information. Because this is driven by user actions, this can be very random and is not reliable.

2) The other options "Always" or "Never" will also not help because "Always" will check multiple times per PHP request, but still driven by user actions.

The desktop client only access the root of the sync folder because looking deeper would be too expensive.
To be able to sync deep changes, it needs to find a trail of changed etags to follow up to the tree level where the changes happened.
For this, it requires someone else to detect those changes.
As described in 1), this can happen by chance if a user happens to be accessing the folder that contains the change, then the etag changes will be propagated automatically up to the root, so the desktop client will find the trail eventually.

What we want is that Nextcloud is able to automatically detect changes in the deep folder structure.

There are two ways to achieve this:

A) For SMB storages, use the files_external:notify command which listens to change events sent by SMB itself and then tells Nextcloud which tree level needs to be rescanned, which will result in the wanted etag propagation.

B) For non-SMB, setup a cron job that runs "occ files:scan -p /path/to/storage" periodically.
This has the drawback that it will rescan the whole storage periodically and the changes are only detected whenever the scan is running, so it can be slow.

Unfortunately most storage kinds do not support notification unlike SMB, so there can be no reliable way of receiving such updates in time.

Limitations: in scenario B) it is not possible for the scanner to detect renamed entries reliably, so it will believe that the old entry was deleted and that a new one was created. This will cause metadata loss because the metadata of the old entry will be deleted. Nextcloud metadata includes shares, tags, comments, activities, etc associated with the entry in question.

skjnldsv:
### ☑️ Resolves

* Fixes #7849

### What changed and why

Expanded the external storage admin documentation to explain a common source of confusion: changes made directly on a mounted storage (outside Nextcloud) are not automatically visible to Nextcloud, desktop clients, or mobile apps.

The thin "Adding files to external storage" section has been replaced with a proper `Detecting changes made outside Nextcloud` section that covers:

- **Why it happens**: Nextcloud maintains an internal filecache; external changes are invisible until a rescan updates it.
- **How "Filesystem check frequency" actually works**: it only fires during a WebDAV `PROPFIND`, driven by user actions, not background polling. The mount option bullet in the UI description is also expanded to make this clear.
- **Why desktop clients miss deep changes**: clients only PROPFIND the sync root and rely on etag propagation up the tree; without a rescan at the changed level there is no trail to follow.
- **Solution A (SMB)**: `occ files_external:notify` for near-real-time detection via SMB change events.
- **Solution B (all others)**: periodic `occ files_external:scan <mount_id>` cron job.
- **Rename limitation**: any rescan (periodic or on direct access) cannot detect renames reliably; the old entry is treated as deleted, causing permanent metadata loss (shares, tags, comments, activities). Documented as a `.. warning::`.

A `.. seealso::` cross-reference was added to `smb.rst` at the top of the "SMB update notifications" section pointing back to the new anchor (`external_storage_change_detection_label`).

### :camera_flash: Screenshots

image
image
image
image

### ✅ Checklist

### DIFF
--- a/admin_manual/configuration_files/external_storage_configuration_gui.rst
+++ b/admin_manual/configuration_files/external_storage_configuration_gui.rst
@@ -103,7 +103,12 @@ individually with the following options:
 * Encryption
 * Previews
 * Enable Sharing
-* Filesystem check frequency (Never, Once per direct access)
+* **Filesystem check frequency** — controls how the server rescans an external
+  storage path when a WebDAV ``PROPFIND`` request arrives for it.
+  **Once per direct access** rescans on each ``PROPFIND`` request;
+  **Never** skips the rescan entirely. This setting does *not* cause Nextcloud
+  or desktop clients to poll automatically. See
+  :ref:`external_storage_change_detection_label` for details.
 * Mac NFD Compatibility
 * Read Only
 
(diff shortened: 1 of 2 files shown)


admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
 
-If you are running Nextcloud AIO, the equivalent command
-in that environment is ``sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ files:scan --all``.
+The **Filesystem check frequency** mount option (see
+:ref:`external_storage_mount_options_label`) controls what happens when a WebDAV
+``PROPFIND`` request reaches the Nextcloud server for a path inside the external
+storage:
+
+* **Once per direct access** — the server rescans that directory level on the first
+  ``PROPFIND`` per PHP request, then updates the filecache for what it finds there.
```
artonge: Not sure whether the initial version made sense.
```suggestion
* **Once per direct access** — during ``PROPFIND`` requests made to a given directory, 
  Nextcloud rescans that directory level then updates the filecache with what it finds there.
```

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
+* **Never** — the server skips the rescan entirely and always returns the cached
+  state.
```
artonge: Just to remove the suggestion that the filecache table is a cache.
```suggestion
* **Never** — the server never rescans the external storage during ``PROPFIND`` requests
  and always returns the content stored in its internal index.
```

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
+The desktop sync client only issues a ``PROPFIND`` on the root of the sync folder
+on every sync cycle because scanning the entire tree would be too expensive. It
+relies on **etag propagation**: when a file changes, the etag of every parent
+folder up to the root must also change, giving the client a trail to follow down
+to the changed file.
+
+When a change is made outside Nextcloud, no etag is updated. The client sees no
+trail from the root and never follows it down to the changed file — even if the
+file itself is eventually rescanned at a deeper level, the propagation up to the
+root must still happen before the client can discover it.
```
artonge: > even if the file itself is eventually rescanned at a deeper level, the propagation up to the
root must still happen before the client can discover it.

The propagation happens immediately during the scan, so no need to phrase it as if it is a long process. I would remove the sentence altogether.

```suggestion
trail from the root and never follows it down to the changed file.
```

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
+
+.. code-block:: console
+
+   sudo -E -u www-data php occ files:scan --path="/user/files/MountName"
+
+Or to rescan all files for all users (including all external storages):
+
+.. code-block:: console
+
+   sudo -E -u www-data php occ files:scan --all
```
artonge: I don't think we should recommend that to admins. This will unnecessarily consume resources.

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
+      sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ files:scan --all
+
+We also recommend configuring the background job as **Webcron** or **Cron**
+(see :doc:`../configuration_server/background_jobs_configuration`) so that
+Nextcloud's regular housekeeping runs reliably alongside any custom scan cron job.
+
+Limitation: renames are not detected reliably
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using periodic rescanning (option B above), the scanner cannot reliably
```
artonge: Is it limited to periodic scans? Isn't it also an issue with the "Once per direct access" change detection setting?

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,116 @@ on your network!
...
+requirements and how to reduce sync delay.
+
+**B) All other storages — periodic rescan via cron**
+
+For storage types that do not support push notifications, set up a cron job that
+rescans the storage periodically:
+
+.. code-block:: console
+
+   sudo -E -u www-data php occ files:scan --path="/user/files/MountName"
```
artonge: We have a `files_external:scan <mount_id>` command nowadays, maybe we should recommend that one.

skjnldsv:
@artonge done in d4004874c0 !

- reworded both bullets (once per direct access + never) per your suggeestions
- dropped the trailing clause from the etag paragraph
- swapped `files:scan --all` and the `--path` variant for `files_external:scan <mount_id>`, also removed the AIO note and webcron recommendation
- broadened the rename limitation, its not just for periodic rescan but also for "once per direct access" as you pointed out

let me know if anything still looks off!

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -103,7 +103,12 @@ individually with the following options:
 * Encryption
 * Previews
 * Enable Sharing
-* Filesystem check frequency (Never, Once per direct access)
+* **Filesystem check frequency** — controls how the server rescans an external
+  storage path when a WebDAV ``PROPFIND`` request arrives for it.
+  **Once per direct access** rescans on the first ``PROPFIND`` per PHP request;
```
artonge: Is it possible to have multiple `PROPFIND` per PHP request?

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -157,24 +162,97 @@ on your network!
...
+   sudo -E -u www-data php occ files_external:scan <mount_id>
+
+See :doc:`../occ_command` for the full ``files_external:scan`` reference. A typical
+interval is every 15 minutes; adjust to balance freshness against server load.
+
+Limitation: renames are not detected reliably
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Whenever Nextcloud rescans an external storage — whether via periodic cron
+(option B) or on direct access (option A setting **Once per direct access**) —
```
artonge: Option A is: _"SMB/CIFS storages — use update notifications"_...

artonge (changes requested):
We should keep the AIO part, I assume Simon put it there because the question come frequently.

admin_manual/configuration_files/external_storage_configuration_gui.rst
skjnldsv: Simultaneously?
I don't think so, no 🤔

artonge: Then the sentence is false

skjnldsv:
Thanks for taking the time, I adjusted a few areas, and fixed the wrong listed option :rocket: 
Rebased and squashed!

admin_manual/configuration_files/external_storage_configuration_gui.rst
```diff
@@ -103,7 +103,12 @@ individually with the following options:
 * Encryption
 * Previews
 * Enable Sharing
-* Filesystem check frequency (Never, Once per direct access)
+* **Filesystem check frequency** — controls how the server rescans an external
+  storage path when a WebDAV ``PROPFIND`` request arrives for it.
+  **Once per direct access** rescans on each ``PROPFIND`` request;
```
artonge: To prevent any expectation that the whole storage will be scanned.

```suggestion
  **Once per direct access** makes a shallow rescans on each ``PROPFIND`` request;
```

skjnldsv: Done as well!

artonge (approved):

PR #15011, opened 2026-05-26
feat: Add grant step to login flow v2 diagram

artonge:
| Before | After |
|--------|--------|
| Screenshot From 2026-05-26 14-56-35 | Screenshot From 2026-05-26 14-56-08 |

### DIFF
--- a/developer_manual/client_apis/LoginFlow/index.rst
+++ b/developer_manual/client_apis/LoginFlow/index.rst
@@ -187,11 +187,14 @@ Note that the 200 will only be returned once.
                 end
 
                 User->>Browser: Enter credentials (and 2FA if required)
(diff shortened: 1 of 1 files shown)


printminion-co (approved):

printminion-co:
@artonge are there is 2 additional "preflight" calls (by apps) to /status.php and "capabilities" route before actual login call can be made.

artonge:
> are there is 2 additional "preflight" calls (by apps) to /status.php and "capabilities" route before actual login call can be made.

Those are not mandatory to the login flow.

- `status.php`: checks whether the server is healthy.
- `capabilies`: generic request to know what kind of operation the server can support

PR #15048, opened 2026-05-30
chore(admin): update ownCloud migration guide

MichaIng:
https://github.com/nextcloud/server/pull/59677 fixes direct migration from ownCloud to latest Nextcloud.

It does not change something about the fact that latest ownCloud 10 does not support any PHP version supported by Nextcloud. But while keeping PHP 7.4 during the migration to Nextcloud 25 makes it easier to revert to ownCloud in case of an issue, that path requires manually patching `version.php`, a lot of major version upgrades, and intermediate PHP upgrade steps. Overall the efforts are a lot higher than migration directly to latest Nextcloud, and upgrade PHP only once prior to that.

Another reason to prefer direct migration to latest Nextcloud is, that Nextcloud 25 migrations are not maintained anymore. There is hence a chance that some recent ownCloud 10 version introduces a change that causes a major or subtil issue during the migration, that is fixed only in Nextcloud versions which are not EOL.

The migration steps have been adjusted to reflect the direct migration path to the upcoming Nextcloud 34.0.0.

The table itself, to in case list different migration paths for different ownCloud versions, has been left in place, including the optional/conditional migration steps to further upgrade Nextcloud and in case PHP afterwards. However, to reduce maintenance efforts and simplify/clarify things, I'd vote for removing support for migration to any older Nextcloud versions entirely, and remove the target version and post-migration upgrade steps from the docs. Once latest Nextcloud removes support for migrations from ownCloud 10.13.x, it is probably better to simply remove this migration path from the docs, and require admins to upgrade ownCloud first. IMO, keeping support for (and documenting) migrations to Nextcloud versions, which themselves are not supported, and hence do not get possibly needed fixes for migration scripts and `version.php`, does not make sense.

Let me know if you agree to the last point, then I'll adjust the commit accordingly.

### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

Before:

{26547216-C39A-42B0-BF9B-B4612EF9AB19}

After:

{4F886D4D-9769-4F0E-A391-40C7CD6195BC}

_Would it make sense to split each ownCloud-Nextcloud version pair into their own table row, when keeping the table at all?_

### ✅ Checklist

- [x] I <s>have built</s> let github-actions build the documentation <s>locally</s> and reviewed the output (😅)

### DIFF
--- a/admin_manual/maintenance/migrating_owncloud.rst
+++ b/admin_manual/maintenance/migrating_owncloud.rst
@@ -15,48 +15,44 @@ See the table below for a version map, where migrating is easily possible:
 +-------------------+-------------------------------+
 | ownCloud          | Nextcloud                     |
 +===================+===============================+
-| 10.13.x           | 25.0.13                       |
-| 10.14.x           | 25.0.13                       |
-| 10.15.x           | 25.0.13                       |
-| 10.16.x           | 25.0.13                       |
+| 10.13.x           | 34.0.0                        |
+| 10.14.x           | 34.0.0                        |
+| 10.15.x           | 34.0.0                        |
+| 10.16.x           | 34.0.0                        |
 +-------------------+-------------------------------+
 
 .. note:: Since ownCloud does not and will not support PHP 8.0 or higher, you
-          need to migrate from ownCloud 10.13.x to Nextcloud 25 and then
-          further upgrade from there. We urge you to migrate to Nextcloud
-          since PHP versions prior PHP 8 are end of life, see
+          need to upgrade your PHP instance first, after which, ownCloud won't
+          be accessible anymore. See :doc:`../installation/system_requirements`
+          about PHP versions supported by Nextcloud. We urge you to migrate to
+          Nextcloud since PHP versions prior PHP 8 are end of life, see
           `<https://www.php.net/supported-versions.php>`_.
 
-1. First download the correct version of Nextcloud from our `older releases page <https://nextcloud.com/changelog/>`_,
+1. First download the correct version of Nextcloud from our `releases page <https://nextcloud.com/changelog/>`_,
 
 2. Make sure to have do a :doc:`backup<backup>` before migrating.
 
-3. Follow the upgrade instructions described in the :doc:`manual_upgrade` manual.
+3. Upgrade PHP to a version supported by the downloaded Nextcloud version.
 
-4. When migrating to Nextcloud 20.0 or later, you will also need to run the following commands after ``occ upgrade``:
+4. Follow the upgrade instructions described in the :doc:`manual_upgrade` manual.
+
+5. Run the following commands after ``occ upgrade``:
 
   * ``occ db:convert-filecache-bigint``
   * ``occ db:add-missing-columns``
   * ``occ db:add-missing-indices``
   * ``occ db:add-missing-primary-keys``
 
-5. If system cron was used, please verify if crontab entry was using the command ``occ system:cron``.
+6. If system cron was used, please verify if crontab entry was using the command ``occ system:cron``.
    If yes, please adjust it to use the ``php`` command instead according to :ref:`the background jobs configuration documentation<system-cron-configuration-label>`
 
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):
Looks good, I'll wait for the other to check as well! Thanks !! :hug:

blizzz:
Looks good and makes sense. I do not have experience from a recent oc upgrade though. 

35.0.0 would be correct for the master branch, as 34.0.0 was branched out to `stable34`.

skjnldsv:
> Looks good and makes sense. I do not have experience from a recent oc upgrade though.
> 
> 35.0.0 would be correct for the master branch, as 34.0.0 was branched out to `stable34`.

I think not, we cannot vauht for 35 compatibility YET, this is not bound to our stable relzase process. I think it should stay 34 int his doc

MichaIng:
What do you think about my last point/suggestion to not define individual Nextcloud versions at all, but suggest to use the latest (at least a still supported one) instead, removing all the subsequent "upgrade major Nextcloud/PHP versions" steps? So there wouldn't be a need to update the table, latest Nextcloud is assumed to support all documented ownCloud origin versions. And once support for e.g. ownCloud 10.13 is dropped, it is removed from the table.

skjnldsv:
> not define individual Nextcloud versions at all, but suggest to use the latest (at least a still supported one) instead

Feels error-prone. We've had many of our releases needing work to make the oc-nc migration viable. We're very far from what owncloud is nowadays. I'm not sure we should take the "try and find out" approach, especially with migrations and potential data loss if admin don't do things carefully. :thinking:

MichaIng:
Of course it makes sense to keep testing it with major Nextcloud releases, and as well with new ownCloud releases, like if there will be a v10.17.y. And if migration from v10.13 breaks at some point a way that is not seen worth the efforts to fix, support can be removed via `version.php`. What I basically would aim to avoid is to support migrations to unsupported/EOL Nextcloud versions, for which things cannot be guaranteed well either, fixes cannot be shipped anymore etc. I mean the potential risk of data loss is on both ends, backups inevitable.

I am just thinking about setting up a GitHub Actions workflow with sets up the supported ownCloud versions, maybe with some additional data and apps via `occ`, does a number of checks via `occ` and `curl`, then follows the documented steps to migrate to HEAD branch version of Nextcloud, and repeats the checks 😄. Should not be that hard, given that can copy&paste setup steps from DietPi. And maybe there is even a reasonable way to pipe the result though some existing Nextcloud CI workflows?

skjnldsv:
@MichaIng if you wanna work on such tests, have a look at how we do upgrades tests on the updater server repository 😊

MichaIng:
@skjnldsv 
You mean https://github.com/nextcloud/updater/tree/master/tests? This is for the `updater.phar` "file replacer" part only, not `occ ugprade` database migration particularly, right? So while what `updater.phar` does would need to be done manually, the post-update `occ` health and version checks done in these tests could be reused.

skjnldsv:
> You mean [nextcloud/updater@`master`/tests](https://github.com/nextcloud/updater/tree/master/tests?rgh-link-date=2026-06-14T13%3A49%3A29Z)?

No, the updater_server repo :)

MichaIng:
Ah, in the dedicated https://github.com/nextcloud-releases orga. Yeah these look very good, only an additional intermediate `setup-php` job needed, and adjusted ownCloud URLs and `occ` commands, of course. I wonder whether the Nextcloud `updater.phar` works as well (will test anyway when creating these workflows 😄)?

I guess such tests should be added to https://github.com/nextcloud/server, where database upgrade and migration code lives. But how to build a production-ready Nextcloud instance from the local Git repo, where `occ upgrade` including integrity checks run through? I see https://github.com/nextcloud-releases/server/blob/main/.github/workflows/release-build.yml, but maybe there is a minimal subset of steps, or a way to invoke the build+cleanup steps from the releases repo, to not add a large copy and maintenance burden?

binford6000:
From my recent oC migrations this is what we've already done with our customers.

skjnldsv:
Thank you, merging then!!

skjnldsv:
Thanks again @MichaIng ! Always a pleasure seeing you around after all those years :hugs:

PR #15053, opened 2026-06-01
docs(exapps_management): document ExApp container log rotation

### LINKED ISSUE #857: Log rotation for ExApp Container
### Describe why it is important and where it will be useful

In our setup we had the issue that the log of the flow container kept growing and growing up to a 100 GB. 
To prevent this is is needed to manually set log rotations for docker in the docker configuration. 

Adding the log rotation automatically via the deploy would help to prevent this situation by default and would help users that are not aware of this. 

### Describe your proposed solution

It looks like the container creation API of docker might support setting the log rotation conf. (https://docs.docker.com/reference/api/engine/version/v1.54/#tag/Container/operation/ContainerCreate) 

Not sure if this is the best solution. 

### Describe alternatives you've considered, if relevant

Seting up a global log rotation configuration. 

### Additional context

_No response_

oleksandr-nc:
### ☑️ Resolves

* Fixes: https://github.com/nextcloud/app_api/issues/857

### 🖼️ Screenshots

Screenshot From 2026-06-01 13-46-40
Screenshot From 2026-06-01 13-47-11

### ✅ Checklist

### DIFF
--- a/admin_manual/exapps_management/DeployConfigurations.rst
+++ b/admin_manual/exapps_management/DeployConfigurations.rst
@@ -568,6 +568,72 @@ Docker Socket Proxy security
 AIO Docker Socket Proxy has strictly limited access to the Docker APIs described in `HAProxy configuration <https://github.com/nextcloud/all-in-one/blob/main/Containers/docker-socket-proxy/haproxy.cfg>`_.
 
 
+.. _exapp-container-log-rotation:
+
+Container log rotation
+----------------------
+
+AppAPI deploys each ExApp as a container but does not set a logging configuration on it, so each ExApp container
+uses the logging settings of the deploy daemon that runs it. If those logs have no size limit they are never
+rotated and can grow until they fill the host disk. Configure rotation on the daemon, where it applies to every
+container the daemon runs.
+
+Docker
+^^^^^^
+
+Docker's default ``json-file`` driver sets no size limit. Add the following to the Docker daemon configuration file
+(:file:`/etc/docker/daemon.json` on most Linux hosts) and restart the daemon:
+
+.. code-block:: json
+
+    {
+        "log-driver": "json-file",
+        "log-opts": {
+            "max-size": "10m",
+            "max-file": "3"
+        }
+    }
+
+.. code-block:: bash
+
+    sudo systemctl restart docker
+
+Each container then keeps at most ``max-file`` log files of ``max-size`` each (here three files of 10 MB, so up to
+30 MB per container). Adjust the values to suit your hosts.
+
+.. note::
+
+    A container keeps the logging configuration it was created with. Re-deploy the affected ExApps (disable and enable
+    them again, or remove and install) so their containers are recreated with the new settings.
+
+.. important::
+
+    Keep the ``json-file`` or ``journald`` driver: AppAPI can only download container logs for these two drivers, so a
+    different driver disables log downloads for ExApp containers.
+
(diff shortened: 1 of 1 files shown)


admin_manual/exapps_management/DeployConfigurations.rst
```diff
@@ -568,6 +568,72 @@ Docker Socket Proxy security
...
+^^^^^^
+
+When the deploy daemon is Podman, set a global default in :file:`/etc/containers/containers.conf` (or the per-user
+file) and re-deploy the affected ExApps:
+
+.. code-block:: toml
+
+    [containers]
+    log_driver = "json-file"
+    log_size_max = 10485760
```
kyteinsky, line 628: `log_rotate_max=3` this too maybe

admin_manual/exapps_management/DeployConfigurations.rst
```diff
@@ -568,6 +568,72 @@ Docker Socket Proxy security
...
+
+    sudo systemctl restart docker
+
+Each container then keeps at most ``max-file`` log files of ``max-size`` each (here three files of 10 MB, so up to
+30 MB per container). Adjust the values to suit your hosts.
+
+.. note::
+
+    A container keeps the logging configuration it was created with. Re-deploy the affected ExApps (disable and enable
+    them again, or remove and install) so their containers are recreated with the new settings.
```
kyteinsky: is disable+enable enough, or maybe only container re-creation (remove+install) would change the logging config?

admin_manual/exapps_management/DeployConfigurations.rst
kyteinsky: also podman should be restarted
```
sudo systemctl restart podman
```

does kubernetes need a restart?

oleksandr-nc: Podman's `containers.conf` has no `log_rotate_max` (or any multi-file retention) key. The available log options are `log_driver`, `log_path`, `log_size_max`, and `log_tag`. `log_size_max` only caps the single log file; there is no `max-file`-style retention like Docker's. So I kept the snippet with just `log_size_max`.

admin_manual/exapps_management/DeployConfigurations.rst
oleksandr-nc: Good catch, you're right. Verified in AppAPI (`AppAPIService.php`): `disableExApp` calls `stopContainer` and `enableExApp` calls `startContainer`, so it only stops/starts the existing container and wouldn't pick up the new logging config. Updated the note to require remove + install (re-creation) instead. Fixed in 4811065.

admin_manual/exapps_management/DeployConfigurations.rst
oleksandr-nc: Two parts here:

- Podman: a `systemctl restart podman` isn't needed. Unlike `dockerd` (a long-running daemon that caches `daemon.json`), Podman reads `containers.conf` at container-creation time, so re-creating the ExApp container is what applies the change. I left the restart out and made the re-create requirement explicit instead.
- Kubernetes: yes, it needs a restart: the kubelet loads its config only at startup, so `containerLogMaxSize`/`containerLogMaxFiles` take effect after `systemctl restart kubelet` on each node. Added that to the section. Fixed in 4811065.

kyteinsky: ah yeah, `containers.conf` indeed does not have that option, seems like I saw some AI generated blog :slightly_smiling_face: 

perfect, thanks!

kyteinsky (approved):
:rocket:

PR #15119, opened 2026-06-12
Document PHP version and official Linux distribution support changes

susnux:
### ☑️ Resolves

* for https://github.com/nextcloud/server/pull/61142

### 🖼️ Screenshots

grafik

grafik

grafik

### ✅ Checklist

- [ ] I have run `codespell` or similar and addressed any spelling issues

### DIFF
--- a/developer_manual/release_notes/critical_changes.rst
+++ b/developer_manual/release_notes/critical_changes.rst
@@ -24,6 +24,11 @@ Update info.xml to add Nextcloud 35 to the support range:
 
 To allow installation on older versions too, just keep the previous min-version.
 
+Updated PHP requirements
+------------------------
+
+The support for PHP 8.2 has been dropped, the minimum supported PHP version of Nextcloud 35 is now 8.3.
+
 Removed front-end APIs and libraries
 ------------------------------------
 
--- a/admin_manual/installation/system_requirements.rst
+++ b/admin_manual/installation/system_requirements.rst
@@ -16,7 +16,6 @@ For best performance, stability and functionality we have documented some recomm
 +==================+=======================================================================+
 | Operating System | - **Ubuntu 26.04 LTS** (recommended)                                  |
 | (64-bit)         | - Ubuntu 24.04 LTS                                                    |
-|                  | - Ubuntu 22.04 LTS                                                    |
 |                  | - **Red Hat Enterprise Linux 10** (recommended)                       |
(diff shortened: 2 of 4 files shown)


developer_manual/release_notes/critical_changes.rst
```diff
@@ -24,6 +24,11 @@ Update info.xml to add Nextcloud 35 to the support range:
 
 To allow installation on older versions too, just keep the previous min-version.
 
+Updated PHP requirements
+------------------------
+
+The support for PHP 8.2 has been dropped, the minimum supported PHP version of Nextcloud 35 is now 8.3.
```
ChristophWurst, line 30: nit: add info.xml example

ChristophWurst (approved):

skjnldsv (approved):

developer_manual/release_notes/critical_changes.rst
susnux: Oh just saw it, but what kind of appinfo example could we do for PHP versions?

ChristophWurst: Like this: https://github.com/nextcloud/documentation/blob/cb7b633f982bc1147766ab131c01540411d04634/developer_manual/release_notes/previous/upgrade_to_28.rst?plain=1#L60-L72

susnux: Ah ok.
Not sure it makes sense for min-php? Because I can imagine this confuses devs to think they have to set it.
But its not needed - except you support a wide range of Nextcloud versions but not PHP versions.
E.g. Nextcloud 30-34 but not PHP 8.1 (which would be supported in 30).

ChristophWurst: It could be worded so that devs only touch if if they have that. But fair enough, we can leave this out

MaxXor:
Hi, was nextcloud 34 skipped on purpose for the critical changes section? Are there none?

susnux:
@MaxXor you mean the developer docs for 34? There are critical changes documented:

https://docs.nextcloud.com/server/stable/developer_manual/release_notes/critical_changes.html#critical-changes

If you mean admin docs, then no there are no critical changes for 34.

MaxXor:
Thanks for clarifying, yes I meant the admin docs.

PR #15140, opened 2026-06-15
fix: show numbered version in manual titles and version picker

skjnldsv:
### ☑️ Resolves

Closes #14891 (partial - addresses the version picker label clarity)

### What changed and why

On the master branch, `version = 'latest'` caused all three manuals to display "Nextcloud latest User Manual" in titles, headings, the version picker, and PDF covers. That's ambiguous - readers can't tell which numbered release "latest" maps to.

This PR introduces `display_version` in `conf.py`, derived automatically from `version_stable + 1`. It's currently `35` (NC35 dev). When NC35 ships and `version_stable` bumps to 35, the display updates to 36 with no other changes needed. `version = 'latest'` is kept intact so URL rewriting, branch logic, and `html_baseurl` are unaffected.

The version picker also gets clearer labels: `34 (stable)` and `35 (latest)` instead of bare `stable` / `latest`. The collapsed picker button shows the number too. And since the stable version was appearing twice in the list, the numbered range now stops just before `version_stable`.

PDF covers on master builds used to say "Release latest". They now derive the version the same way - `version_stable + 1` from `conf.py`.

On the CI side, this also introduces a validation step that fails the `build-html` job (master and PRs targeting master only) if `version_stable` or `version_start` in `conf.py` have drifted from the actual remote branch state. The detection logic that was previously duplicated in `sphinxbuild.yml` and `build-index.php` is now in a single `build/detect-versions.php` script that both call. It checks the GitHub API to confirm a `v{N}.0.0` release actually exists before accepting a branch as "stable" - so an RC branch won't be counted prematurely. The lowest supported version is also derived from the support window (released within the last year) rather than the raw minimum branch number, which would otherwise include ancient branches like stable10.

### 🖼️ Screenshots

imageimage
image

### ✅ Checklist

### DIFF
--- a/.github/workflows/sphinxbuild.yml
+++ b/.github/workflows/sphinxbuild.yml
@@ -60,6 +60,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Validate version constants in conf.py
+        if: github.ref == 'refs/heads/master' || github.base_ref == 'master'
         run: |
           highest_stable=$(git ls-remote --heads origin \
             | sed -n 's?.*refs/heads/stable\([0-9]\{2\}\)$?\1?p' \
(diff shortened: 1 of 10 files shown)


ChristophWurst (approved):
🐘

PR #15267, opened 2026-07-02
fix: formatting in PHP parameters documentation

skjnldsv:
See https://github.com/nextcloud/server/pull/61719


rakekniven:
@skjnldsv This file is autogenerated.

See comment at https://github.com/nextcloud/server/blob/eb864139108d7611a206bbf244585d1cc9b2cad8/config/config.sample.php#L18

rakekniven (changes requested):
See my comment. Hopefully I am right 🤞 .

skjnldsv:
Yes, but until the cron job starts again, this will basically be doing the same fix as what the cron job will do tonight :)

rakekniven (approved):

PR #15317, opened 2026-07-15
fix(l10n): normalize de_DE to de before user manual build

### LINKED ISSUE #13370: Move formal german user documentation from "de" to "de_DE"
Actually the server and all apps (desktop and mobile as well) have the formal german translation at language code "de_DE".

The user documentation uses code "de" for the formal version (there is no informal version for the user docs).

This confuses translators at Transifex. Not every translator is member of both german teams at Transifex.

Target is to use "de_DE" for user docs as well.

**Benefit**
* It is clear to the translators that "de_DE" is the formal version.
* Team member are not confused about the mix of de and "de_DE"
* The translation memory (TM) is clean and mature, which means fewer errors occur during automatic matching.

Possible solutions:
* Shift translations at Tx from "de" to "de_DE" and move files to "de" before Sphinx compile.
* change transifex sync to map from de_DE to de -> See comments from @nickvergessen 
I'm not sure if all possibilities have been tried here (due to time constraints)
* ... ?

miaulalala:
### ☑️ Resolves

Relates to #13370

### 💡 Context

Every other Nextcloud project (server, apps, clients) syncs formal German as `de_DE` in Transifex. These docs are the outlier, using bare `de` — which confuses translators (not everyone is on both German teams) and keeps the translation memory from pooling with the mature `de_DE` TM used everywhere else.

The published URL (`user_manual/de/`) can't change — it's linked from many external sites (see discussion on the issue).

This PR decouples the two: it adds a build-time step (`build/normalize_locale_codes.sh`, wired into `user_manual/Makefile`'s `html` target) that folds `locale/de_DE` into `locale/de` right before Sphinx builds, if `de_DE` exists. It follows the same pattern already used by `change_file_extension.sh` for `.pot` → `.po` renaming.

**This is currently a no-op** — `locale/de_DE` doesn't exist yet. It only takes effect once the Transifex resource for the user manual is reconfigured to sync German under `de_DE` instead of `de`, which needs to happen on the Transifex side (cc @rakekniven).

Opening as a draft to get eyes on the approach before the Transifex-side config change happens — cc @AndyScherzinger @nickvergessen since you both weighed in on the issue.

### 🖼️ Screenshots

N/A — build tooling change only, no rendered content changes.

### ✅ Checklist

### DIFF
--- a/build/normalize_locale_codes.sh
+++ b/build/normalize_locale_codes.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+#
+# Nextcloud server, apps and clients use "de_DE" for formal German in Transifex.
+# These docs historically used "de", causing translator confusion and a fragmented
+# translation memory (see https://github.com/nextcloud/documentation/issues/13370).
+#
+# Transifex is being reconfigured to sync German as "de_DE" like every other
+# Nextcloud project, but the published docs URL must stay "de" (it's linked from
+# many external sites). This folds "de_DE" into "de" right before Sphinx builds,
+# so the sync stays de_DE-native while the published path is unaffected.
+#
+# No-op until locale/de_DE actually exists.
+
+if [ -d ./locale/de_DE ]; then
+    mkdir -p ./locale/de
+    rsync -a ./locale/de_DE/ ./locale/de/
+    rm -rf ./locale/de_DE
+fi
--- a/user_manual/Makefile
+++ b/user_manual/Makefile
@@ -49,6 +49,7 @@ clean:
 
 html:
 	../build/change_file_extension.sh
+	../build/normalize_locale_codes.sh
 	make html-all merge-folders
 
 html-all: $(foreach lang, $(LANGS), html-allow-warnings-lang-$(lang))


nickvergessen:
I'm overly busy and not familiar with the new deployment mechanism, so I will stay out of this

rakekniven:
Oh, that sounds good.
I’d be happy to write an announcement at Transifex before the change goes live.

Renaming a language at Transifex is not possible.
I can add "de_DE" language, which will be empty.
The sync needs to be adapted as well and after a sync the strings should be present in "de_DE".
Afterwards I will remove "de" language from project "Nextcloud user documentation" at Tx.

Question:
Where are the available languages for the documentation defined? In Sphinx, it should only be “de”, and on Transifex, “de_DE”.

miaulalala:
Good question! There's no static list to edit — the available languages are derived dynamically from directory names in both places:

- **Sphinx build** (`user_manual/Makefile`): `LANGS = $(shell touch locale/en && ls locale | grep -v source && rm locale/en)` — just lists whatever folders exist under `user_manual/locale/`
- **Language switcher template** (`user_manual/conf.py`, `available_languages`): same idea, `os.listdir(locale_path)` over `user_manual/locale/`

So as long as `build/normalize_locale_codes.sh` runs before Sphinx builds (it does, wired into `user_manual/Makefile`'s `html` target), it folds `locale/de_DE` into `locale/de` before Sphinx ever sees the directory listing. Sphinx and the language switcher will only ever see `de`.

On the Transifex side: once you add `de_DE` and the sync lands those files in the repo under `user_manual/locale/de_DE/`, our build script takes it from there — no further config needed on this end.

AndyScherzinger (approved):
Code looks good to me, but I have no knowledge myself on how generation and deployment works on the docs repo. So I'd be relying on @skjnldsv and @nickvergessen for that.

But generally would be happy if we can make the translation community happy with this.

PR #15320, opened 2026-07-16
feat(activity): Adding documentation for new user/event filter option

Aiiaiiio:
### ☑️ Resolves

Documentation for Activity's new config option.

### ✅ Checklist

### DIFF
--- a/admin_manual/configuration_server/activity_configuration.rst
+++ b/admin_manual/configuration_server/activity_configuration.rst
@@ -89,6 +89,10 @@ The following ``config.php`` options control Activity app behavior.
      - ``[]``
      - An array of user IDs whose activity records are never deleted
        by the expiration job. See :ref:`label-activities-exclude-users`.
+    * - ``activity_log_exclude_users``
+     - ``[]``
+     - An associative array of [ user_ID => event_types_array ]. For each user the
+       associated event types won't be logged. See :ref:`_label-activity_log_exclude_users`.
 
 
 .. _label-activities-groupfolders:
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_server/activity_configuration.rst
```diff
@@ -146,6 +150,22 @@ data, for example administrators. Set the config value
...
+
+Excluding event types for users
+----------------------------------------
+
+For certain users, it might make sense to exclude some event types from
+being logged. For example set the config value
+``activity_log_exclude_users`` in your ``config.php``::
+
+  'activity_log_exclude_users' => [
+    'alice' => ['file_created'],
```
miaulalala, line 164: not sure if there is a reference for event types already - if there is, please link it somewhere here

miaulalala (changes requested):

admin_manual/configuration_server/activity_configuration.rst
Aiiaiiio: There isn't one. At least not that I know of. I'll try and collect them. I wonder how we will make sure this is maintained though.

miaulalala: If there is no such reference, then this is fine. 

Unsure if the copilot comments need addressing?

miaulalala:
Thanks for adding this! I cross-checked against `nextcloud/activity`'s `lib/Data.php` (`isExcludedAuthor()`) and the config key/shape (`'user' => ['event_type', ...]`) matches exactly.

One wording nit though: the phrasing "For certain users, it might make sense to exclude some event types from being logged... For these users and event types a log record won't be created" reads as if this filters what shows up in *that user's own* activity feed.

Looking at the implementation, the filter is actually keyed on the event **author** (`$event->getAuthor()`), and it suppresses the activity record entirely — for every affected user's stream, not just the configured user's own. So `'alice' => ['file_created']` means "when alice creates a file, no activity entry is logged for anyone" (e.g. it also disappears from a shared folder's activity for other members) — not "alice stops seeing file_created events in her own feed."

Could we tweak the wording to make that clearer, something like:

> For certain users, it might make sense to suppress activity log entries for actions *they perform*, for specific event types — for example to reduce noise from automated/service accounts. Set the config value ``activity_log_exclude_users`` in your ``config.php``::
>
>   'activity_log_exclude_users' => [
>     'alice' => ['file_created'],
>     'bob' => ['file_created', 'file_changed'],
>   ]
>
> When one of these users performs one of the listed event types, no activity log record is created at all — including in other users' streams (e.g. shared folder activity).

Just want to make sure admins don't configure this expecting per-viewer filtering when it's actually a global suppression based on who performed the action.

miaulalala:
@Aiiaiiio can you check my comment and implement what makes sense?

PR #15321, opened 2026-07-16
docs: Clarify app icon instructions, add link to Material Symbols & Icons

jancborchardt:
### ☑️ Resolves

Makes it more clear what to use for app icons. Ref @miaulalala 

### ✅ Checklist

### DIFF
--- a/developer_manual/design/foundations.rst
+++ b/developer_manual/design/foundations.rst
@@ -185,8 +185,9 @@ Element sizes and spacing
 Icons
 -----
 
-Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use Material Symbols (outlined, 20px).
+Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use `Material Symbols <https://fonts.google.com/icons>`__ (outlined, 20px).
 
--  Never use a custom icon except for the icon of the app itself. Still, most apps use an app icon from Material Symbols to keep consistency.
+-  Never use a custom icon except for the icon of the app itself.
 -  When possible, use text together with icons so it’s clear what they refer to.
+-  For the main app icon, you can use a custom icon or just pick a fitting one from `Material Symbols <https://fonts.google.com/icons>`__ to keep consistency. While the interface icons should be outlined, we recommend using the filled version for the main app icon so it stands out better, also on the header background which can be an image and/or themed.
 -  Elements that contain AI-generated content should have a clear indication that they are AI-generated in the form of a written disclaimer or by the use of the Assistant icon ('creation' in Material Symbols) in the Assistant color.


developer_manual/design/foundations.rst
```diff
@@ -185,8 +185,9 @@ Element sizes and spacing
 Icons
 -----
 
-Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use Material Symbols (outlined, 20px).
+Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use `Material Symbols <https://fonts.google.com/icons>`__ (outlined, 20px).
 
--  Never use a custom icon except for the icon of the app itself. Still, most apps use an app icon from Material Symbols to keep consistency.
+-  Never use a custom icon except for the icon of the app itself.
 -  When possible, use text together with icons so it’s clear what they refer to.
+-  For the main app icon, you can use a custom icon or just pick a fitting one from `Material Symbols <https://fonts.google.com/icons>`__ to keep consistency. While the interface icons should be outlined, we recommend using the filled version for the main app icon so it stands out better, also on the header background which can be an image and/or themed.
```
miaulalala, line 192: are there also some recommendations for custom app icons? Should they also be filled versions? Or should the provide more than one variant (light, dark)?

miaulalala (changes requested):

developer_manual/design/foundations.rst
miaulalala: @jancborchardt

jancborchardt: No recommendations (yet). Usually people would simply use their logo, like e.g. for Zammad. I'm not sure if technically the white monochrome version suffices. @susnux @ShGKme do you know?

Either way I would suggest merging this, as we can add any custom app icon clarifications afterwards and it's not a blocker.

ShGKme: > Or should the provide more than one variant (light, dark)?

We invert the icon automatically on dark.
But it must be a white monochrome and invertible by design.

> are there also some recommendations for custom app icons?

I'd say:
- Be monochrome white-ish
- Be invertible (it looks fine and valid when inverted)
- Allow a gradient effect on top (some brand logos forbid any modifications)

> Usually people would simply use their logo, like e.g. for Zammad. I'm not sure if technically the white monochrome version suffices

I'll double-check if we support anything like a "no-modifications" flag on the app icon.

developer_manual/design/foundations.rst
```diff
@@ -185,8 +185,9 @@ Element sizes and spacing
 Icons
 -----
 
-Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use Material Symbols (outlined, 20px).
+Icons can be used to communicate the intent of an action, or to provide visual interest to an element. On the browser always use `Material Symbols <https://fonts.google.com/icons>`__ (outlined, 20px).
```
ShGKme, line 188: Currently we use `Material Design Icons`, not `Symbols` in the most of the places. Just because 2 libraries we are using provide them.

Though, many icons are identical.

jancborchardt: Ok, but this is not anything new in this change? I'm a bit hesitant to link to Pictogrammers as it's non-standard.

ShGKme: There is now a link to the font, which, IMO, might be confusing (one of the differences between Material Design Symbols and Material Design Icons was targeting variable fonts).

Maybe clarity that SVG is to be used?

developer_manual/design/foundations.rst
ShGKme: > I'll double-check if we support anything like a "no-modifications" flag on the app icon.

Nothing on the server side to mark an icon as "modifications not allowed"...

ShGKme: > We invert the icon automatically on dark.
> But it must be a white monochrome and invertible by design.

Here I wasn't entirely correct. We do invert automatically on the web frontend. However, we have some places on the server-side where we expect a direct link to the dark and light icons. So, both are needed.

PR #15384, opened 2026-07-28
docs(profile): add impact of preset feature on visibility defaults

joshtrichards:
### ☑️ Resolves

Documents bare essentials of the impact of presets added in nextcloud/server#54158 on Profile visibility defaults.

### ✅ Checklist

### DIFF
--- a/admin_manual/configuration_user/profile_configuration.rst
+++ b/admin_manual/configuration_user/profile_configuration.rst
@@ -82,11 +82,21 @@ These correspond to the visibility options in **Personal settings** -> **Persona
 Defaults
 ^^^^^^^^
 
-By default, most profile fields are configured as **Show to everyone**, while some
-contact-related fields default to **Show to logged in accounts only**.
+Most profile fields default to **Show to everyone**, while some contact-related fields
+(e.g. address, email, and phone) default to **Show to logged in accounts only**.
 
 Administrators should note that these defaults are independent from the default
 *property scopes* described below.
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_user/profile_configuration.rst
```diff
@@ -82,12 +82,22 @@ These correspond to the visibility options in **Personal settings** -> **Persona
 Defaults
 ^^^^^^^^
 
-By default, most profile fields are configured as **Show to everyone**, while some
-contact-related fields default to **Show to logged in accounts only**.
+Most profile fields default to **Show to everyone**, while some contact-related fields
+(e.g. address, email, and phone) default to **Show to logged in accounts only**.
```
miaulalala, line 86: Is there a way we could have a list of default profile fields and their settings? Not a blocker, maybe a follow up even.

miaulalala (approved):

admin_manual/configuration_user/profile_configuration.rst
susnux: lets do that in a follow up

joshtrichards: I'm feeling like this entire chapter has a lot of the key info collected together now, but the next major improvement (or series of improvements) should be to organize/present/explain it all better. 

I wrote a lot of the latest revisions and I still find it confusing AF. ;)

PR #15414, opened 2026-08-05
docs(office): document all env vars and add size limits guide

chrip:
Document every environment variable supported by the Document Server Docker image in the installation page. Previously only a subset was listed.

Add a 'Size limits' section with a 200 MB PPTX example showing how the four size-gate variables interact at different stages (nginx upload, temp file buffer, converter download, uncompressed XML size).

Related to https://github.com/Euro-Office/DocumentServer/pull/237 and https://github.com/Euro-Office/server/pull/35

Assisted-by: OpenCode:qwen3.6-27b

### DIFF
--- a/admin_manual/office/euro-office/installation_docker.rst
+++ b/admin_manual/office/euro-office/installation_docker.rst
@@ -92,33 +92,161 @@ Environment variables
     * - ``JWT_HEADER``
       - ``Authorization``
       - HTTP header carrying the JWT
-    * - ``WOPI_ENABLED``
+    * - ``JWT_HEADER_INBOX`` / ``JWT_HEADER_OUTBOX``
+      - ``JWT_HEADER``
+      - Separate headers per direction
+    * - ``JWT_IN_BODY``
       - ``false``
-      - Enable WOPI protocol support
-    * - ``ALLOW_PRIVATE_IP_ADDRESS``
-      - ``false``
-      - Allow the Document Server to fetch files from private IP ranges
-    * - ``NGINX_WORKER_PROCESSES``
-      - ``1``
-      - Number of nginx worker processes
-    * - ``GENERATE_FONTS``
-      - ``true``
-      - Regenerate font cache on startup
+      - Accept the token in the request body
+    * - ``JWT_SECRET_INBOX`` / ``JWT_SECRET_OUTBOX``
+      - ``JWT_SECRET``
+      - Separate secrets per direction
+    * - ``JWT_ENABLED_INBOX`` / ``JWT_ENABLED_OUTBOX``
+      - ``JWT_ENABLED``
+      - Enable JWT per direction
+    * - ``DB_TYPE``
+      - ``postgres``
+      - Database engine (standalone image supports ``postgres`` only)
     * - ``DB_HOST``
       - ``localhost``
-      - PostgreSQL host (for external database)
+      - Database host
+    * - ``DB_PORT``
+      - ``5432``
+      - Database port
     * - ``DB_NAME``
       - ``eurooffice``
-      - PostgreSQL database name
+      - Database name
     * - ``DB_USER``
       - ``eurooffice``
-      - PostgreSQL user
+      - Database user
+    * - ``DB_PWD``
+      - —
+      - Database password
     * - ``REDIS_SERVER_HOST``
       - ``localhost``
       - Redis host (for external Redis)
+    * - ``REDIS_SERVER_PORT``
+      - ``6379``
+      - Redis port
+    * - ``REDIS_SERVER_PASS``
+      - —
+      - Redis password
     * - ``AMQP_HOST``
       - ``localhost``
       - RabbitMQ host (for external RabbitMQ)
+    * - ``AMQP_PORT``
+      - ``5672``
+      - RabbitMQ port
+    * - ``AMQP_USER`` / ``AMQP_PWD``
+      - ``guest``
+      - RabbitMQ credentials
+    * - ``WOPI_ENABLED``
+      - ``false``
+      - Enable WOPI protocol support
+    * - ``PLUGINS_ENABLED``
+      - ``true``
+      - Enable editor plugins
+    * - ``GENERATE_FONTS``
+      - ``true``
+      - Regenerate font cache on startup
+    * - ``METRICS_ENABLED``
+      - ``false``
+      - Enable StatsD metrics collection
+    * - ``ALLOW_PRIVATE_IP_ADDRESS``
+      - ``false``
+      - Allow the Document Server to fetch files from private IP ranges
+    * - ``ALLOW_META_IP_ADDRESS``
+      - ``false``
+      - Allow fetching documents from meta-private IPs (169.254.0.0/16)
+    * - ``USE_UNAUTHORIZED_STORAGE``
+      - ``false``
+      - Allow fetching documents from HTTP (non-TLS) storage
+    * - ``SSL_VERIFY_CLIENT``
+      - ``off``
+      - Enable SSL client certificate verification
+    * - ``ONLYOFFICE_HTTPS_HSTS_ENABLED``
+      - ``true``
+      - Enable HSTS headers
+    * - ``ONLYOFFICE_HTTPS_HSTS_MAXAGE``
+      - ``31536000``
+      - HSTS max-age in seconds (default 1 year)
+    * - ``NGINX_WORKER_PROCESSES``
+      - ``1``
+      - Number of nginx worker processes
+    * - ``NGINX_ACCESS_LOG``
+      - ``false``
+      - Enable nginx access logging
+    * - ``NGINX_CLIENT_MAX_BODY_SIZE``
+      - ``100m``
+      - Nginx client max body size (upload limit for nginx)
+    * - ``MAX_FILE_SIZE``
+      - ``104857600``
+      - Max temp file upload size in bytes (default 100 MB)
+    * - ``FILECONVERTER_MAX_DOWNLOAD_BYTES``
+      - ``524288000``
+      - Max file download size for the FileConverter in bytes (default 500 MB)
+    * - ``FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED``
+      - ``500MB``
+      - Max uncompressed zip size for office files (docx, xlsx, pptx, vsdx)
+
+Size limits
+-----------
+
+Each limit guards a different stage of the file lifecycle. Imagine a user tries to
+open a **200 MB** ``.pptx`` file:
+
+**1. Nginx accepts the upload** — ``NGINX_CLIENT_MAX_BODY_SIZE`` must be higher than
+the file. Default is ``100m``. With a 200 MB file the user gets
+``413 Request Entity Too Large``. Set it to ``250m``:
+
+.. code-block:: bash
+
+    -e NGINX_CLIENT_MAX_BODY_SIZE=250m
+
+**2. Document Server temp file** — ``MAX_FILE_SIZE`` gates the internal upload buffer
+(bytes). Default is ``104857600`` (100 MB). A 200 MB file fails here too. Set it to
+``268435456`` (256 MB):
+
+.. code-block:: bash
+
+    -e MAX_FILE_SIZE=268435456
+
+**3. FileConverter downloads the file** — ``FILECONVERTER_MAX_DOWNLOAD_BYTES`` is the
+max bytes the converter will fetch (default ``524288000`` = 500 MB). Already
+sufficient for a 200 MB file. No change needed.
+
+**4. FileConverter unzips the archive** — ``FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED``
+checks the internal XML size. A **200 MB** ``.pptx`` on disk might contain 800 MB of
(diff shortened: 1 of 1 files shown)


moodyjmz (changes requested):
**TL;DR:** Most of the new env-var table checks out against the entrypoint script and the two upstream PRs it's based on, but the new "Size limits" walkthrough gets the request flow wrong for its own headline scenario (opening a stored document), and `USE_UNAUTHORIZED_STORAGE` is mis-described in a way that understates its security impact. Requesting changes on those two before merge; the completeness gaps below can be a fast follow.

<details>
<summary>Full review, with exact lines</summary>

### 1. "Size limits" section's core narrative is backwards

[`installation_docker.rst#L192-L228`](https://github.com/nextcloud/documentation/blob/1eba328af5065315232497fcf90e2bfab3b00a6e/admin_manual/office/euro-office/installation_docker.rst#L192-L228)

Opening a document already stored in Nextcloud is a server-side **download**, not a browser upload:

- `canvasService.downloadFile` ([`canvasservice.js#L1744`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/DocService/sources/canvasservice.js#L1744)) is what actually fetches the document. It reads `FILECONVERTER_MAX_DOWNLOAD_BYTES` at [`L1771`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/DocService/sources/canvasservice.js#L1771) and passes it into `utils.downloadUrlPromise` at [`L1849`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/DocService/sources/canvasservice.js#L1849) — that's stage 3/4 territory, not stage 1/2.
- The route is registered as `app.get('/downloadfile/:docid', canvasService.downloadFile)` with **no body parser at all** — see the route table in [`server.js#L206-L244`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/DocService/sources/server.js#L206-L244). `rawFileParser` (the `bodyParser.raw({ limit: limits_tempfile_upload })` that `MAX_FILE_SIZE` controls, [`L206-L211`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/DocService/sources/server.js#L206-L211)) is only attached to `/command`, `/converter`, `/upload/:docid`, `/savefile/:docid`, `/downloadfile/:docid` **POST** — image inserts, save-back, WOPI callbacks. Nothing in the "open a stored document" path.
- `NGINX_CLIENT_MAX_BODY_SIZE` gates inbound POST bodies to nginx — same problem, no inbound POST body exists when the Document Server is the one fetching the file.

So of the four stages walked through in the doc, only 3 and 4 (`FILECONVERTER_MAX_DOWNLOAD_BYTES`, `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED`) apply to the "user opens a 200 MB PPTX" scenario as written. Following the guide's stage 1/2 advice for that scenario won't do anything.

### 2. `USE_UNAUTHORIZED_STORAGE` description is wrong, and understates the risk

[`installation_docker.rst#L161-L163`](https://github.com/nextcloud/documentation/blob/1eba328af5065315232497fcf90e2bfab3b00a6e/admin_manual/office/euro-office/installation_docker.rst#L161-L163) documents it as "Allow fetching documents from HTTP (non-TLS) storage."

It actually flows into `requestDefaults.rejectUnauthorized`, cloned into the HTTPS agent options for the exact same `downloadUrlPromise` call used to fetch documents — see [`utils.js#L345-L355`](https://github.com/Euro-Office/server/blob/c438fd3d336497b2acf7bfe6b4f9ee4fde1fcfbf/Common/sources/utils.js#L345-L355), specifically `httpsAgentOptions = {...https.globalAgent.options, ...options}` at L354. `rejectUnauthorized: false` is the standard Node.js TLS option for skipping certificate-chain validation — it has no bearing on plain HTTP (no certificate is ever presented on an HTTP connection either way). Flip this thinking you're permitting a legacy plaintext endpoint, and you've disabled cert validation on every HTTPS storage connection too — an unlabelled MITM exposure.

### 3. The PR's stated goal — "document every environment variable" — isn't quite true

`entrypoint.sh` still reads several operator-facing vars the table skips (all in [`entrypoint.sh`](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh)):
`SSL_CERTIFICATE_PATH`/`SSL_KEY_PATH`/`SSL_DHPARAM_PATH` ([L349-L367](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L349-L367)), `SECURE_LINK_SECRET` ([L118-L127](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L118-L127)), `METRICS_HOST`/`PORT`/`PREFIX` ([L45-L47](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L45-L47)), `REDIS_SERVER_USER`/`DB`, `AMQP_URI`/`VHOST`, `NGINX_WORKER_CONNECTIONS`, `ADMINPANEL_ENABLED`/`EXAMPLE_ENABLED`, `DS_LOG_LEVEL`.

Worth prioritizing the SSL ones: this PR *does* document `SSL_VERIFY_CLIENT` and both `ONLYOFFICE_HTTPS_HSTS_*` vars, but all three only take effect inside the `if [ -n "$SSL_CERTIFICATE_PATH" ] && [ -n "$SSL_KEY_PATH" ]` gate at [L349-L367](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L349-L367). Document those three without the two vars that gate them, and a reader can set all three and see zero effect.

### 4. Minor — undocumented edge-case behavior (low priority)

- `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED` replaces the *entire* `inputLimits` array rather than patching one entry ([entrypoint.sh#L261-L270](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L261-L270)). Harmless today (only four format groups exist in `default.json`), but a footgun if a fifth type group is ever added upstream.
- `FILECONVERTER_MAX_DOWNLOAD_BYTES` is silently ignored (stderr warning only) if it isn't a plain integer ([entrypoint.sh#L255-L260](https://github.com/Euro-Office/DocumentServer/blob/eca0e49acfb6cf14ad47e6a9d31ea367ac5f722d/build/scripts/standalone/entrypoint.sh#L255-L260)) — worth a one-line callout.

</details>

chrip:
Thanks @moodyjmz — this was a genuinely useful review, and you were right on both
blocking points. Pushed as `743a812b16`. Item by item:

### 1. Size limits narrative — fixed

You were right that opening a stored document is a server-side download, so
`NGINX_CLIENT_MAX_BODY_SIZE` and `MAX_FILE_SIZE` never enter that path. Rather than
reorder the four stages I split the section into the two paths that actually exist,
because a single sequence implies a single pipeline that isn't there:

- **Documents the server downloads** — `FILECONVERTER_MAX_DOWNLOAD_BYTES`,
  `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED`
- **Files posted to the server** — `NGINX_CLIENT_MAX_BODY_SIZE`, `MAX_FILE_SIZE`, scoped
  to the routes your `rawFileParser` list covers: image inserts, save-back, and
  conversion or command requests

The 200 MB PPTX example stays but now resolves correctly: the 500 MB download default is
already sufficient, so only `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED` has to be raised. The
summary table now maps each variable to which path it governs instead of to a stage
number.

### 2. `USE_UNAUTHORIZED_STORAGE` — fixed

Rewritten to say it disables TLS certificate validation, with a warning that it sets
`rejectUnauthorized: false` on every outbound HTTPS connection, that chain, host name and
expiry all stop being checked, and that it has no effect on plain HTTP since no
certificate is presented there. Recommends adding the CA to the container trust store
instead.

### 3. Missing variables — added

All of them: `SSL_CERTIFICATE_PATH`, `SSL_KEY_PATH`, `SSL_DHPARAM_PATH`,
`SECURE_LINK_SECRET`, `METRICS_HOST`/`PORT`/`PREFIX`, `REDIS_SERVER_USER`/`DB`,
`AMQP_VHOST`/`URI`, `NGINX_WORKER_CONNECTIONS`, `ADMINPANEL_ENABLED`, `EXAMPLE_ENABLED`,
`DS_LOG_LEVEL`.

On the SSL gate you flagged — one refinement. The condition is stricter than
`-n`-on-both; it also requires the files to exist:

```sh
if [ -n "${SSL_CERTIFICATE_PATH:-}" ] && [ -n "${SSL_KEY_PATH:-}" ] \
   && [ -f "$SSL_CERTIFICATE_PATH" ] && [ -f "$SSL_KEY_PATH" ] \
   && [ -f "$NGINX_DS_SSL_TMPL" ]; then
```

So a typo'd path fails the same way an unset one does, silently and with plain HTTP
served. The note says set **and** both files exist, since that's the more likely way an
admin trips over it.

### 4. Edge cases — both documented

Not deferred, they were cheap:

- `FILECONVERTER_MAX_DOWNLOAD_BYTES` with a unit suffix is silently ignored — warning to
  stderr, built-in default kept, container starts normally.
- `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED` replaces the whole `inputLimits` array rather
  than patching one entry, so a future fifth format group would lose its own default.

### Two things beyond the review

**Tables grouped by concern.** At ~40 variables one flat table had stopped being
readable, so it is now split into Authentication, Database, Redis, RabbitMQ, WOPI,
Outbound requests, HTTPS and TLS termination, Nginx, Metrics, and Logging and optional
services. Larger diff than the review asked for, but it seemed better to do it while the
section was already being rewritten than to touch every row twice.

**`/var/www/euro-office/Data` documented.** Not from your review, but adjacent and worth
having: the volume list omitted it, and it holds `runtime.json`, the generated
`jwt_secret` and `secure_link_secret`, and the WOPI key pair. Lose it and the JWT secret
regenerates on next start, silently breaking the connector until the new value is copied
into Nextcloud. Same omission existed downstream and is fixed there too.

### Verification

`sphinx-lint`, `codespell`, and `sphinx-build -W --keep-going` on `admin_manual` all
clean. Every default in the new tables was checked against `entrypoint.sh` rather than
carried over — which is how the `NGINX_WORKER_CONNECTIONS` default landed on `768`
(Ubuntu's packaged `nginx.conf` in the `ubuntu:24.04` base) rather than the `1024` in the
repo's own orchestrated config, and `DS_LOG_LEVEL` on `WARN` from
`Common/config/log4js/production.json`.

Worth noting this same content also lives in the Euro-Office project docs, which is where
it was written first. Both defects were present there too and are now corrected in
[Euro-Office/documentation#13](https://github.com/Euro-Office/documentation/pull/13), so
the two aren't going to disagree.

Ready for another look when you have time.

moodyjmz (approved):
**Verdict: Approve.**

Re-verified `743a812b` independently against source rather than taking the writeup on trust — both blocking points are actually fixed, not just described as fixed.

- **Size limits narrative:** now correctly split into "documents the server downloads" vs. "files posted to the server." Confirmed against `server.js`: `GET /downloadfile/:docid` has no `rawFileParser` attached, only the `POST` variant does — matches the new text.
- **`USE_UNAUTHORIZED_STORAGE`:** correctly describes `rejectUnauthorized: false` on outbound HTTPS, with the MITM warning and the "no effect on plain HTTP" caveat.

Spot-checked ~15 of the newly-added variable defaults directly against `entrypoint.sh` / `log4js/production.json` (`SECURE_LINK_SECRET` 20-char, `JWT_SECRET` 32-char, `DS_LOG_LEVEL` default `WARN`, `METRICS_*` gating, `REDIS_SERVER_USER`/`DB`, `AMQP_VHOST`/`URI` precedence, `ADMINPANEL_ENABLED`/`EXAMPLE_ENABLED` defaults, the SSL cert/key gate condition, `FILECONVERTER_INPUT_LIMIT_UNCOMPRESSED` array-replace behavior) — all accurate.

Also checked the "document every environment variable" claim by diffing every `${VAR}` reference in `entrypoint.sh` against the doc table; the only misses (`WOPI_MODULUS`/`WOPI_EXPONENT`/`WOPI_PRIVATE_KEY_DATA`/`WOPI_PUBLIC_KEY_DATA`/`WELCOME_PAGE`/`EXAMPLE_LOCAL`/`EXAMPLE_DISABLED_PAGE`) are internal derived shell variables, never read from the environment — correctly excluded.

Nice work on the fix, and thanks for the detailed writeup — checked out end to end.

PR #15470, opened 2026-08-20
Add HSTS header to static assets

janepie:
### ☑️ Resolves

nginx only inherits add_header directives from an outer level if the inner level defines none of its own. Because this block defines its own headers, the server-level Strict-Transport-Security is dropped and has to be added here too

### DIFF
--- a/admin_manual/installation/nginx-root.conf.sample
+++ b/admin_manual/installation/nginx-root.conf.sample
@@ -225,6 +225,15 @@ server {
     # Serve static files
     location ~ \.(?:css|js|mjs|svg|gif|ico|jpg|png|webp|wasm|tflite|map|ogg|flac|mp4|webm)$ {
         try_files $uri /index.php$request_uri;
+	
+		# HSTS settings
+		# WARNING: Only add the preload option once you read about
+		# the consequences in https://hstspreload.org/. This option
+		# will add the domain to a hardcoded list that is shipped
+		# in all major browsers and getting removed from this list
+		# could take several months.
+		#add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
+
         # HTTP response headers borrowed from Nextcloud `.htaccess`
         add_header Cache-Control                     "public, max-age=15778463$asset_immutable";
         add_header Referrer-Policy                   "no-referrer"       always;


joshtrichards (commented):
I can't think of a better way to do this right now, even with restructuring. When [Nginx 1.29.3+](https://blog.nginx.org/blog/nginx-open-source-1-29-3-and-1-29-4) is ubiquitous we can switch to using `add_header_inherit merge;`, but don't think we're there yet...

That said, I'm not sure it really matters since an HSTS policy is host-level not request path  or asset driven. Once the browser gets the policy, it'll apply it to all requests thereafter. Maybe `max-age` refreshing, but other requests will cover that in practice.

Technically, having it here though would be more consistent and possibly in some weird edge cases it would head-off compliance / heading-scanning noise. But the confusion and noise from being duplicated in the config bugs me unless someone thinks it's really important... ;-)

Good catch, @janepie, in any case!

pabzm (approved):
Looks good to me, thank you for the catch and the fix!

One idea, which shall not block this from being merged: Maybe we should add a comment to both occurrences of this block to also take care of the other occurrence, if it shall be changed?

(That would consequently also matter for the other repeated options, though.)

admin_manual/installation/nginx-root.conf.sample
```diff
@@ -225,6 +225,15 @@ server {
...
     location ~ \.(?:css|js|mjs|svg|gif|ico|jpg|png|webp|wasm|tflite|map|ogg|flac|mp4|webm)$ {
         try_files $uri /index.php$request_uri;
+	
+		# HSTS settings
+		# WARNING: Only add the preload option once you read about
+		# the consequences in https://hstspreload.org/. This option
+		# will add the domain to a hardcoded list that is shipped
+		# in all major browsers and getting removed from this list
+		# could take several months.
+		#add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
```
MichaIng, line 235: Makes sense, just noting that the indentation is off (2 tabs vs 8 spaces):
```suggestion
        # HSTS settings
        # WARNING: Only add the preload option once you read about
        # the consequences in https://hstspreload.org/. This option
        # will add the domain to a hardcoded list that is shipped
        # in all major browsers and getting removed from this list
        # could take several months.
        #add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
```
_That Nginx skip any parent location headers, once a child location has any header-related directive (`add_header Cache-Control` here), is pretty annoying, an non-intuitive if not already used to nested Nginx configurations._

PR #15476, opened 2026-08-22
docs(auth): clarify password storage and length limits

joshtrichards:
### ☑️ Resolves

N/A - prompted by a question about the ambiguity of the existing documentation [on the forum](https://help.nextcloud.com/t/encrypting-passwords-instead-of-hashing/248436) along with my recurring own need to refer to the code and/or notes in this area.

### Summary

Clarifies password storage, authentication-token storage, and password-length guidance in the server hardening documentation.

The existing text can be interpreted as saying that Nextcloud stores account passwords using reversible encryption instead of standard password hashing. In practice, these are separate mechanisms:

- The built-in database user backend stores a salted, one-way account-password hash, preferring Argon2id when supported by PHP and falling back to Argon2i or bcrypt.
- Authentication tokens are represented in the database by a hash derived from the token and the instance-specific secret.
- When `auth.storeCryptedPassword` is enabled and the login password is available, a separate, reversibly encrypted copy may be stored in the server-side record associated with an authentication token.

The documentation now distinguishes those representations and explains why a recoverable copy may exist, how it is protected, and the security consequences of a combined token, configuration, and database compromise.

It also clarifies that:

- The authentication token itself does not contain the login password.
- Passwordless and some SSO-generated token records may not contain a recoverable password.
- Administrators can disable recoverable password storage with `auth.storeCryptedPassword`, subject to the documented functional consequences.
- External user backends control their own account-password storage and verification.
- Standard Nextcloud interfaces accept account passwords up to 469 bytes.
- The 214-byte RSA-key threshold is a token-encryption performance consideration, not a password-length limit.
- Bcrypt's 72-byte input behavior is specific to bcrypt and is not a general Nextcloud password-length limit.
- Passwords for public link and mail shares are handled separately from token-associated password storage.

Related implementation follow-up:

- ~~nextcloud/server#54091~~ nextcloud/server#63806 documents the `auth.storeCryptedPassword` configuration parameter specifically
- Somewhat dependent on nextcloud/server#63804, but technically speaking the 214-byte limit is accurate either way; that PR is more about correcting an internal implementation discrepancy.

### TODO

- [ ] Backport to v35-v34
- [ ] Determine whether safe to backport to v33 based on implementation details (I think so but have not verified)
- [ ] Decide whether to bother backporting to v32

### ✅ Checklist

### DIFF
--- a/admin_manual/installation/harden_server.rst
+++ b/admin_manual/installation/harden_server.rst
@@ -100,14 +100,49 @@ impact depends on the data and capabilities available to the actor:
   way to recover the password. However, password hashes must still be protected
   against offline password-guessing attacks.
 
-Limit on password length
-^^^^^^^^^^^^^^^^^^^^^^^^
+.. _password_length_limits:
 
-Nextcloud uses the bcrypt algorithm, and thus for security and performance
-reasons, e.g. Denial of Service as CPU demand increases exponentially, it only
-verifies the first 72 characters of passwords. This applies to all passwords
-that you use in Nextcloud: user passwords, passwords on link shares, and
-passwords on external shares.
+Password Length Limits
+^^^^^^^^^^^^^^^^^^^^^^
+
+Nextcloud accepts account passwords of up to 469 bytes through its standard
+account-creation, password-change, and password-reset interfaces. This is the
+maximum account-password length enforced by these interfaces. Because the
+limit is measured in bytes, a password containing multibyte characters (such as
+emojis or characters from non-Latin scripts) may reach the limit with fewer than
+469 characters.
+
+Administrators can use the :doc:`Password Policy app </configuration_user/user_password_policy>`
+to configure requirements such as a minimum password length and other
+complexity rules. External user backends may impose additional or different
+requirements.
+
+The following implementation details do not reduce the general password limits,
+but they are relevant when selecting a password policy:
+
+Token Encryption Performance
+    When ``auth.storeCryptedPassword`` is enabled and an account password is
+    longer than 214 bytes, Nextcloud uses a larger RSA key when creating
+    authentication-token records. This increases token-generation overhead, but
+    does not prevent passwords between 214 and 469 bytes from being accepted.
+    The 214-byte threshold is therefore a performance consideration, not a
+    password-length limit.
+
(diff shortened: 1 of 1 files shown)


admin_manual/installation/harden_server.rst
```diff
@@ -14,15 +14,91 @@ in a Linux environment.
...
+
+The authentication token itself does not contain the login password. For each
+token record, Nextcloud generates a separate RSA key pair. The login password
+is encrypted with the record's public key, while the corresponding private key
+is encrypted using the authentication token together with the
+instance-specific ``secret`` from ``config.php``. The token and instance secret
+are therefore required to decrypt the password stored in that token record.
+
+The recoverable copy is used by features that require the login credentials,
+such as connecting to external storage, configuring mail accounts, and
```
tflidd, line 72: I would put this part first for clarification. I was first thinking about the Nextcloud user password, why it needs to be store, if this is for the server-side encryption etc.

come-nc (approved):

admin_manual/installation/harden_server.rst
```diff
@@ -14,15 +14,91 @@ in a Linux environment.
...
+when no password or equivalent user secret is available, as can occur with
+some SSO or passwordless authentication flows. Existing token records can also
+remain without a stored password until Nextcloud receives the password during
+a later login or password update.
+
+Administrators can disable this behavior with
+``auth.storeCryptedPassword``. Disabling it does not affect the one-way
+account-password hash used by the built-in user backend, but features that
+require recovery of the login password may no longer work. Password changes
+made directly in an external user backend might also no longer automatically
```
artonge, line 83: > might also no longer automatically

There are a lot of incertitude in this PR. This is one example. It would be better to fact check them to be able to provide a more solid documentation. But I know it takes time 🙈

admin_manual/installation/harden_server.rst
joshtrichards: Done / improved.

admin_manual/installation/harden_server.rst
joshtrichards: Good point. Done / attempted.

tflidd (approved):

PR #15479, opened 2026-08-24
docs(admin): clarify purpose of utf8mb4/full Unicode support

joshtrichards:
### ☑️ Resolves

Clarify that configuring MySQL/MariaDB with `utf8mb4` enables support for the full Unicode character range, not only emoji.

This change:

- Reframes the chapter around full Unicode support while retaining emoji as a familiar example.
- Explains that the older three-byte `utf8` character set cannot store characters outside the Basic Multilingual Plane.
- Distinguishes the `utf8mb4` character set from the collation used for comparison and sorting.
- Updates the migration steps to use consistent MySQL/MariaDB and `utf8mb4` terminology.
- Clarifies the backup guidance for preserving four-byte Unicode characters.
- Fixes minor grammar and formatting issues.

## Motivation

The previous wording could imply that four-byte support is needed only for emoji. It is also required for less-common writing systems, historic scripts, specialized symbols, and other Unicode characters outside the Basic Multilingual Plane.

### ✅ Checklist

### DIFF
--- a/admin_manual/configuration_database/mysql_4byte_support.rst
+++ b/admin_manual/configuration_database/mysql_4byte_support.rst
@@ -1,13 +1,16 @@
-=====================================
-Enabling MySQL/MariaDB 4-byte support
-=====================================
+===========================================
+Enabling MySQL/MariaDB full Unicode support
+===========================================
 
 .. note::
 
-    Be sure to backup your database before performing this database upgrade.
+    Be sure to back up your database before performing this database upgrade.
 
-In order to use Emojis (textbased smilies) on your Nextcloud server with a MySQL database, the
-installation needs to be tweaked a bit.
+To store the full range of Unicode characters in Nextcloud, including emoji,
+less-common writing systems, historic scripts, and specialized symbols, a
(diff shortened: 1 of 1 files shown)


admin_manual/configuration_database/mysql_4byte_support.rst
```diff
@@ -38,35 +41,50 @@ The result should look like this::
...
 
     ALTER DATABASE nextcloud CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 
-5. Set the ``mysql.utf8mb4`` config to true in your config.php::
+.. note::
+
+    The character set determines which characters can be stored. The collation
+    determines how text is compared and sorted.
+
+5. Enable ``utf8mb4`` support in the Nextcloud system configuration::
```
nickvergessen, line 63: sounds a bit confusing that this is not the config name, but it sounds close to it, especially since it's marked as code. Why not keeping it `mysql.utf8mb4`

skjnldsv (approved):

PR #15495, opened 2026-08-26
fix(ai): correct the Context Chat embedding API key env var name

bakiburakogun:
The Context Chat scaling section documents the embedding API key as `CC_EM_API_KEY`, but `context_chat_backend` reads `CC_EM_APIKEY`.

In `context_chat_backend/config_parser.py`:

```python
if os.getenv('CC_EM_BASE_URL'):
    if os.getenv('CC_EM_APIKEY'):
        auth = TEmbeddingAuthApiKey(apikey=os.environ['CC_EM_APIKEY'])
    elif os.getenv('CC_EM_USERNAME') and os.getenv('CC_EM_PASSWORD'):
```

`CC_EM_APIKEY` appears four times across the repository; `CC_EM_API_KEY` appears nowhere. The other three names in the same sentence — `CC_EM_BASE_URL`, `CC_EM_MODEL_NAME`, `CC_EM_USERNAME`, `CC_EM_PASSWORD` — all match the code.

The failure mode is quiet: an administrator who follows the documentation sets a variable nothing reads, `auth` falls through to `None`, and the backend talks to the embedding service unauthenticated. Depending on how that service answers, the first sign of trouble is an indexing failure some distance away from the cause.

I hit this while setting up an on-premises embedding service for a Nextcloud 34 installation.

Only the documented name is changed here. If you would rather the code accept both spellings, I am happy to open that against `context_chat_backend` instead.

### DIFF
--- a/admin_manual/ai/app_context_chat.rst
+++ b/admin_manual/ai/app_context_chat.rst
@@ -100,7 +100,7 @@ Listed below are the major parts of the system that can be scaled independently 
 3. The embedding model performance
 
    | The embedding model performance can be scaled by using a hosted embedding service, locally or remotely hosted. It should be able to serve an OpenAI-compatible API.
-   | The embedding service URL can be set using the environment variable ``CC_EM_BASE_URL`` during deployment in the "Deploy Options". Other options like the model name, api key, or username and password can be set using the environment variables ``CC_EM_MODEL_NAME``, ``CC_EM_API_KEY``, ``CC_EM_USERNAME``, and ``CC_EM_PASSWORD`` respectively.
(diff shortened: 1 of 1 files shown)


kyteinsky (approved):
thanks!

kyteinsky:
will backport this after https://github.com/nextcloud/documentation/pull/15491 and its backports are in.

bakiburakogun:
This has been approved since 27 August and the backports were queued on the 31st, so I think it is only the two red checks holding it, and neither of them can go green from here.

`Deploy Netlify preview` and `Label by changed files` fail on fork pull requests in this repository and pass on branch ones. Of the pull requests open right now that have results recorded for those two jobs, the three from forks — this one, #15557 and #15610 — fail both, and every one from a branch in the repository passes or skips them.

The change itself is one line in one file: `CC_EM_API_KEY` in `admin_manual/ai/app_context_chat.rst` does not exist, the variable the backend reads is `CC_EM_APIKEY`.

Happy to rebase or reopen from a branch if that is easier for you, just say which.

kyteinsky:
thanks for the ping, had created an issue for the labeller, did one for the netlify PR comment too now.
will see if this can be force merged soon.

PR #15527, opened 2026-08-29
docs(ai/assistant): document the assistant data folder name

bakiburakogun:
Documents the assistant data folder name, added in nextcloud/assistant#632.

The assistant creates a folder in each user's files for the content it generates — generated images, speech-to-text output. Its name used to be a constant; it is now an app config value with a per-user override, and both are editable in the admin and personal settings. Neither the keys nor the behaviour around existing folders was written down anywhere.

Added as item 5 of the existing "Assistant configuration" list, covering:

- `occ config:app:set assistant default_data_folder` for the server-wide default,
- `occ user:setting $USER_ID assistant data_folder` for the per-user value,
- that changing either does not rename or move folders that already exist,
- that a user who already has a folder under the built-in `Assistant` name keeps it, instead of getting a second folder and leaving their earlier output behind.

The last two are the parts that surprise people, so they seemed worth stating rather than leaving to be discovered.

Note that the app change is merged but not in a release yet, so this targets `master` only — happy to follow up with backports once it ships.


kyteinsky (commented):
would be nice to document this in the user manual too: https://github.com/nextcloud/documentation/blob/master/user_manual/ai_assistant.rst

but the changes look good!

bakiburakogun:
Added, thanks — `user_manual/ai_assistant.rst` now covers it from the user's side: where the assistant saves what it generates, that the folder is called `Assistant` unless the administrator picked another name for the instance, that it can be renamed under **Data folder** in the personal settings with the placeholder showing the instance default, and the note that renaming it leaves the folder you already have alone.

One thing I ran into while writing it. The skills section said personal skills live under `Assistant/Context Agent/Skills/`, but `AgentSkillsService::getSkillsFolder()` resolves `Context Agent/Skills` inside `getAssistantDataFolder()`, so that path only holds while the folder still has its default name. I have stated it relative to the data folder and kept the default spelling as the example, so the folder tree and the `weekly-report-format` example below it still read the same way.

I also rebased onto `master`, since the branch had drifted 153 commits behind. Same single commit, and the added and removed lines are identical to before the rebase.

kyteinsky (approved):
perfect, thanks!

kyteinsky:
yeah the example paths in skills section are fine, users can deduce "Assistant/" means the Assistant's folder if they renamed it, the comment at the top clarifies it too with your change.

bakiburakogun:
The build jobs had not run yet when you approved this — they were still sitting in the queue. They have finished since: `Lint RST files`, `Building user_manual HTML`, `Building admin_manual HTML` and `Build index.html` are all green, along with `Check spelling`.

That leaves 21 passed, 4 skipped and one red, and the red one is `Deploy Netlify preview` — the same fork job you opened the issue for yesterday.

No rush from my side, just flagging it since the state changed after your review.

